PR #2, opened 2019-08-02
Pst/custom token

purestaketdb:
Add support for multiple different headers - backwards compatible for existing code

### DIFF
--- a/algosdk/algod.py
+++ b/algosdk/algod.py
@@ -13,11 +13,11 @@ class AlgodClient:
     Client class for kmd. Handles all algod requests.
 
     Args:
-        algod_token (str): algod API token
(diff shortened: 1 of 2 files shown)


algosdk/algod.py
```diff
@@ -37,12 +37,15 @@ def algod_request(self, method, requrl, params=None, data=None):
...
         """
-        if requrl in constants.no_auth:
-            header = {}
+        if isinstance(self.algod_token, dict):
+            header = self.algod_token
         else:
-            header = {
-                constants.algod_auth_header: self.algod_token
-                }
+            if requrl in constants.no_auth:
```
algobolson: bump this up to one set of if/elif/else ?

PR #9, opened 2019-08-15
add __eq__ for objects in transaction and auction

mjiang102628:


algosdk/util.py
```diff
@@ -28,7 +29,7 @@ def algos_to_microalgos(algos):
     Returns:
         int: how many microalgos
     """
-    return algos*constants.microalgos_to_algos_ratio
+    return round(algos*constants.microalgos_to_algos_ratio)
```
EvanJRichard: We `round` here because we only want whole-number microAlgos?

algosdk/util.py
```diff
@@ -15,7 +16,7 @@ def microalgos_to_algos(microalgos):
     Returns:
         int or float: how many algos
     """
-    return microalgos/constants.microalgos_to_algos_ratio
+    return decimal.Decimal(microalgos)/constants.microalgos_to_algos_ratio
```
EvanJRichard: We use `decimal` here to make sure we properly get the decimal amount of Algos?

algosdk/transaction.py
```diff
@@ -198,6 +218,7 @@ class KeyregTxn(Transaction):
         votekd (int)
         note (bytes)
         genesis_id (str)
+        type (str)
```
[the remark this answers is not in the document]
mjiang102628: Type of transaction: payment transaction ("pay") or key registration ("keyreg").

algosdk/util.py
mjiang102628: Yes; since microAlgos never get split, I don't think we should have fractional microAlgo amounts.

algosdk/util.py
mjiang102628: Yes, because python's built in floating point arithmetic doesn't have enough significant digits when you get into higher numbers.

PR #15, opened 2019-09-13
Draft: PY Asset Creation

EvanJRichard:


mjiang102628:
I don't think we need a separate AssetCreationTxn class -- all the fields are included in AssetConfigTxn. It would probably be useful to write a factory function under AssetConfigTxn for each of creating, updating config, and deleting assets (less confusing for users, since they don't have to figure out which fields are needed for each type).

EvanJRichard:
@mjiang102628 +1 :+1: to that, I'll add a ticket for circling back and doing that to the ungroomed backlog. Closing.

PR #16, opened 2019-09-16
Minor fixes for dependencies, integration test, crypto

pzbitskiy:
1. Use setup.py as source of truth for dependencies
2. Allow ALGORAND_DATA to set data dir in integration test
3. Add encoding.checksum function to incapsulate hashing logic


EvanJRichard:
LGTM but the travis tests are failing - and, I may be mistaken, but it seems the test failures are unrelated? I'm taking a look now, but I'm not sure.

mjiang102628:
LGTM also; I looked at the error message and it seems like the js sdk is causing the problem (tests start failing on the commit that changes which msgpack package the js sdk is using).

pzbitskiy:
> LGTM also; I looked at the error message and it seems like the js sdk is causing the problem (tests start failing on the commit that changes which msgpack package the js sdk is using).

Right, node js version needs to be 10+ for "for await" expression. Updated but some JS tests still failing.

pzbitskiy:
@rotemh please merge :) Travis CI - Branch does not want to build changes from my fork but for some reason is a prereq for merge.

PR #17, opened 2019-09-16
Test build

pzbitskiy:


pzbitskiy:
Verified that cross-sdk build is broken, closing.

PR #19, opened 2019-09-16
Transaction group support

pzbitskiy:
## Summary
group field in transactions
transaction.calculate_group_id function and internal TxGroup class for group id calc
send_transactions for group of [signed] transaction sending

## Test
unit tests: serialization, group id agains goal clerk
integration test: create/sign/send
Note: sending does not actually work, groups are not enabled in the protocol yet


algosdk/encoding.py
```diff
@@ -59,12 +59,16 @@ def msgpack_decode(enc):
...
+    if "txlist" in decoded:
+        return transaction.TxGroup.undictify(decoded)
     if "t" in decoded:
         return auction.NoteField.undictify(decoded)
     if "bid" in decoded:
         return auction.SignedBid.undictify(decoded)
     if "auc" in decoded:
         return auction.Bid.undictify(decoded)
+    if "auc" in decoded:
+        return auction.Bid.undictify(decoded)
```
EvanJRichard: Are lines 70-71 a duplicate of lines 68-69?

pzbitskiy: Good catch! Not sure how it happened, fixed.

README.md
```diff
@@ -244,6 +244,43 @@ decoded = encoding.msgpack_decode(base64.b64encode(note_field_bytes))
...
+last_round = params["lastRound"]
+fee = params["fee"]
+
+# create a transaction
+amount = 10000
+txn1 = transaction.PaymentTxn(sender, fee, last_round, last_round+100, gh, receiver, amount)
+txn2 = transaction.PaymentTxn(receiver, fee, last_round, last_round+100, gh, sender, amount)
+
+# get group id and assign it to transactions
+gid = transaction.calculate_group_id([txn1, txn2])
```
rotemh: This might be cumbersome with a large number of txns. Please consider changing the function `calculate_group_id` to `assign_group_id([txn1, txn2,...])` that returns the list of txns with the tags already assigned.

rotemh:
once small suggestion, otherwise, looks good to me.

README.md
pzbitskiy: I also thought about it and it does not look practical - main use case is to group transactions signed by different keys, possible received from different people, so looping and signing would not be the main use case. My feeling calc function would be used more often than assign, so let's have both of them.

rotemh: Actually, here is a better idea than my previous one - let it take an optional Algorand address and output just the relevant ones. 

Also, we can definitely have both.

pzbitskiy: done

PR #21, opened 2019-09-27
Support for asset freeze/un-freeze transactions

EvanJRichard:
## Summary
This PR proposes to add a new `Transaction` subclass, `AssetFreezeTransaction`. It allows the freeze-manager of an asset to block an account from transferring assets, or for the freeze-manager to allow a frozen account to transfer assets.

### Testing
New test added to `test_unit.py`, `test_serialize_assetfreeze`.

### DIFF
--- a/algosdk/transaction.py
+++ b/algosdk/transaction.py
@@ -100,7 +100,7 @@ class PaymentTxn(Transaction):
         amt (int): amount in microAlgos to be sent
         close_remainder_to (str, optional): if nonempty, account will be closed
             and remaining algos will be sent to this address
-        note (bytes, optional): encoded NoteField object
+        note (bytes, optional): arbitrary optional bytes
         gen (str, optional): genesis_id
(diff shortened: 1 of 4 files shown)


algosdk/transaction.py
```diff
@@ -496,6 +496,111 @@ def __eq__(self, other):
...
+        sender (str): address of the sender, who must be the asset's freeze manager.
+        fee (int): transaction fee (per byte if flat_fee is false)
+        first (int): first round for which the transaction is valid
+        last (int): last round for which the transaction is valid
+        gh (str): genesis_hash
+        creator (str): creator of the asset
+        index (int): index of the asset
+        target (str): address having its assets frozen or unfrozen
+        new_freeze_state (bool): true if the assets should be frozen, false if they should be transferrable
+        note (bytes, optional): encoded NoteField object
```
mjiang102628: not necessarily an encoded NoteField object; could just be random bytes

PR #22, opened 2019-10-09
LogicSig support

pzbitskiy:


algosdk/logic.py
```diff
@@ -0,0 +1,122 @@
...
+        raise error.InvalidProgram("empty program")
+    if not args:
+        args = []
+
+    if spec is None:
+        script_path = os.path.realpath(__file__)
+        script_dir = os.path.dirname(script_path)
+        langspec_file = os.path.join(script_dir, "data", "langspec.json")
+        with open(langspec_file, "rt") as fin:
+            spec = json.load(fin)
```
mjiang102628: This won't work with the distribution package since that only includes things under the algosdk package. I think you can just move langspec.json into algosdk (something like algosdk/data/langspec.json) and then reference it with open("data/langspec.json") since relative paths should work in this case.

algosdk/logic.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Performs program checking for max length and cost
+
+    Args:
+        program (bytes)
+        args (list[bytes])
+
+    Returns:
+        True on success
+        Raises InvalidProgram on error
```
mjiang102628:
```
Args:
     program (bytes)
     args (bytes[])

Returns:
     True on success

Raises:
     InvalidProgram
```

algosdk/logic.py
```diff
@@ -0,0 +1,122 @@
+import json
+import os
+
+from . import error
+
+spec = None
+opcodes = None
+
+MAX_COST = 20000
+MAX_LENGTH = 1000
```
mjiang102628: Should MAX_COST and MAX_LENGTH go in constants.py? Also, is not setting spec and opcodes to avoid processing langspec.json if not needed?

algosdk/logic.py
```diff
@@ -0,0 +1,122 @@
...
+    for arg in args:
+        length += len(arg)
+
+    if length > MAX_LENGTH:
+        raise error.InvalidProgram("program too long")
+
+    if opcodes is None:
+        opcodes = dict()
+        for op in spec['Ops']:
+            opcodes[op['Opcode']] = op
```
mjiang102628: opcodes = {opcode: op for op, opcode in spec['Ops'].items()}

algosdk/logic.py
```diff
@@ -0,0 +1,122 @@
...
+        op = opcodes.get(program[pc], None)
+        if op is None:
+            raise error.InvalidProgram("invalid instruction")
+
+        cost += op['Cost']
+        size = op['Size']
+        if size == 0:
+            if op['Opcode'] == 32:  # intcblock
+                size += check_int_const_block(program, pc)
+            elif op['Opcode'] == 38:    # bytecblock
```
mjiang102628: make 32 and 38 constants

algosdk/logic.py
pzbitskiy: I added langspec.json to setup.py, so it is in the package. that's the point, script_dir is abs path of logic.py whatever it run from. and it opens data/langspec.json that is expected to be near it.

algosdk/transaction.py
```diff
@@ -693,6 +695,31 @@ def address(self):
         addr = encoding.checksum(msig_bytes)
         return encoding.encode_address(addr)
 
+    def verify(self, message):
+        """Verify that the multisig is valid for the message"""
+        try:
+            self.validate()
```
mjiang102628: does validate() refer to MultisigTransaction.validate()? If so, self.validate() will never work since MultisigTransaction isn't a subclass of Transaction. Maybe this function should go under MultisigTransaction?

algosdk/logic.py
pzbitskiy: this does not work, Ops is an array of dicts, but anyway this is a matter of opinion.

algosdk/transaction.py
pzbitskiy: this is part of Multisig class so that it calls Multisig.validate

pzbitskiy:
@mjiang102628 thank you for reviewing this, the change is quite large but I keep it very similar between SDKs. I'll fix remaining comments.

algosdk/logic.py
pzbitskiy: done

algosdk/logic.py
pzbitskiy: done

algosdk/logic.py
pzbitskiy: done

algosdk/logic.py
mjiang102628: Oddly, this works with the wheel file but not the tar. Seems like the fix is to add a MANIFEST.in file.

pzbitskiy: Added to MANIFEST.in as well.

PR #32, opened 2019-10-28
encoding fixes and standard golden tests

mjiang102628:
- add additional asset description fields (#27)
- update encoding and algod wrapper for new asset IDs
- add wrapper for list_assets


algosdk/algod.py
```diff
@@ -159,17 +159,39 @@ def account_info(self, address, **kwargs):
...
+            max_index (int, optional): maximum asset index; defaults to 0,
+                which lists most recent assets
+            max_assets (int, optional): maximum number of assets (0 to 100);
+                defualts to 100
+        """
+        query = dict()
+        if max_index is not None:
+            query["assetIdx"] = max_index
+        else:
+            query["assetIdx"] = 0
```
rotemh: nit - 
`query["assetIdx"] = max_index if max_index is not None else 0`

algosdk/transaction.py
```diff
@@ -336,46 +335,67 @@ class AssetConfigTxn(Transaction):
...
-        self.creator = creator
         self.index = index
         self.total = total
         self.default_frozen = default_frozen
         self.unit_name = unit_name
         self.asset_name = asset_name
+        if unit_name != None:
+            if len(unit_name) > constants.unit_name_length:
+                raise error.WrongUnitNameLengthError
+            self.unit_name = unit_name + (constants.unit_name_length - len(unit_name))*b'\x00'
```
rotemh: Note that we're changing the go code to use var length for unitName, AssetURL and assetName

PR #34, opened 2019-11-01
Changed selkey and votekey inputs to base64 encoded strings.

ejbaran:


mjiang102628:
Closing: this will be added with the change to transaction params as both are backward incompatible.

PR #39, opened 2019-11-07
add leases to transactions

mjiang102628:
- add group field to asset transaction encode/decode


rotemh:
LGTM but we should probably test it with a golden as well

PR #48, opened 2019-11-19
add Split and HTLC templates

mjiang102628:
- add docs for logic and templates
- issue #42 Add TEAL templates


algosdk/template.py
```diff
@@ -0,0 +1,229 @@
...
+        self.owner = owner
+        self.receiver = receiver
+        self.hash_function = hash_function
+        self.hash_image = hash_image
+        self.expiry_round = expiry_round
+        self.max_fee = max_fee
+
+    def get_program(self):
+        """
+        Return a byte array to be used in LogicSig.
```
rotemh: Same here w.r.t b64

algosdk/transaction.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, sender, fee, first, last, note, gen, gh, lease):
         self.group = None
         self.lease = lease
         if self.lease is not None:
-            if len(self.lease) != 32:
+            if len(self.lease) != constants.lease_length:
```
rotemh: I think that this update belongs to another PR

Issue #55, opened 2019-12-03
Add decimals field to assets


rotemh:
see https://github.com/algorand/go-algorand/pull/601

PR #58, opened 2019-12-05
add decimals to AssetConfigTxn

mjiang102628:
#55 Add decimals field to assets
tests will fail until https://github.com/algorand/algorand-sdk-testing/pull/31 is merged


algosdk/transaction.py
```diff
@@ -418,6 +424,7 @@ def __init__(self, sender, fee, first, last, gh, index=None,
         self.clawback = clawback
         self.url = url
         self.metadata_hash = metadata_hash
+        self.decimals = decimals
```
rotemh: Please add a check that verifies that the number is indeed between 0 and 19.

PR #60, opened 2019-12-12
PeriodicPayment template

mjiang102628:
- #57 Implement Periodic Payment Contract
- fixes bug in Split template


test_unit.py
```diff
@@ -1156,6 +1159,24 @@ def test_HTLC(self):
...
+                       "564BHC7C5CMZB2FVXKGZSIVPJFPT4")
+
+        golden = ("ASAHAQoLAOcHkE7AxAcmAiB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK" +
+                  "3dIAEm2QaSD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiEzEQ" +
+                  "IhIxASMOEDECJBglEhAxBCEEMQIIEhAxBigSEDEJMgMSMQcpEhAxCCEF" +
+                  "EhAxCSkSMQcyAxIQMQIhBg0QMQglEhAREA==")
+        p = s.get_program()
+        self.assertEqual(p, base64.b64decode(golden))
+        self.assertEqual(s.get_address(), golden_addr)
+        sk, pk = account.generate_account()
```
EvanJRichard: what is the generated account used for here?

mjiang102628: Since we don't have a golden for the created transactions, I'm just using this to make sure the get_transactions() function doesn't error.

EvanJRichard: What I mean is, `s.get_withdrawal_transaction(1234, 2234, "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=", 10)` doesn't seem to use `sk` or `pk`

mjiang102628: You're right! My mistake.

algosdk/template.py
```diff
@@ -175,6 +175,69 @@ def get_program(self):
...
+                "MQghBRIQMQkpEjEHMgMSEDECIQYNEDEIJRIQERA=")
+        orig = base64.b64decode(orig)
+        offsets = [4, 5, 7, 8, 9, 12, 15]
+        values = [self.max_fee, self.period, self.withdrawing_window,
+                  self.amount, self.timeout, base64.b64encode(
+                      self.lease_value), self.receiver]
+        types = [int, int, int, int, int, "base64", "address"]
+        return inject(orig, offsets, values, types)
+
+    def get_withdrawal_transaction(self, first_valid, last_valid, gh, fee):
```
rotemh:
```
&&
txn FirstValid
int TMPL_PERIOD
%
int 0
==
&&
txn LastValid
int TMPL_DUR
txn FirstValid
+
==
```


This is from the contract. `FirstValid` must be a multiple of `period` and `LastValid` must be equal to `duration + firstValid` so worth checking that and computing `LastValid` ourselves

algosdk/template.py
```diff
@@ -175,6 +175,72 @@ def get_program(self):
...
+                "MQghBRIQMQkpEjEHMgMSEDECIQYNEDEIJRIQERA=")
+        orig = base64.b64decode(orig)
+        offsets = [4, 5, 7, 8, 9, 12, 15]
+        values = [self.max_fee, self.period, self.withdrawing_window,
+                  self.amount, self.timeout, base64.b64encode(
+                      self.lease_value), self.receiver]
+        types = [int, int, int, int, int, "base64", "address"]
+        return inject(orig, offsets, values, types)
+
+    def get_withdrawal_transaction(self, first_valid, gh, fee):
```
rotemh: Do you expect the withdrawer to re-initiate the contract by him/herself? seems like this function should make a use of `program` or at least make it clear how to use. Note that w/o initiating the contract, the withdrawer doesn't have all the `self.` arguments.

rotemh: What is the solution?

algosdk/template.py
```diff
@@ -175,6 +175,84 @@ def get_program(self):
...
+        orig = base64.b64decode(orig)
+        offsets = [4, 5, 7, 8, 9, 12, 15]
+        values = [self.max_fee, self.period, self.withdrawing_window,
+                  self.amount, self.timeout, base64.b64encode(
+                      self.lease_value), self.receiver]
+        types = [int, int, int, int, int, "base64", "address"]
+        return inject(orig, offsets, values, types)
+
+    @staticmethod
+    def get_withdrawal_transaction(contract, first_valid, gh, fee):
```
winder: The fee argument is never used (overridden below).

test_unit.py
```diff
@@ -1209,6 +1209,26 @@ def test_HTLC(self):
...
+        golden = ("ASAHAQoLAOcHkE7AxAcmAiB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK" +
+                  "3dIAEm2QaSD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiEzEQ" +
+                  "IhIxASMOEDECJBglEhAxBCEEMQIIEhAxBigSEDEJMgMSMQcpEhAxCCEF" +
+                  "EhAxCSkSMQcyAxIQMQIhBg0QMQglEhAREA==")
+        p = s.get_program()
+        self.assertEqual(p, base64.b64decode(golden))
+        self.assertEqual(s.get_address(), golden_addr)
+        ltxn = s.get_withdrawal_transaction(p, 1210, "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=", 10)
+
+        golden_txn = transaction.PaymentTxn(golden_addr, 10, 1210, 2209, "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=", addr, 10000, lease=s.lease_value)
```
winder: Do you have the b64 encoding for this? In the Java client I have:
```
gqRsc2lngaFsxJcBIAcBCgsA5weQTsDEByYCIH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITMRAiEjEBIw4QMQIkGCUSEDEEIQQxAggSEDEGKBIQMQkyAxIxBykSEDEIIQUSEDEJKRIxBzIDEhAxAiEGDRAxCCUSEBEQo3R4bomjYW10zScQo2ZlZQqiZnbNBLqiZ2jEIH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpomx2zQihomx4xCB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2QaaNyY3bEIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITo3NuZMQgQl70jotmWwuFBhLFar+v5Ufqd9wJxfF0TMh0Wt1GzJGkdHlwZaNwYXk=
```

algosdk/template.py
mjiang102628: I'm reading from the contract given by the payer; this is outdated.

algosdk/logic.py
```diff
@@ -23,7 +23,11 @@ def check_program(program, args=None):
     Raises:
         InvalidProgram: on error
     """
+    ok, _, _ = read_program(program, args)
+    return ok
 
+
+def read_program(program, args=None):
```
EvanJRichard: I like this and am pulling it into my open PRs (see team slack discussion). Does `read_program` ever return `False` for the first return value `ok`? I don't see it if so

EvanJRichard: I see the `check program` it's based on does the same thing, "returns true, else throws". Ok I get the context now :+1:

test_unit.py
```diff
@@ -1209,6 +1209,36 @@ def test_HTLC(self):
         self.assertEqual(p, base64.b64decode(golden))
         self.assertEqual(s.get_address(), golden_addr)
 
+    def test_periodic_fee(self):
```
EvanJRichard: nit: test name could maybe be `periodic_pay` rather than `periodic_fee`

EvanJRichard: Does this test pass if it uses the test vector at https://algorand.atlassian.net/wiki/spaces/FDE/pages/445513729/SDK+Templates+integration#Testing-Vectors.3 ?

algosdk/template.py
```diff
@@ -175,6 +175,83 @@ def get_program(self):
...
+            gh (str): genesis hash in base64
+        """
+        address = logic.address(contract)
+        _, ints, bytearrays = logic.read_program(contract)
+        amount = ints[5]
+        withdrawing_window = ints[4]
+        period = ints[2]
+        fee = ints[1]
+        lease_value = bytearrays[0]
+        receiver = encoding.encode_address(bytearrays[1])
```
winder: There's an opportunity to add some validation that `len(ints) == 7` and `len(bytes) == 2` in case a very different contract was provided.

test_unit.py
rotemh: `periodic_payment` :)

mjiang102628: Changed to use that test vector below.

PR #61, opened 2019-12-12
Add LimitOrder template

mjiang102628:
- #56 Implement Limit Order Contract


test_unit.py
```diff
@@ -1156,6 +1159,24 @@ def test_HTLC(self):
...
+        golden = ("ASAKAAEKAsDEBwSQTiSBAotqJgEgkq+RhOQTPAl/ZqvMk7ERGxKiAb2d" +
+                  "DMo+SkihzhPM9MUxFiISMRAjEhAxASQOEDIEIxJAAFUyBCUSMQghBA0Q" +
+                  "MQkyAxIQMwEQIQUSEDMBESEGEhAzARQoEhAzARMyAxIQMwESIQcdNQI1" +
+                  "ATEIIQgdNQQ1AzQBNAMNQAAkNAE0AxI0AjQEDxBAABYAMQkoEjECIQkN" +
+                  "EDEHMgMSEDEIIhIQEA==")
+        p = s.get_program()
+        self.assertEqual(p, base64.b64decode(golden))
+        self.assertEqual(s.get_address(), golden_addr)
+        sk, pk = account.generate_account()
+        s.get_swap_assets_transactions(1000, p, sk, 1234, 2234, "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=", 10)
```
winder: Looks like this is missing an assertion that these transactions are correct

algosdk/template.py
```diff
@@ -175,6 +175,89 @@ def get_program(self):
...
+            first_valid (int): first valid round for the transactions
+            last_valid (int): last valid round for the transactions
+            gh (str): genesis hash in base64
+            fee (int): fee per byte
+        """
+        txn_1 = transaction.PaymentTxn(self.get_address(), fee,
+                                       first_valid, last_valid, gh,
+                                       account.address_from_private_key(
+                                       private_key), int(
+                                           amount/self.ratn*self.ratd))
```
winder: I'm not sure we should compute this for the buyer automatically. We could verify that the amount is correct, but computing it for them could lead to mistakes (i.e. if someone sends them a contract with the wrong ratn/ratd)

PR #67, opened 2020-01-16
fix LimitOrder.get_swap_assets_transactions() implementation

mjiang102628:


algosdk/template.py
```diff
@@ -301,34 +300,58 @@ def get_program(self):
...
         the contract's ratio.
 
         Args:
-            amount (int): the amount of assets to be sent
-            contract (bytes): the contract containg information, should be
-                recieved from payer
+            contract (bytes): the contract containing information, should be
+                received from payer
+            asset_amount (int): the amount of assets to be sent
+            microalgo_amount (int): the amount of algos to be received
```
EvanJRichard: nit: the amount of microalgos maybe

algosdk/template.py
```diff
@@ -301,34 +300,55 @@ def get_program(self):
...
+        asset_id = ints[6]
+        ratn = ints[8]
+        ratd = ints[7]
+        owner = encoding.encode_address(bytearrays[0])
+
+        if microalgo_amount < min_trade:
+            raise error.TemplateError("At least " + str(min_trade) +
+                                      " microalgos must be requested")
+
+        if asset_amount*ratd < microalgo_amount*ratn:
```
winder: I think it needs to be exactly `asset_amount*ratd == microalgo_amount*ratn`

mjiang102628: I don't think it needs to be exact, at least based on https://github.com/algorand/go-algorand/blob/master/tools/teal/templates/limit-order-a.teal.tmpl

test_unit.py
```diff
@@ -1241,46 +1241,44 @@ def test_periodic_payment(self):
...
-                        "eXBlo3BheQ==")
-        golden_txn_2 = ("gqNzaWfEQMvoyBUpU+UgLIRDEOYa7FLhEKUuYWh5esjSqk6ZUm" +
-                        "bSjjGcsv60c18noWTOEIodQ7Rt0gvSqLGCenrnaSs0Nw+jdHhu" +
-                        "iqRhYW10zQPopGFyY3bEIJKvkYTkEzwJf2arzJOxERsSogG9nQ" +
-                        "zKPkpIoc4TzPTFo2ZlZc0JJKJmds0E0qJnaMQgf4OxZX/x/FO5" +
-                        "LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjZ3JwxCAWSrc1gydW6+" +
-                        "0s4EcVQeBrpXZ54Jx6rrHD47ivmMVP4aJsds0IuqNzbmTEIM1F" +
-                        "qEtrTNUb7bkEMhocHV1KggcixRDwUVx2cJ0qgb4GpHR5cGWlYX" +
-                        "hmZXKkeGFpZM0nEA==")
+        [stx_1, stx_2] = s.get_swap_assets_transactions(
```
winder: My only question is, should this return a concatenated byte array like the DynamicFee template?

mjiang102628: Do you mean for the test? DynamicFee doesn't return a byte array in the Python SDK.

winder: This might be a UX question then, I in the other SDKs we format these in a format that could immediately be submitted.

For the test I think it's fine to keep them separate since the goldens can be combined to compare the result either way.

mjiang102628: The function to submit transactions takes transaction objects so it can be immediately submitted.

PR #69, opened 2020-01-30
enable specifying fee for getting transactions, remove precise option since it must be precise

mjiang102628:


mjiang102628:
It technically doesn't in that with those values it would still produce the same transactions; it's just that with precise being mandatory and the amount not being divisible by the ratio, it'll raise an exception.

algosdk/template.py
```diff
@@ -96,17 +94,14 @@ def get_send_funds_transaction(self, amount: int, first_valid, last_valid,
```
winder: You can also verify the `min_pay` field here.

test_unit.py
```diff
@@ -1177,18 +1177,16 @@ def test_split(self):
...
                   "MgMSEDMABykSEDMBByoSEDMACCEFCzMBCCEGCxIQMwAIIQcPEBA=")
-        golden_addr = ("KPYGWKTV7CKMPMTLQRNGMEQRSYTYD" +
-                       "HUOFNV4UDSBDLC44CLIJPQWRTCPBU")
+        golden_addr = "VNG63AMZ7YXA3EJS2LEBIS2BRCTQQ5V4I3GY3ZXXAT4XYBUJKQIVOFDTIM"
         self.assertEqual(s.get_program(), base64.b64decode(golden))
         self.assertEqual(s.get_address(), golden_addr)
-        s.get_send_funds_transaction(1000, 1234, 2234,
+        s.get_send_funds_transaction(1000, 10, 1234, 2234,
                                      "f4OxZX/x/FO5LcGBSKHWXf" +
                                      "wtSx+j1ncoSt3SABJtkGk=")
```
winder: We're all missing golden's for the send funds transaction.

PR #71, opened 2020-02-07
template bug fixes

mjiang102628:
- add HTLC.get_transaction()
- new goldens
- includes changes from michelle/split
- new tests pass

### DIFF
--- a/algosdk/template.py
+++ b/algosdk/template.py
@@ -17,34 +17,36 @@ class Template:
 
 class Split(Template):
     """
-    Split allows locking assets in an account which allows transfering
-    to two predefined addresses in a specific M:N ratio. Note that the ratio is
-    specified by the first address part. For example, if you would like to
-    have a split where the first address receives 30 percent and the second
-    receives 70, set ratn and ratd to 30 and 100, respectively. Split also
-    have an expiry round, in which the owner can transfer back the assets.
+    Split allows locking algos in an account which allows transfering to two
+    predefined addresses in a specified ratio such that for the given ratn and
+    ratd parameters we have:
+        
+        first_recipient_amount * rat_2 == second_recipient_amount * rat_1
+    
+    Split also has an expiry round, after which the owner can transfer back
+    the funds.
 
     Arguments:
-        owner (str): an address that can receive the asset after the expiry
+        owner (str): an address that can receive the funds after the expiry
             round
-        receiver_1 (str): first address to receive assets
-        receiver_2 (str): second address to receive assets
-        ratn (int): the numerator of the first address fraction
-        ratd (int): the denominator of the first address fraction
-        expiry_round (int): the round on which the assets can be transferred
+        receiver_1 (str): first address to receive funds
+        receiver_2 (str): second address to receive funds
+        rat_1 (int): how much receiver_1 receives (proportionally)
+        rat_2 (int): how much receiver_2 receives (proportionally)
+        expiry_round (int): the round on which the funds can be transferred
             back to owner
-        min_pay (int): the minimum number of assets that can be transferred
+        min_pay (int): the minimum number of microalgos that can be transferred
             from the account to receiver_1
         max_fee (int): half the maximum fee that can be paid to the network by
             the account
     """
-    def __init__(self, owner: str, receiver_1: str, receiver_2: str, ratn: int,
-                 ratd: int, expiry_round: int, min_pay: int, max_fee: int):
+    def __init__(self, owner: str, receiver_1: str, receiver_2: str, rat_1: int,
+                 rat_2: int, expiry_round: int, min_pay: int, max_fee: int):
         self.owner = owner
         self.receiver_1 = receiver_1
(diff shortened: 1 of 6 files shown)


algosdk/template.py
```diff
@@ -174,6 +183,40 @@ def get_program(self):
         types = [int, int, "address", "base64", "address", int]
         return inject(orig, offsets, values, types)
 
+    @staticmethod
+    def get_transaction(contract, preimage, first_valid, last_valid, gh, fee):
+        """
+        Return a group transactions array which transfer funds according to
+        the contract's ratio.
```
winder: Wrong comment `Return a transaction which will release funds if a matching preimage is used.`

algosdk/template.py
```diff
@@ -174,6 +183,40 @@ def get_program(self):
...
+        if not (len(ints) == 4 and len(bytearrays) == 3):
+            raise error.WrongContractError("split")
+        max_fee = ints[0]
+        receiver = encoding.encode_address(bytearrays[0])
+
+        lsig = transaction.LogicSig(contract, [base64.b64decode(preimage)])
+        txn = transaction.PaymentTxn(logic.address(contract), fee, first_valid, last_valid, gh, None, 0, close_remainder_to=receiver)
+
+        if txn.fee > max_fee:
+            raise error.TemplateInputError("the transaction fee should not be greater than " + str(max_fee))
```
winder: I think we should be able to verify the hash image as well

test_unit.py
```diff
@@ -1209,6 +1230,21 @@ def test_HTLC(self):
...
+            "gqRsc2lngqNhcmeRxCB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2Qaa" +
+            "FsxJcBIAToBwEAwM8kJgMg5pqWHm8tX3rIZgeSZVK+mCNe0zNjyoiRi7nJOKkV" +
+            "tvkgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGkg/ryguxRKWk6ntD" +
+            "ikaBrIDmyhBby2B/xWUyXJVpX2ohMxASIOMRAjEhAxBzIDEhAxCCQSEDEJKBIt" +
+            "ASkSEDEJKhIxAiUNEBEQo3R4boelY2xvc2XEIOaalh5vLV96yGYHkmVSvpgjXt" +
+            "MzY8qIkYu5yTipFbb5o2ZlZc0D6KJmdgGiZ2jEIH+DsWV/8fxTuS3BgUih1l38" +
+            "LUsfo9Z3KErd0gASbZBpomx2ZKNzbmTEIFNCP4JtpWHGzW5H9EJSwasC1THntU" +
+            "IiTJGurfnrDvCqpHR5cGWjcGF5")
+        ltxn = template.HTLC.get_transaction(
+            p, "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=", 1,
```
winder: I don't think this is a valid input, don't we need to pass in something which, once hashed, is equal to `f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk=`?

algosdk/template.py
```diff
@@ -174,6 +184,56 @@ def get_program(self):
...
+        
+        Returns:
+            LogicSigTransaction: transaction to claim algos from
+                contract account
+        """
+        _, ints, bytearrays = logic.read_program(contract)
+        if not (len(ints) == 4 and len(bytearrays) == 3):
+            raise error.WrongContractError("split")
+        max_fee = ints[0]
+        hash_function = contract[-15]
```
winder: :+1:

algosdk/template.py
```diff
@@ -60,65 +63,72 @@ def get_program(self):
...
-        if amount % ratd == 0:
-            amt_1 = amount // ratd * ratn
-            amt_2 = amount - amt_2
-        elif precise:
-            raise error.NotDivisibleError
-        else:
-            amt_1 = round(amount / ratd * ratn)
+        if amount % (rat_1 + rat_2) == 0:
+            amt_1 = amount // (rat_1 + rat_2) * rat_1
             amt_2 = amount - amt_1
```
winder: This is how I computed the amounts:
```
        Double fraction = Double.valueOf(rat1) / Double.valueOf(rat1+rat2);
        int receiverOneAmount = Long.valueOf(Math.round(fraction * amount)).intValue();
        int receiverTwoAmount = Long.valueOf(Math.round((1.0 - fraction) * amount)).intValue();
```

mjiang102628: I think these are equivalent since amount must be divisible by (rat_1 + rat_2) because I've simplified the fraction first.

winder: I didn't realize `//` did an implicit floor. There might be cases where the number might need to round up after the divide though. I noticed in my code that for our test values `receiverTwoAccount` needed to round up while `receiverOneAmount` needed to round down.

PR #72, opened 2020-02-12
implement suggested params

mjiang102628:
- add SuggestedParams object to handle fee, first valid round, last valid round, and genesis hash
- tests will fail since this is backwards incompatible
    - locally updated tests are passing

### DIFF
--- a/algosdk/transaction.py
+++ b/algosdk/transaction.py
@@ -1069,9 +1069,9 @@ class LogicSig:
 
     def dictify(self):
         od = OrderedDict()
-        od["l"] = self.logic
         if self.args:
             od["arg"] = self.args
+        od["l"] = self.logic
         if self.sig:
             od["sig"] = base64.b64decode(self.sig)
         elif self.msig:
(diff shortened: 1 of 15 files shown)


algosdk/transaction.py
```diff
@@ -10,19 +10,47 @@
 from nacl.exceptions import BadSignatureError
 
 
+class SuggestedParams:
```
EvanJRichard: Could this instead be `Params` or `TransactionParams`? I feel like they're only `suggested` when they come from algod, which isn't always the case. Or phrased another way, I wonder if it will be weird to users to construct a `SuggestedParams` themselves, especially when function signatures refer to them as only coming from algod, or if they'll wonder if their user-set params are only a suggestion.

EvanJRichard: NVM I see this was discussed already :+1: :+1:

rotemh:
What about genesisID l? Seems like we lack support of it. It can be optional but I don’t think we should skip it.

mjiang102628:
genesisID is supported; it's in both the SuggestedParams constructor and the Transaction constructor.

PR #76, opened 2020-03-11
move cucumber tests into python sdk

mjiang102628:


test/cucumber/docker/run_docker.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+rm -rf temp
+rm -rf test/cucumber/features
+git clone --single-branch --branch templates https://github.com/algorand/algorand-sdk-testing.git temp
+
+cp test/cucumber/docker/sdk.py temp/docker
+mv temp/features test/cucumber/features
+
+docker build -t sdk-testing -f test/cucumber/docker/Dockerfile "$(pwd)"
```
winder: The reason my original script used this pattern:
```
rootdir=`dirname $0`
pushd $rootdir
```

instead of `$(pwd)` is because it allows you to run the program from a different directory.

PR #78, opened 2020-03-19
Changed selkey and votekey inputs to base64 encoded strings

mjiang102628:


mjiang102628:
> Looks right to me. Should the `KeyregTxn` constructor args docs around L270-271 note it's a b64 string? "No" is an okay answer, the SDKs pretty much always use b64 strings, it might be inconsistent to call it out here.

Changed; it makes sense to specify it since it's a change from before.

PR #107, opened 2020-05-13
Add checks that the note field is properly formed

fabrice102:
Check that:
* the note field is of type `bytes`.
* the note field is at most 1024-byte long.
Raises exception otherwise.

Without these checks, everything would appear to work but the transaction would be rejected by the node at the end (in the first case because the signature would be invalid).


winder:
@fabrice102 what types appear to work, but cause the transaction to be rejected? A string value also fails?

fabrice102:
@winder The issue is when you use `string` instead of `bytes`. Everything seems to work, but then when sending the transaction, it gets rejected because the signature is invalid.

PR #108, opened 2020-05-14
Indexer health endpoint and integration test.

winder:


test/steps/v2_steps.py
```diff
@@ -255,6 +255,15 @@ def parse_txns_by_addr(context, roundNum, length, idx, sender):
...
         assert context.response["transactions"][int(idx)]["sender"] == sender
 
+@when('I use {indexer} to check the services health')
+def icl_health(context, indexer):
+    context.response = context.icls[indexer].health()
+
+@then('I receive status code {code}')
+def icl_health_check(context, code):
+    # An exception is thrown when the code is not 200
+    assert code == 200
```
EvanJRichard, line 265: I don't quite see how this checks `context.response`, above. It seems to be taking `200` from the `.feature` file and comparing it against literal `200` here, without touching the response received on L260, do I have that right?

winder: An exception would have been thrown in the earlier step.

The status isn't captured anywhere so we can only verify the 200 case.

PR #110, opened 2020-05-19
Rekeying features

EvanJRichard:
##
An upcoming protocol change will allow accounts to change the spending key associated with the public address. This PR proposes to support this feature:
- `Transaction`s can all be made with a `rekey_to` argument, taking a string address; an account issuing a `Transaction` with `rekey_to` set will then need future transactions to be signed by the SK corresponding to `rekey_to`
- `Transaction.sign` will correctly add the `sgnr` field, if the provided SK does not match the `from` address of the `Transaction`
- `stepdefs` added per `algorand-sdk-testing:rotem/rekey`
### Testing
In progress.


run_integration.sh
```diff
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch develop https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch evan/rekey_tester https://github.com/algorand/algorand-sdk-testing.git test-harness
```
winder, line 10: Need to merge in `evan/rekey_tester` and change these back to develop

EvanJRichard: Changed back to `develop`. No need to merge `evan/rekey_tester`, I think, it was a temp branch for modified testing

PR #112, opened 2020-05-21
Update cucumber tests to use a whitelist.

winder:
Run the tests with a tags whitelist. When adding new functionality we will need to update the whitelist.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,10 @@ RUN mkdir -p $HOME/py-algorand-sdk
 COPY . $HOME/py-algorand-sdk
 WORKDIR $HOME/py-algorand-sdk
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -5,9 +5,10 @@ RUN mkdir -p $HOME/py-algorand-sdk
 COPY . $HOME/py-algorand-sdk
 WORKDIR $HOME/py-algorand-sdk
 
+# SDK dependencies, and source version of behave with tag expression support
 RUN pip3 install git+https://github.com/algorand/py-algorand-sdk/ -q \
-    && pip3 install behave -q
+    && pip install git+https://github.com/behave/behave
```
EvanJRichard, line 10: maybe want to leave the `-q` in for `quiet`?

PR #113, opened 2020-05-23
Release 1.3.0

rotemh:
# Add
 - Clients for Indexer and algod API V2


EvanJRichard:
Leaving as a separate comment: we could move the suggestedParams-based transactions out of `future` for this release, but we want to wait for 2.0 since that would break backwards compatibility, right?

PR #120, opened 2020-06-16
Application call support

### LINKED ISSUE #94: Add support for applications to indexer client
- [x] Update client 
- [x] Implement steps for the integration test
- [x] Implement steps for the unit tests
- [x] Implement function to compare response file to object

### LINKED ISSUE #109: Implement Application Transaction support
Implementation 
- [x] Implement support for `Application` Transaction type
- [x] Implement a wrapper for each of the 7 types
- [x] Write an example

Testing 
- [x] Implement cucumber unitTests steps 
- [x] Implement cucumber Integreation steps

EvanJRichard:
# Summary
This PR proposes to add support to future TEAL application calls with a new transaction type, `ApplicationCallTxn` in package `future`. It also adds implementations for the corresponding cucumber unit and integration tests.

## Testing
Tests pass locally. Note that since `run_integration.sh`, the `Makefile`, and the `Dockerfile` target `algorand-sdk-testing:develop` (and so test against the version of `go-algorand` that doesn't yet have Application support), the application tests are not added to the whitelist in the `Makefile`.

## See Also
Closes #94 
Closes #109


test/steps/v2_steps.py
```diff
@@ -711,12 +883,125 @@ def expect_path(context, path):
...
+
+def operation_string_to_enum(operation):
+    if operation == "call":
+        return transaction.OnComplete.NoOpOC
+    elif operation == "create":
+        return transaction.OnComplete.NoOpOC
+    elif operation == "update":
+        return transaction.OnComplete.UpdateApplicationOC
+    elif operation == "optin":
+        return transaction.OnComplete.NoOpOC
```
winder, line 925: optin is 1

test/steps/v2_steps.py
```diff
@@ -711,12 +883,125 @@ def expect_path(context, path):
...
+            approval_program = bytearray(f.read())
+    if clear_program == "none":
+        clear_program = None
+    elif clear_program:
+        with open(dir_path + "/test-harness/features/resources/" + clear_program, "rb") as f:
+            clear_program = bytearray(f.read())
+    if app_args == "none":
+        app_args = None
+    elif app_args:
+        app_args = [bytes(arg, 'ascii') for arg in app_args.split(",")]
```
winder, line 961: Args is a little more complicated than this. I followed the goal format thinking that we might test string/int/address. 

`str:hello` would be `bytes('hello', 'ascii')`

test/steps/v2_steps.py
```diff
@@ -711,12 +883,228 @@ def expect_path(context, path):
...
+
+
+@given('I sign and submit the transaction, saving the txid. If there is an error it is "{error_string}".')
+def sign_submit_save_txid_with_error(context, error_string):
+    signed_app_transaction = context.app_transaction.sign(context.transient_sk)
+    context.app_txid = context.app_acl.send_transaction(signed_app_transaction)
+
+
+@given('I sign and submit the transaction, saving the txid. If there is an error it is "".')
+def step_impl(context):
```
winder, line 1094: you shouldn't need to define this twice

test/steps/v2_steps.py
```diff
@@ -711,12 +883,228 @@ def expect_path(context, path):
...
+    actual_base64 = encoding.msgpack_encode(context.signed_transaction)
+    assert (golden == actual_base64)
+
+
+@given('an algod v2 client connected to "{host}" port {port} with token "{token}"')
+def step_impl(context, host, port, token):
+    algod_address = "http://" + str(host) + ":" + str(port)
+    context.app_acl = algod.AlgodClient(token, algod_address)
+    # also need a v1 client for things like transaction_info
+    context.acl = legacyclient.AlgodClient(token, algod_address)
```
winder, line 1020: context.acl is already initialized by the "Background", no need to re-initialize it here

test/steps/v2_steps.py
```diff
@@ -711,12 +883,228 @@ def expect_path(context, path):
...
+
+@given('I sign and submit the transaction, saving the txid. If there is an error it is "".')
+def step_impl(context):
+    sign_submit_save_txid_with_error(context, "")
+
+
+@given('I wait for the transaction to be confirmed.')
+def step_impl(context):
+    sp = context.app_acl.suggested_params()
+    context.most_recent_pending_info = context.acl.pending_transaction_info(context.app_txid)
```
winder, line 1101: Move this after `status_after_block`. There is no application ID because it is still pending at this point.

EvanJRichard:
I changed the setup back to targeting `algorand-sdk-testing:develop`/latest `go-algorand`. Since that version of `go-algorand` doesn't have app-call support, I haven't added the application tests to the whitelist. But, I verified they do pass locally.

algosdk/future/transaction.py
```diff
@@ -87,20 +88,22 @@ def get_txid(self):
         txid = base64.b32encode(txid).decode()
         return encoding._undo_padding(txid)
 
-    def sign(self, private_key):
+    def sign(self, private_key, skip_rekey_check=False):
```
winder, line 91: Can we implement this without adding a public parameter? Maybe making the Signed transaction directly instead of pretending to sign it with a random key?

PR #123, opened 2020-06-23
[Found Work] fix incorrect size estimation exposed by test_unit.py

EvanJRichard:
## Summary
Rekey-to transaction support #110 changed size estimation. This caused slightly different values for fee calculation, and in turn, broke the golden comparison checks in `test_unit.py`. This PR proposes to change size estimation back to the legacy way, while maintaining the new automatic-auth-address detection feature of `sign`. 

## Testing
Manually ran `python3 test_unit.py` locally; allowed Travis to run rest of tests.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -82,12 +82,14 @@ class Transaction:
         txid = base64.b32encode(txid).decode()
         return encoding._undo_padding(txid)
 
-    def sign(self, private_key):
+    def sign(self, private_key, skip_rekey_check=False):
         """
         Sign the transaction with a private key.
 
         Args:
             private_key (str): the private key of the signing account
+            skip_rekey_check (bool, optional): if set to True, do not handle authorizing_address logic
+                (typically used for size estimation)
 
         Returns:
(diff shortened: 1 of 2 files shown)


algosdk/future/transaction.py
```diff
@@ -82,20 +82,22 @@ def get_txid(self):
         txid = base64.b32encode(txid).decode()
         return encoding._undo_padding(txid)
 
-    def sign(self, private_key):
+    def sign(self, private_key, skip_rekey_check=False):
```
algorotem, line 85: Do we expect the user to know when to use it?

EvanJRichard: I expect the user to ignore it, which is why it's defaulted to `False`. The function comment notes that `True` is `typically used for size estimation`. I could change it to do a separate, private `_sign_and_skip_rekey_check` instead, so it's not exposed to the user.

algorotem: If the user doesn't need to know about it. Let's make it private.

EvanJRichard:
Fix also implemented in reviewed and merged PR #120. PR is obsolete, closing.

PR #125, opened 2020-07-02
Fix test_integration

algorandskiy:


algosdk/future/transaction.py
```diff
@@ -1860,6 +1860,8 @@ def retrieve_from_file(path):
             txns.append(LogicSigTransaction.undictify(txn))
         elif "type" in txn:
             txns.append(Transaction.undictify(txn))
+        elif "txn" in txn:
+            txns.append(Transaction.undictify(txn['txn']))
```
algorandskiy, line 1864: same as `algosdk/encoding.py`, here is also lots of missed cases compare to the original one

algorandskiy:
The build failed with
```
@when(u'the parsed Dryrun Response should have global delta "Creator" with 2')
def step_impl(context):
    raise NotImplementedError(u'STEP: When the parsed Dryrun Response should have global delta "Creator" with 2')
make: *** [Makefile:2: unit] Error 1
Makefile:8: recipe for target 'docker-test' failed
```
that is non-related to this change, and I work on dryrun now.
@EvanJRichard

PR #139, opened 2020-08-07
[Draft] Evan/sort indexer dict

EvanJRichard:


test/steps/v2_steps.py
```diff
@@ -925,11 +925,30 @@ def step_impl(context, jsonfile):
...
         loaded_response = bytearray(f.read())
-    if context.response != json.loads(loaded_response):
+    # sort context.response
+    def recursively_sort_on_key(dictionary):
+        returned_dict = dict()
+        for k, v in sorted(dictionary.items()):
+            if isinstance(v, dict):
+                returned_dict[k] = recursively_sort_on_key(v)
+            elif isinstance(v, list) and all(isinstance(item, dict) for item in v):
+                if all('key' in item.keys() for item in v):
```
jasonpaulos, line 935: Nit: I think it's sufficient to use `'key' in item` instead of `'key' in item.keys()` here.

Issue #115, opened 2020-05-28
Issue when the receiver is the 0 address


fabrice102:
When the receiver is the 0 address (`AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ`), the SDK does not sign correctly the transaction. It also does not compute correctly the group ID of a group containing such a transaction.

A 0 receiver address is useful to close an account. Some TEAL script may also require a 0 receiver address.

Minimum example assuming `$ALGORAND_DATA` is set up properly (the transaction gets rejected because the signature is invalid):

```python
from os import environ
import algosdk

data_dir_path = environ.get("ALGORAND_DATA")
if not data_dir_path[-1] == "/":
    data_dir_path += "/"
algod_token = open(data_dir_path + "algod.token", "r").read().strip("\n")
algod_address = "http://" + open(data_dir_path + "algod.net",
                                    "r").read().strip("\n")

sender_sk = algosdk.mnemonic.to_private_key("various whip fetch bonus explain leg bean imitate mechanic lesson rely swift very autumn river husband lawsuit belt slim rabbit menu thought elephant able shaft")
sender_pk = "4TYPVYXMC5WNSXQHK7UCIMSMV7VOBBJD4L7T6FYJXEHB4XUYKSGF4ALFRY"

algod_client = algosdk.algod.AlgodClient(algod_token, algod_address)

params = algod_client.suggested_params()
first_valid = params.get("lastRound")
last_valid = first_valid + 1000
gh = params.get("genesishashb64")
min_fee = params.get("minFee")

data = {
    "sender": sender_pk,
    "receiver": 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ',
    "fee": min_fee,
    "first": first_valid,
    "last": last_valid,
    "gh": gh,
    "amt": 0,
    "flat_fee": True,
    "close_remainder_to": "OEOFA6R5HQA7KBQ5BGMI2B5FQ2U4S6TWF5DUEHWJ4PSYOFZ5A7LGXSM4TM"
}
tx = algosdk.transaction.PaymentTxn(**data)

stx = tx.sign(sender_sk)

algod_client.send_transaction(stx)
```

With goal, there is no such issue:

```
[...]

PR #148, opened 2020-09-23
check for zero receiver in transaction

mjiang102628:
- add unit test for zero receiver
- issue #115


EvanJRichard (approved):
It's illegal to have a zero receiver, right? If so LGTM

EvanJRichard:
Update, it's legal at the protocol level to have a zero receiver (such as for a close-to transaction), see https://github.com/algorand/js-algorand-sdk/issues/175 . I think this change might still be useful, though, so letting approval stand.

EvanJRichard (changes requested):
After speaking with @algorotem in standup, we might not want this change, or maybe we want to allow the zero receiver but only if the user-dev explicitly calls for it, or something.

EvanJRichard (approved):

jasonpaulos (approved):
Looks good, but can you add unit tests to ensure errors will be raised if the receiver is None for `PaymentTxn` and `AssetTransferTxn`?

jasonpaulos:
LGTM 👍

Issue #111, opened 2020-05-20
Creating New Wallet error -


purestaketdb:
As brought up by a community member, with the removal of the default wallet from the standard Algorand install, creating a new wallet with the Python SDK when no other wallets exist will result in an error and fail. 

https://forum.algorand.org/t/python-sdk-error-creating-a-new-wallet/1656/5

Essentially the API sends back an empty dict {} but the SDK expects:

`{
  "wallets": []
}`

It is not clear if this should be addressed in the Python SDK to handle the empty case or in the Algod API to return the expected message.

fabrice102:
Another person stumbled upon a similar issue: https://forum.algorand.org/t/python-kmdclient-list-wallets-responds-with-wallets/1893/3

PR #149, opened 2020-09-29
check for empty result in list_wallets

mjiang102628:
#111 fixes error where creating a new wallet expects something different from the node

### DIFF
--- a/algosdk/kmd.py
+++ b/algosdk/kmd.py
@@ -87,7 +87,11 @@ class KMDClient:
             dict[]: list of dictionaries containing wallet information
         """
         req = "/wallets"
-        return self.kmd_request("GET", req)["wallets"]
+        res = self.kmd_request("GET", req)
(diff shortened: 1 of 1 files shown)


EvanJRichard (approved):

algosdk/kmd.py
```diff
@@ -87,7 +87,11 @@ def list_wallets(self):
             dict[]: list of dictionaries containing wallet information
         """
         req = "/wallets"
-        return self.kmd_request("GET", req)["wallets"]
+        res = self.kmd_request("GET", req)
+        if "wallets" in res:
+            return res["wallets"]
+        else:
+            return res
```
jasonpaulos, line 94: It's probably better to return an empty list if there are no wallets.

jasonpaulos (changes requested):

jasonpaulos (approved):

Issue #127, opened 2020-07-07
Inconsistency in argument naming: `block` vs `round`


fabrice102:
The Python SDK is the only one using the name `block` instead of `round` for the arguments of the API calls of the indexer: https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/v2client/indexer.py

This is inconsistent with the other SDK, the REST API, and with the Python `min_round` / `max_round` arguments. For the last one, see https://github.com/algorand/py-algorand-sdk/blob/ec7f143e4b272110848d2bc3adcc0ac119bbd081/algosdk/v2client/indexer.py#L371-L375

Proposed change: allow to use either `block` or `round` to not break backward compatibility.

PR #150, opened 2020-09-30
add optional round_num argument to v2 clients for 'round' vs 'block' naming consistency

mjiang102628:
- add round_num variable to functions using block or round, since block is inconsistent with other variable names and round is a reserved function in python

#127

### DIFF
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -117,17 +117,21 @@ class AlgodClient:
             **kwargs)
         return res
 
-    def block_info(self, block, response_format="json", **kwargs):
+    def block_info(self, block=None, response_format="json", round=None, **kwargs):
         """
         Get the block for the given round.
 
         Args:
             block (int): block number
-            response_format (str): the format in which the response is returned: either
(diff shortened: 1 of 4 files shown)


EvanJRichard:
I like this change, pending tests passing :+1:

algosdk/v2client/algod.py
```diff
@@ -117,17 +117,22 @@ def pending_transactions_by_address(self, address, limit=0, response_format="jso
...
-            response_format (str): the format in which the response is returned: either
-                "json" or "msgpack"
+            response_format (str): the format in which the response is 
+                returned: either "json" or "msgpack"
+            round (int, optional): alias for block; only specify one of these
         """
         query = {"format": response_format}
-        req = "/blocks/" + str(block)
+        req = "/blocks/"
+        if block is not None:
```
fabrice102, line 132: Maybe this should raise an exception if both `block` and `round` are provided?
Maybe this should be a function as it's used several time?

jasonpaulos (approved):
LGTM

Issue #156, opened 2020-10-26
add wait_for_confirmation helper function


mjiang102628:
algorand/go-algorand-internal#730

PR #158, opened 2020-10-27
add wait_for_confirmation() helper function

mjiang102628:
#156

### DIFF
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -191,7 +191,7 @@ class AlgodClient:
     def wait_for_confirmation(self, transaction_id, timeout):
         """
         Wait until the transaction is confirmed or rejected, or until 'timeout'
-        number of rounds have passed.
+        number of rounds have passed. 
 
         Args:
             transaction_id (str): the transaction to wait for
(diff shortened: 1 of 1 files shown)


rfustino:
Might be nice to add a print line with the confirmed round

```        if pending_txn.get("confirmed-round", 0) > 0:
            print("txID: {}".format(transaction_id), " confirmed in round: {}".format(
                pending_txn.get("confirmed-round", 0)))
            return pending_txn```

File: algosdk/v2client/algod.py [hunk 1]
```diff
@@ -188,6 +188,34 @@ def send_raw_transaction(self, txn, **kwargs):
...
+
+        Raises:
+            AlgodHTTPError: if the transaction is not found
+        """
+        start_round = self.status()["next-version-round"]
+        current_round = start_round
+
+        while current_round < start_round + timeout:
+            self.status_after_block(current_round)
+            pending_txn = self.pending_transaction_info(transaction_id)
```
winder, line 212: Some customers may be running algod behind a load balancer. In that situation, `pending_transaction_info` might throw an exception because the transaction_id isn't in the transaction pool but we should continue waiting until the timeout.

PR #161, opened 2020-11-12
Add init file to algosdk.testing

### LINKED ISSUE #160: Can't import algosdk.testing module
I am trying to run the example from https://github.com/algorand/docs/pull/166, but it fails on the following line:

```python
from algosdk.testing.dryrun import DryrunTestCaseMixin, Helper
```

Error: `ModuleNotFoundError: No module named 'algosdk.testing'`

I think the algosdk/testing folder may need a `__init__.py` file.

## Urgency/Priority
Pretty urgent because the TEAL testing suite doesn't work if we can't import it (this bug)

jasonpaulos:
Add an \_\_init\_\_.py file to algosdk/testing, which makes that directory a proper Python module.

From https://docs.python.org/3/tutorial/modules.html:

> The \_\_init\_\_.py files are required to make Python treat directories containing the file as packages. This prevents directories with a common name, such as string, unintentionally hiding valid modules that occur later on the module search path. In the simplest case, \_\_init\_\_.py can just be an empty file, but it can also execute initialization code for the package or set the \_\_all\_\_ variable, described later.

Closes #160.


EvanJRichard (approved):
Possibly the simplest code review of my life. :shipit:

PR #163, opened 2020-11-20
Convert APP and ASA indexes to ints.

jannotti:
Previously, code like this:
  transaction.ApplicationNoOpTxn(bidder, params, "10382")

would silently continue, but after signing and submitting, would
report:
algosdk.error.AlgodHTTPError: msgpack decode error [pos 93]: cannot decode unsigned integer: unrecognized descriptor byte: a8/string|bytes

This is a pretty obscure error to get for using a string id instead of
an int. We could type check and give an error when creating the
transaction, but Python APIs are usually more forgiving.  This change
calls int() on the index argument, so it will try to convert strings
if needed.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -497,7 +497,7 @@ class AssetConfigTxn(Transaction):
         if strict_empty_address_check:
             if not (manager and reserve and freeze and clawback):
                 raise error.EmptyAddressError
-        self.index = index
+        self.index = int(index)
         self.total = total
         self.default_frozen = default_frozen
         self.unit_name = unit_name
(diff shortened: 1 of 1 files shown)


jasonpaulos:
I think I might be in favor of raising an error if the argument is not an int since it'll be simpler to support moving forward (unless we already have examples which use strings). If the error is kindly worded, I don't think developers will have any issue performing the cast themselves.

jasonpaulos:
@EvanJRichard any thoughts here?

jannotti:
I'm pretty strongly in favor of this kind of change, I have several more queued up, in fact, so let's discuss.  We also fail in a pretty bad way when strings are provided instead of 'bytes' as app args.  (The failure goes unreported, but later the signature fails to verify when the txn is submitted.)

My feeling is that a) This is the Python way.  APIs in python are never so picky. b) There's no advantage to us making our developer's lives more tedious.  There's no hidden gotcha we're trying to protect people from.  They might reasonably be expected to pick up the app-id from an environment variable, which is a string.  That works for addresses (because we don't treat that like a separate data type), but not app IDs. Same thing for app args (and converting them to bytes is even more tedious).

EvanJRichard:
I think it's a good change, I don't see any scenario where the changed code would do anything astonishing. 
User provides a string -> converted to int just fine
User provides an int -> conversion is a no-op, also fine

jasonpaulos (approved):
Thanks for explaining, this seems like a good change then.

PR #164, opened 2020-11-30
Be forgiving about typing of app_args, foreign_a*

jannotti:
Allow callers to specify args as strings, not just bytes, and
foreign_apps and assets as strings instead of just ints.  This makes
it more convenient to pull these things from environment variables.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -959,15 +959,35 @@ class ApplicationCallTxn(Transaction):
         self.global_schema = global_schema
         self.approval_program = approval_program
         self.clear_program = clear_program
-        self.app_args = app_args
+        self.app_args = self.bytes_list(app_args)
         self.accounts = accounts
-        self.foreign_apps = foreign_apps
-        self.foreign_assets = foreign_assets
+        self.foreign_apps = self.int_list(foreign_apps)
+        self.foreign_assets = self.int_list(foreign_assets)
         if sp.flat_fee:
             self.fee = max(constants.min_txn_fee, self.fee)
         else:
             self.fee = max(self.estimate_size() * self.fee,
                            constants.min_txn_fee)
+
+
+    @staticmethod
+    def bytes_list(lst):
+        def as_bytes(e):
+            if isinstance(e, bytes):
+                return e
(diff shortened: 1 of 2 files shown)


algosdk/future/transaction.py
```diff
@@ -959,16 +959,36 @@ def __init__(self, sender, sp, index,
...
         if sp.flat_fee:
             self.fee = max(constants.min_txn_fee, self.fee)
         else:
             self.fee = max(self.estimate_size() * self.fee,
                            constants.min_txn_fee)
 
+
+    @staticmethod
+    def bytes_list(lst):
+        def as_bytes(e):
```
jasonpaulos, line 975: I'm wondering if this function should accept ints too and encoding them to bytes with [to_bytes](https://docs.python.org/3.2/library/stdtypes.html#int.to_bytes).

jannotti: I considered it, but I decided against because I didn't/don't think that the encoding scheme for integers sent to contracts is well defined.  Maybe they are encoded as varuints, or maybe as 4 byte or 8 byte (or more), or maybe as ascii bytes that represent the number, "123" - depends on what the TEAL does with the arg.  If that's right, then I don't think we should make the decision here.  

I suppose there is an argument to be made for encoding ints in whichever way it's easiest to decode them in TEAL, even if it isn't universal.  I guess that would be:
```
arg n
btoi
```
and that would argue for encoding the int as 8 bytes, big endian.

jannotti: I'm going to give this more thought. I notice that goal does have a syntax for passing an int as an app argument. So I guess in that instance we decided that picking a format and encoding it that way was the right thing to do. That makes me think it would be reasonable to do the same thing here. So I'll probably add that before merging. Thanks for bringing it up.

jasonpaulos: Great, I agree 8 byte big endian seems like the way to go. If users have a different encoding they'd rather use they could still encode to bytes before this function gets invoked.

jasonpaulos (approved):
Looks good, thanks for adding tests.

winder (approved):

PR #165, opened 2020-11-30
Install local package in dockerfile

jasonpaulos:
Changed the dockerfile to install the local version of the SDK. This will make testing local changes a lot easier.

Also updated the python base image.


winder (approved):
LGTM

PR #166, opened 2020-12-08
Robust mnemonic

jannotti:
Accept mnemonic words without regard to leading/trailing whitespace,
or the type of whitespace (we used to require exactly one space (no
newlines).  Words also work even if we only have the first for (or
more) letters, since they are guarenteed unique starting there, and
some key storage devices (crytosteel) explicitly tell you only to use
the first four letters.

This also speeds things up because we store the words in a
dict instead of a list.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -225,6 +225,9 @@ class Transaction:
                 self.type == other.type and
                 self.rekey_to == other.rekey_to)
 
+    def __repr__(self):
+        return repr(self.dictify())
+
 
 class PaymentTxn(Transaction):
     """
--- a/algosdk/mnemonic.py
+++ b/algosdk/mnemonic.py
@@ -6,7 +6,17 @@ from . import constants
 from . import encoding
 
 
-word_list = wordlist.word_list_raw().split("\n")
+word_to_index = {}
+index_to_word = {}
+for i, word in enumerate(wordlist.word_list_raw().split("\n")):
+    index_to_word[i] = word
+    # Put all prefixes of words at least four letters long into map,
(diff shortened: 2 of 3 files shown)


jasonpaulos (commented):
This looks like a good change, I just have one small concern about what happens when a user tries to decode a mnemonic with an invalid word. In the old implementation a `ValueError` gets raised, but here it looks like a `KeyError` gets raised. I know both of these errors are byproducts of their implementations, but I think `ValueError` makes more sense in this situation and avoids changing the interface.

jannotti:
I was trying to decide whether we need to maintain that compatibility. I
managed to convince myself that both types "made sense" (a bad word can be
seen as either type of error). Since our API didn't document the type of
error, I felt compatibility was not an issue. I don't feel strongly about
it, though.

How do we decide what's a breaking change? (All changes can break some
hypothetical program that was, for example, expecting an error for leading
whitespace, but we think such code is non existent) Do we think people are
dependant on the error type?

jasonpaulos:
Good question, @mjiang102628 do you think that changing the error type in this case is significant enough to be breaking?

mjiang102628:
I think it makes sense not to change it, at least for now in the case that someone is depending on the error type; if KeyError is more fitting then we can change it in the 2.0 version. Also, @jannotti could you still keep a wordlist checksum test? Just to make sure nobody accidentally does a replace all or something like that and changes the word list.

jannotti:
Sure, I'll convert the exception and do something to check the word list.
Thanks.

jasonpaulos:
I think this looks good. @jannotti could you rebase/merge this with develop? I can approve after that.

PR #167, opened 2020-12-14
Asset txns

jannotti:


algosdk/future/transaction.py
```diff
@@ -626,7 +646,8 @@ def __eq__(self, other):
         return (super(AssetConfigTxn, self).__eq__(other) and
                 self.index == other.index and
                 self.total == other.total and
-                self.default_frozen == other.default_frozen and
+                # coerce default_frozen so that None == False
```
mjiang102628, line 649: Can you also change these in dictify()? So that when encoding we also force default_frozen to be bool.

mjiang102628 (changes requested):

algosdk/future/transaction.py
jannotti: Sure. Actually I think if I just change the constructor and coerce it to a bool there, we'll always use a bool internally and we won't need the extra work in `__eq__`

PR #171, opened 2021-01-26
Pack "empty" schema properly, detect the sending of unsigned transactions.

### LINKED ISSUE #168: Python SDK should detect the meaningless schema and not msgpack it.
If schemas are defined as 0s a signature validation error occurs. This should be checked as described below:

These tests:
        if self.local_schema:
            d["apls"] = self.local_schema.dictify()
        if self.global_schema:
            d["apgs"] = self.global_schema.dictify()
Should be something like if self.schema and not self.schema.empty():
(with empty() defined to return True if both ints and slices are 0)

jannotti:
Reporting errors more clearly, we now detect and report several different typing errors at txn construction time, and fix #168 where an "empty" StateSchema is msgpacked wrong. (It shouldn't be packed at all.)

We also report attempts to send unsigned transactions.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -1151,8 +1151,8 @@ class ApplicationCallTxn(Transaction):
     @staticmethod
     def state_schema(schema):
         """Confirm the argument is a StateSchema, or false which is coerced to None"""
-        if not schema:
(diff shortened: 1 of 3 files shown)


jasonpaulos (approved):
Looks good, and I added a quick unit test for empty state schema.

PR #173, opened 2021-01-27
Do not encode an empty lease

bantalon:


jannotti (commented):
This is very reasonable and we probably would have merged it if we knew it was coming.  But I also fixed it and added a unit test.  The fix I went with is slightly more in line with the way we do similar checking on state_schemas, app_args, and so on.

bantalon:
Cool. I was not  aware of the unit tests since the checks for the PR were green. Had a look now and saw them. I will have a look at your implementation once the PR  is ready.
Let me know if you change your mind and want me to add the unit test to this PR.

bantalon:
Resolved in #174 using new and more generic implementation.

PR #174, opened 2021-01-29
Do not msgpack empty leases.

jannotti:
A lease of 32 zeros should not be msgpacked, or we get signature
errors, as the Go code does not pack it.  Check and remove such leases
at Transaction construction time.

### DIFF
--- a/algosdk/constants.py
+++ b/algosdk/constants.py
@@ -55,6 +55,8 @@ logic_data_prefix = b"ProgData"
 """bytes: program (logic) data prefix when signing"""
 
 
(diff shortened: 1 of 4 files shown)


jasonpaulos:
@jannotti thanks for working on this. Can you also extend this to work for the `metadata_hash` field of `AssetConfigTxn`? It's just like lease (32 byes, should omit if all zeros).

jannotti:
Sure. I meant to look into that and check metadata rules.

jasonpaulos (approved):
Looks good!

PR #175, opened 2021-02-06
Add support for logic sig on rekeyed accounts

fabrice102:
Before, logic sig could only be used on accounts that were not rekeyed.


jasonpaulos:
Closing this in favor of #216

Issue #176, opened 2021-02-10
The link to "install a node" in the readme is broken


loibucket:
https://developer.algorand.org/docs/introduction-installing-node

403 Forbidden
Code: AccessDenied
Message: Access Denied
RequestId: 3110BCA925B8012C
HostId: +fXfOnwGC41G4tgtjDkHA7nkdyIK/vyuiekR32aseYGLCDwCCjn9TY0rA0BKVh3c/gRW2wChEpg=
An Error Occurred While Attempting to Retrieve a Custom Error Document
Code: AccessDenied
Message: Access Denied

jasonpaulos:
Fixed by #178

PR #178, opened 2021-02-15
Updated Node setup section with good link to setup a node

KHendricks:
Created in response to #176


jasonpaulos (approved):
Thanks for fixing!

PR #182, opened 2021-03-10
Support for offline and non-participating key registration transactions

fabrice102:
Before, only online key registrations were supported.

Note: unit testing only test for `msgpack_decode` but not `future_msgpack_decode`.
However, I only made the change for "future" transactions.
I added some comments in `test_unit.py`.


jasonpaulos (approved):
This looks good, thanks for implementing and testing!

PR #183, opened 2021-03-12
Support new features for indexer 2.3.2

### LINKED ISSUE #180: Changes for updating indexer 2.3.2 release
There are some changes we need to validate for the upcoming indexer release. These include:

- Cumulative rewards are now computed and stored in the account "Rewards" field.
- Round at which things are added or removed from the database are now reported. There is also a 'deleted' flag for each of these indicating whether the record is currently deleted:
  - account: 'created-at-round', 'closed-at-round'
  - asset: 'created-at-round', 'destroyed-at-round'
  - asset holding: 'opted-in-at-round', 'opted-out-at-round'
  - application: 'created-at-round', 'deleted-at-round'
  - account application: 'opted-in-at-round', 'closed-out-at-round'

There may be additional changes related to TEAL updates in an upcoming go-algorand release as well.

Includes cucumber testing

jasonpaulos:
Add support for the new `include-all` query param on some indexer endpoints, and add support for relevant cucumber tests.

Like the JS SDK (algorand/js-algorand-sdk#296), the application endpoints also contained query params for `round`. I chose not to remove these arguments completely since that would cause a breaking change for anyone who calls the functions normally (i.e. without keyword args), so instead I've changed their descriptions to discourage use.

Closes #180.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 unit:
-	behave --tags="@unit.offline or @unit.algod or @unit.indexer or @unit.rekey or @unit.tealsign or @unit.dryrun or @unit.responses" test -f progress2
+	behave --tags="@unit.offline or @unit.algod or @unit.indexer or @unit.rekey or @unit.tealsign or @unit.dryrun or @unit.responses or @unit.responses.231" test -f progress2
 
 integration:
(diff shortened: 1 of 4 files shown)


winder (commented):
Looks good, just missing the new endpoints

winder (approved):
LGTM

PR #190, opened 2021-04-12
Add type check to AssetTransferTxn index

sigsalerno:
If index is passed as a string, a hardly understandable error 400 is passed by the API with error `msgpack decode error [pos 486]: cannot decode unsigned integer: unrecognized descriptor byte: a8/string|bytes`
This change raise an error in case the type of index variable is not an int


jasonpaulos (changes requested):
Thanks for contributing! I'd be happy to accept this with the following changes:
* Can you make a new error type for this in `algosdk/error.py`, similar to `WrongAmountType`?
* Can you add the same check to `algosdk/future/transaction.py`?

PR #196, opened 2021-06-01
"Rent" larger program

shiqizng:
This PR adds a parameter, extra-pages, to ApplicationCreateTxn(). This allows user to request extra space for application programs.


algosdk/future/transaction.py
```diff
@@ -1158,6 +1305,7 @@ class ApplicationCallTxn(Transaction):
         accounts (list[string], optional): list of additional accounts involved in call
         foreign_apps (list[int], optional): list of other applications (identified by index) involved in call
         foreign_assets (list[int], optional): list of assets involved in call
+        extra_pages (int, optional): additional program space for supporting larger TEAL assembly program. A maximum of 3 extra pages is allowed. A page is 1024 bytes.
```
jannotti, line 1308: I'd prefer we say "larger programs", and avoid "TEAL assembly".

Also, let's keep the limits out of this, so we don't have multiple places to update when they change. (For example, the foreign_apps and assets don't say what the limit is.)

jannotti (changes requested):
Until / unless we can all agree on a standard code formatter, can we all turn them off for python code?  It's hard to find the real changes here because there are a lot of reformatting changes, and I imagine someone else's reformatter is going to switch them back.

I eventually got through the transaction.py changes, but the testing changes are just too long to dig through because of all of the automatic changes.

Once we agree on a standard format, we could perhaps run it it on all files at once, in one PR, and then refuse changes that violate it.

shiqizng:
updated cucumber tests, 
https://github.com/algorand/algorand-sdk-testing/commit/c37a172bd692e24653514d090014c4edebf8b745
https://github.com/algorand/algorand-sdk-testing/commit/a06d2dc8757d81a028bb696bf61a88805eff6574

algosdk/future/transaction.py
```diff
@@ -1175,13 +1176,14 @@ class ApplicationCallTxn(Transaction):
...
         foreign_assets (list[int])
+        extra_pages (int)
     """
 
     def __init__(self, sender, sp, index,
                  on_complete, local_schema=None, global_schema=None,
                  approval_program=None, clear_program=None, app_args=None,
                  accounts=None, foreign_apps=None, foreign_assets=None,
-                 note=None, lease=None, rekey_to=None):
+                 note=None, lease=None, rekey_to=None, extra_pages=0):
```
jdtzmn, line 1186: nit: can you move extra_pages to be before `note`, `lease`, and `rekey_to`. It is convention for those three to appear last in the parameter list.

jdtzmn (changes requested):
I added one comment and I also second JJ's request for un-linting the `v2_steps.py` file. Everything else looks great.

jdtzmn:
I seem to recall Jason mentioning that the `future` directory isn't really used anymore. I'm going to look back at recent commits to see if that checks out. @jannotti do you know whether that's the case?

algosdk/future/transaction.py
jannotti: I don't think we can change the order of arguments.  It would silently break existing code.  Unless we introduced a * early in the list, which would require the use of keyword args.  Then the code would break, but not silently.

jannotti:
> ~I seem to recall Jason mentioning that the `future` directory isn't really used anymore. I'm going to look back at recent commits to see if that checks out. @jannotti do you know whether that's the case?~
> 
> Edit: `future/transaction.py` has had more recent updates than the regular `transaction.py` file so my recollection was incorrect.

Yes, my understanding is that future is where the real action is.  I suspect the intent was to have the code in future replace the other code.  Maybe we can do that in a major revision numbered release.

algosdk/future/transaction.py
jdtzmn: Ah yes, I forgot about that. Well if this SDK requires putting the extra pages field last, we should probably put it last across all the SDKs to be consistent.

jasonpaulos (approved):
This looks good to me.

jannotti (approved):

jdtzmn (approved):

PR #198, opened 2021-06-09
update extraPage description

shiqizng:


jdtzmn (approved):
Looks good! Consider adding some info about how the 2kb is shared to be more consistent with [the message](https://github.com/algorand/js-algorand-sdk/pull/369/files#diff-ed81f9fbb03dd87e910584dd63db8543bb94d2f9d73fe04091a252b68af02381R373) in Ayush's PR; I'm not sure I'm sold yet on sacrificing conciseness for the extra info, but thought I would at least mention it.

PR #203, opened 2021-06-16
install steps addon

figurestudios:
+ inconsistency (python > Python)
+ inconsistency (`make docker-test` > `make docker-test`.)
alternatively it can be removed from the previous line^^
+ requirements for pip installations (for newer users)
+ requirement / pyenv mention

I understand that most people don't need this, but at the same time, newer people who get into Python development for bounties might find them useful instead of giving up when they can't even install the packages. Trust me, I've been there LOL


algoanne:
sorry this fell through the cracks!

Issue #179, opened 2021-02-16
Better sanity checks when creating/encoding transactions


fabrice102:
Currently, if arguments of transactions are of the wrong type, most of the time the SDK will not throw any exception and will even allow to sign and send the transactions.
But then, the node will reject the transaction with a relatively cryptic error message.

For example, if you make the asset ID a string:
```py
from algosdk import mnemonic
from algosdk.v2client import algod
from algosdk.future.transaction import AssetTransferTxn

passphrase = "patrol dog clean ancient negative despair pulp right false figure faint ethics film garment absorb gold capable busy juice tape arrange smooth twin able supply"
# address = TD5MTOHW7LE7RVIQW52WNQDIHUIJFLKTIAB5BGYLB3ZYHQAN4NU3MYNC4M
pk = mnemonic.to_public_key(passphrase) 
sk = mnemonic.to_private_key(passphrase)

algod_client = algod.AlgodClient(algod_token='', algod_address='https://api.testnet.algoexplorer.io', headers={ 'User-Agent': 'DoYouLoveMe?' })

params = algod_client.suggested_params()

asset_id = "14075399" # error: asset_id should be an int, not a string
txn = AssetTransferTxn(sender=pk, sp=params, receiver=pk, amt=0, index=asset_id)
stxn = txn.sign(sk)

txid = algod_client.send_transaction(stxn)
```
you get the following error:
```plain
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/algosdk/v2client/algod.py", line 72, in algod_request
    resp = urlopen(req)
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/urllib/request.py", line 214, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/urllib/request.py", line 523, in open
    response = meth(req, response)
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/urllib/request.py", line 632, in http_response
    response = self.parent.error(
  File "/usr/local/Cellar/python@3.9/3.9.1_8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/urllib/request.py", line 561, in error
    return self._call_chain(*args)
[...]

PR #204, opened 2021-06-17
feat: runtime type checker created

Niraj-Kamdar:
relates: #179 

Same method can be used for `Transaction` class to perform typecheck
Let me know if you like the solution and would want me to make the other classes type safe.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -1,18 +1,16 @@
 import base64
-from enum import IntEnum
-import msgpack
 from collections import OrderedDict
-from .. import account
-from .. import constants
-from .. import encoding
-from .. import error
-from .. import logic
-from .. import transaction
-from nacl.signing import SigningKey, VerifyKey
+from enum import IntEnum
+
+import msgpack
 from nacl.exceptions import BadSignatureError
+from nacl.signing import SigningKey, VerifyKey
+
+from .. import account, constants, encoding, error, logic, transaction
+from ..util import TypeCheck
 
 
-class SuggestedParams:
+class SuggestedParams(TypeCheck):
     """
     Contains various fields common to all transaction types.
 
@@ -37,8 +35,17 @@ class SuggestedParams:
         min_fee (int)
     """
 
-    def __init__(self, fee, first, last, gh, gen=None, flat_fee=False,
-                 consensus_version=None, min_fee=None):
+    def __init__(
+        self,
+        fee: int,
+        first: int,
+        last: int,
+        gh: str,
+        gen: str = None,
+        flat_fee: bool = False,
+        consensus_version: str = None,
+        min_fee: int = None,
+    ):
         self.first = first
         self.last = last
         self.gh = gh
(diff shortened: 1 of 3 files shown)


Niraj-Kamdar:
Let me know which formatter are you guys using. I have used my default one for now.

Niraj-Kamdar:
Unittests are working locally but I think there seems to be errors related to invalid types in integration tests. It seems like I have to go through CI log and change relevant test suite.

Niraj-Kamdar:
There is bug in testing scripts:
```
FAILURE in step 'I build an application transaction with operation "create", application-id 0, sender "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4", approval-program "programs/loccheck.teal.tok", clear-program "programs/one.teal.tok", global-bytes 1, global-ints 0, local-bytes 1, local-ints 0, app-args "str:test", foreign-apps "5555,6666", foreign-assets "", app-accounts "", fee 1234, first-valid 9000, last-valid 9010, genesis-hash "Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M=", extra-pages 0':
  Feature:  Transaction fee test
  Scenario: App Call Fee Test -- @1.1 
Assertion Failed: expected type of approval_program: <class 'bytes'> but got <class 'bytearray'>
exception: expected type of approval_program: <class 'bytes'> but got <class 'bytearray'>
```

As mentioned in log `approval_program` should be `bytes` type but passed `bytearray` instead. It can be solved by typecasting to `bytes` but I don't know how to do it in `.feature` files. 

An alternate solution can be auto typecasting in `bytes`

michielmulders:
@jasonpaulos Can you help here? Thanks!

Niraj-Kamdar:
Any updates?

jasonpaulos (changes requested):
Hi @Niraj-Kamdar, I apologize for the delay in reviewing. I believe we would be interested in this change if you could make `TypeCheck` into a [function decorator](https://book.pythontips.com/en/latest/decorators.html) instead of a base class, since then we could type check more than just `__init__` methods.

Additionally, could you refrain from running a formatter in this PR? Soon we will add a formatter to this repo and reformat all the code, but that will happen in a separate PR.

Niraj-Kamdar:
> Hi @Niraj-Kamdar, I apologize for the delay in reviewing. I believe we would be interested in this change if you could make `TypeCheck` into a [function decorator](https://book.pythontips.com/en/latest/decorators.html) instead of a base class, since then we could type check more than just `__init__` methods.

Having it as a class has its own benefits. This will work for nested inheritance because of MRO and I think it's better than decorating every classes out there.

I can create a friendly decorator though if you want to type check a particular function.

jannotti:
Is this technique usable when the argument types or not simple `int`, `str`, etc?  Does `isinstance(x, Union[int,str])` work?

jasonpaulos:
Closing due to inactivity

PR #205, opened 2021-06-22
Raise JSONDecodeError instead of None (#193)

### LINKED ISSUE #193: Don't ignore JSON decoding errors
## Summary
Currently if a JSON algod request cannot parse the response, it just returns `None`: https://github.com/algorand/py-algorand-sdk/blob/6b304e4b080d95286f836d8cd6d8cc3aeeefdfd5/algosdk/v2client/algod.py#L81-L84

## Scope
We should really return an error message in this case instead.

## Urgency
It's important to display the proper error if decoding is unsuccessful.

jkschin:
Fixes #193. @jasonpaulos let me know if this works. I ran `make docker-test` and all tests passed (I didn't expect any to fail in any case).

I took the simple approach of just raising the `JSONDecodeError` that's caught.

### DIFF
--- a/algosdk/error.py
+++ b/algosdk/error.py
@@ -152,5 +152,11 @@ class AlgodHTTPError(Exception):
         super().__init__(msg)
         self.code = code
 
+
+class AlgodResponseError(Exception):
+    def __init__(self, msg):
+        super().__init__(msg)
+
+
 class IndexerHTTPError(Exception):
     pass
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -81,7 +81,7 @@ class AlgodClient:
             try:
                 return json.load(resp)
             except json.JSONDecodeError:
-                return None
+                raise json.JSONDecodeError
         else:
             return resp.read()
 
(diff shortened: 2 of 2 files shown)


jkschin:
I did a pseudo test to make sure this works:

```python
import json

f = open("data.json")
try:
    data = json.load(f)
except json.JSONDecodeError:
    raise
```

where, `data.json` is missing a double quote:

```yaml
{
  test": "hello",
}
```

The error received was:
```
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 2 column 3 (char 4)
```
which should be sufficient to help someone debug a potential error in a bad JSON received. 

However, after doing this, it makes me wonder if we should just not do a try-except and `return json.load(resp)` as is. `json.load` can also raise a `TypeError`, which is something we might miss with this current implementation that only catches a `JSONDecodeError`.

jasonpaulos:
@jkschin how about this: we catch any exception that could result from decoding the JSON value and use [exception chaining](https://stackoverflow.com/a/792163) to raise a more descriptive error. Something like:

```python
try:
  json.load(...)
except Exception as e:
  # this would be a new error class
  raise AlgodResponseError("Failed to parse JSON response from algod") from e
```

jkschin:
Yup that makes sense as well! Made the changes you suggested.

jasonpaulos (approved):
Looks great, thanks!

Issue #187, opened 2021-03-24
mnemonic.to_public_key


bhemen:
The function mnemonic.to_public_key() returns the *address* not the public key.  The code below highlights the inconsistency:

```
from algosdk import account, mnemonic

sk, addr = account.generate_account()
mnemonic_secret = mnemonic.from_private_key(sk)
pk = mnemonic.to_public_key(mnemonic_secret)
assert pk == addr
addr = account.address_from_private_key(sk)
assert pk == addr
```

Since the [address and the public key are different in Algorand ](https://developer.algorand.org/docs/features/accounts/), I would suggest changing the name of the function to "mnemonic.to_address()" or something similar.

Thanks

jkschin:
@ian-algorand what does the `FDE` label mean? Is the team open to taking PRs for this? I can see that `to_public_key` is being used only in `account_backup_example.py` and `test_unit.py`, but a change of method name would be quite a breaking change for users / other APIs who use this method, so I'm not sure if it's a good idea to change it for semantic reasons.

jasonpaulos:
@jkschin we are open to taking a PR that renames this function to `mnemonic.to_address`, but keeps `mnemonic.to_public_key` as a deprecated method. Like you pointed out, we unfortunately cannot change the functionality of  the existing `mnemonic.to_public_key` function without breaking existing code, however it would be nice to fix this going forward.

PR #207, opened 2021-06-22
Deprecate to_public_key and remove internal usage of to_public_key

jkschin:
Fixes https://github.com/algorand/py-algorand-sdk/issues/187.

### DIFF
--- a/algosdk/mnemonic.py
+++ b/algosdk/mnemonic.py
@@ -79,10 +79,10 @@ def to_private_key(mnemonic):
 
 
 def to_public_key(mnemonic):
-    warnings.warn("to_public_key returns the Algorand address and will be deprecated, "
-                  "use account.address_from_private_key instead", DeprecationWarning)
     """
     Return the public key for the mnemonic.
+    This method returns the Algorand address and will be deprecated, use
+    account.address_from_private_key instead.
 
     Args:
         mnemonic (str): mnemonic of the private key
@@ -90,6 +90,9 @@ def to_public_key(mnemonic):
     Returns:
         str: public key in base32
     """
+    warnings.warn("to_public_key returns the Algorand address and will be "
+                  "deprecated, use account.address_from_private_key instead",
+                  DeprecationWarning)
     key_bytes = _to_key(mnemonic)
     key = signing.SigningKey(key_bytes)
     return encoding.encode_address(key.verify_key.encode())
--- a/algosdk/mnemonic.py
+++ b/algosdk/mnemonic.py
@@ -4,6 +4,7 @@ from . import wordlist
 from . import error
 from . import constants
 from . import encoding
+import warnings
 
 
 word_to_index = {}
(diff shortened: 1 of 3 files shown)


algosdk/mnemonic.py
```diff
@@ -92,6 +94,21 @@ def to_public_key(mnemonic):
     return encoding.encode_address(key.verify_key.encode())
 
 
+def to_address(mnemonic):
+    """
+    Return the address for the mnemonic of the public key.
+
+    Args:
+        mnemonic (str): mnemonic of the public key
```
jkschin, line 102: @jasonpaulos I followed the description of `to_public_key` when writing this. I realized that this might be a little confusing. The mnemonic is actually the concatenation of the private key and the public key, as seen in `generate_account` [here](https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/account.py#L16). For most of the functions in `mnemonic.py`, I see snippets of "mnemonic of the public key" or "mnemonic of private key". I'm wondering if it should just be renamed and redocumented in general as simply "mnemonic" to reduce confusion.

jkschin:
`make docker-test` ran locally passed.

jasonpaulos:
@jkschin thanks for working on this. I was not aware that the private key we return actually also contains the public key. Upon closer inspection I believe the mnemonic only contains the private key though.

I think in light of this, and to make this SDK consistent with the others, the change should be:
* Deprecate `mnemonic.to_public_key` (as you've already done)
* Don't introduce a new `mnemonic.to_address` method, instead use `account.address_from_private_key` to obtain the address from the private key returned by `mnemonic.to_private_key`
* Change all occurrences of `mnemonic of the public key` to `mnemonic of the private key`, which is more accurate.

jkschin:
Yeah you are right. The code truncates it by taking the first 32 bytes. I guess the definition of `private_key` in the context of this repository is the concatenation of the `private_key` and the `public_key`. Changing that will probably be more involved. As an example, `mnemonic.from_private_key` actually takes in the private key and the public key but the code documentation writes only private key.

Your suggestions make sense, I'll go ahead and make them now.

jkschin:
@jasonpaulos if this looks good, we can merge this. On the administrative side, we might want to write a specific version number where this method will be removed, or write a date where this method will be removed.

algosdk/mnemonic.py
```diff
@@ -78,11 +79,13 @@ def to_private_key(mnemonic):
 
 
 def to_public_key(mnemonic):
+    warnings.warn("to_public_key returns the Algorand address and will be deprecated, "
```
jasonpaulos, line 82: Could you move this below the docstring, and also add a line to the docstring that says a similar thing?

jasonpaulos (changes requested):
Just one comment. I think this is ready to merge after it's fixed.

algosdk/mnemonic.py
jkschin: Ah yeah I missed that! Fixed.

jasonpaulos (approved):
Looks good, thanks for this

PR #209, opened 2021-06-25
Add OnlineyKeyregTxn and OfflineKeyregTxn class and additional tests.

### LINKED ISSUE #192: SDK does not recognize offline keyreg transaction
There is no offline key registration type, so you have to instantiate it with the base Transaction class and set the type manually.
`txn = transaction.Transaction(addr, sp, None, None, constants.keyreg_txn, None)`

This is not ideal but fine, until you go to use `transaction.retrieve_from_file(<offline-txn-file>)` and then it fails because it tries to look for partkey info.

jkschin:
Fixes #192.


algosdk/future/transaction.py
```diff
@@ -477,6 +483,161 @@ def __eq__(self, other):
...
+            self.fee = max(self.estimate_size() * self.fee,
+                           constants.min_txn_fee)
+
+    def dictify(self):
+        d = super(OfflineKeyregTxn, self).dictify()
+        od = OrderedDict(sorted(d.items()))
+        return od
+
+    @staticmethod
+    def _undictify(d):
```
jkschin, line 631: We could remove this method but I wanted to conform to the style of the other code. My preference is to leave an empty method here.

jkschin:
@jasonpaulos this PR is ready for review. I noticed that `KeyregTxn` is also implemented in the non-future `transaction.py` and it seems to account for backward compatibility with the future `transaction.py`. Since `OnlineKeyregTxn` and `OfflineKeyregTxn` will implemented and `KeyregTxn` will be deprecated (I assume), I only implemented the two new classes in the future folder. Here's a summary of the change:

1. Added two new constants `offline_keyreg_txn` and `online_keyreg_txn`.
2. Added two new classes `OfflineKeyregTxn` and `OnlineKeyregTxn`. 
3. Added 4 new unit tests.

I also noticed that `KeyregTxn` is used [here](https://github.com/algorand/py-algorand-sdk/blob/develop/test/steps/steps.py#L612). However, it doesn't seem to be called by the `behave` framework at all. I'm not sure if I understand this correctly so would like to hear your thoughts on it. How should we account fo the KeyregTxn here?

algosdk/constants.py
```diff
@@ -19,6 +19,10 @@
 """str: indicates a payment transaction"""
 keyreg_txn = "keyreg"
 """str: indicates a key registration transaction"""
+offline_keyreg_txn = "offline_keyreg_txn"
+"""str: indicates an offline key registration transaction"""
+online_keyreg_txn = "online_keyreg_txn"
+"""str: indicates an online key registration transaction"""
```
jasonpaulos, line 25: These types should be removed.

algosdk/future/transaction.py
```diff
@@ -217,6 +217,12 @@ def undictify(d):
         elif txn_type == constants.keyreg_txn:
             args.update(KeyregTxn._undictify(d))
             txn = KeyregTxn(**args)
+        elif txn_type == constants.offline_keyreg_txn:
+            args.update(OfflineKeyregTxn._undictify(d))
+            txn = OfflineKeyregTxn(**args)
+        elif txn_type == constants.online_keyreg_txn:
+            args.update(OnlineKeyregTxn._undictify(d))
+            txn = OnlineKeyregTxn(**args)
```
jasonpaulos, line 225: These cases should be removed.

algosdk/future/transaction.py
```diff
@@ -477,6 +483,161 @@ def __eq__(self, other):
                 self.votekd == other.votekd)
 
 
+class OnlineKeyregTxn(Transaction):
```
jasonpaulos, line 486: This class (and `OfflineKeyregTxn`) should extend `KeyregTxn`.

jasonpaulos (changes requested):
Thanks for working on this. I have requested changes because `OfflineKeyregTxn` and `OnlineKeyregTxn` are not new protocol-level transaction types, rather they are new library-level types for convenience. What I mean by this is that there exists only a single key registration transaction type for Algorand, but this type can be used to register an offline or online keys.

Therefore, `OfflineKeyregTxn` and `OnlineKeyregTxn` should inherit from the `KeyregTxn` class. They should be similar to `AssetDestroyTxn` and `AssetCreateTxn`, which are also library-level types, not protocol-level types. They both inherit from a common `AssetConfigTxn` type.

jasonpaulos:
> I also noticed that `KeyregTxn` is used [here](https://github.com/algorand/py-algorand-sdk/blob/develop/test/steps/steps.py#L612). However, it doesn't seem to be called by the `behave` framework at all. I'm not sure if I understand this correctly so would like to hear your thoughts on it. How should we account for the `KeyregTxn` here?

This instance of `KeyregTxn` should probably be replaced by `OnlineKeyregTxn` once the above changes have been made.

jkschin:
@jasonpaulos thanks for the comments. I'll make the fixes and update within the next 24 hours hopefully.

algosdk/constants.py
jkschin: Removed.

algosdk/future/transaction.py
jkschin: Removed.

algosdk/future/transaction.py
jkschin: Fixed.

test_unit.py
```diff
@@ -1317,11 +1415,12 @@ def test_keyreg_txn_online(self):
 
     def test_keyreg_txn_offline(self):
         keyregtxn = (
-            "hqNmZWXNA+iiZnbOALutq6JnaMQgSGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/c"
-            "OUJOiKibHbOALuxk6NzbmTEIAn70nYsCPhsWua/bdenqQHeZnXXUOB+jFx2mGR9tu"
-            "H9pHR5cGWma2V5cmVn")
+            "h6NmZWXNA+iiZnbOALutq6JnaMQgSGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/c"
```
jkschin, line 1418: This change was necessary to add a new `nonpart` field into the message as `KeyregTxn` undictify sorts it into the respective online or offline transaction.

algosdk/future/transaction.py
```diff
@@ -216,7 +216,10 @@ def undictify(d):
             txn = PaymentTxn(**args)
         elif txn_type == constants.keyreg_txn:
             args.update(KeyregTxn._undictify(d))
-            txn = KeyregTxn(**args)
+            if "nonpart" in d and d["nonpart"]:
```
jkschin, line 219: Since the recommendation is to not create a new `txn_type`, it seems like the only way to distinguish the transactions is through the `nonpart` parameter, which we do here. I'm not sure if there's a better solution.

test_unit.py
```diff
@@ -403,6 +405,102 @@ def test_serialize_keyreg_nonpart(self):
             "TEIAn70nYsCPhsWua/bdenqQHeZnXXUOB+jFx2mGR9tuH9pHR5cGWma2V5cmVn")
         self.assertEqual(golden, encoding.msgpack_encode(signed_txn))
 
+    def test_serialize_onlinekeyreg(self):
```
jkschin, line 408: Note that this test is identical to `test_serialize_keyreg_online`. The only difference is that this uses `OnlineKeyregTxn`. The reason for this is that in the wild, `KeyregTxn` might still be directly called so we want this test to still catch potential bugs that might be introduced in future.

test_unit.py
```diff
@@ -403,6 +405,102 @@ def test_serialize_keyreg_nonpart(self):
...
+            fee, 322575, 323575, gh, flat_fee=True)
+        txn = transaction.OnlineKeyregTxn(
+            pk, sp, votepk, selpk, votefirst, votelast, votedilution)
+        path = "/tmp/%s" % uuid.uuid4()
+        transaction.write_to_file([txn], path)
+        txnr = transaction.retrieve_from_file(path)[0]
+        os.remove(path)
+        self.assertEqual(txn, txnr)
+
+    def test_serialize_offlinekeyreg(self):
```
jkschin, line 463: Similarly, this test is identical to `test_serialize_keyreg_offline`.

algosdk/future/transaction.py
```diff
@@ -215,8 +215,12 @@ def undictify(d):
             args.update(PaymentTxn._undictify(d))
             txn = PaymentTxn(**args)
         elif txn_type == constants.keyreg_txn:
-            args.update(KeyregTxn._undictify(d))
```
jkschin, line 218: We remove `KeyregTxn._undictify` method as this is the only place where this method is called, and undictify is dependent on the type of KeyregTxn. For example, in the offline case, there are no additional arguments to add.

algosdk/future/transaction.py
jasonpaulos: This looks good to me

algosdk/future/transaction.py
```diff
@@ -434,47 +438,148 @@ def dictify(self):
...
+        if not isinstance(other, (
+                KeyregTxn,
+                transaction.KeyregTxn)):
+            return False
+        return (super(KeyregTxn, self).__eq__(other) and
+                self.votepk == other.votepk and
+                self.selkey == other.selkey and
+                self.votefst == other.votefst and
+                self.votelst == other.votelst and
+                self.votekd == other.votekd)
```
jasonpaulos, line 451: I think this should also compare `nonpart`.

algosdk/future/transaction.py
```diff
@@ -434,47 +438,148 @@ def dictify(self):
...
+        self.votepk = votekey
+        self.selkey = selkey
+        self.votefst = votefst
+        self.votelst = votelst
+        self.votekd = votekd
+        if not sp.flat_fee:
+            self.fee = max(self.estimate_size() * self.fee,
+                           constants.min_txn_fee)
+
+    def dictify(self):
```
jasonpaulos, line 506: I believe you can remove this method, since it doesn't add anything new from the inherited method.

algosdk/future/transaction.py
```diff
@@ -434,47 +438,148 @@ def dictify(self):
...
+
+    def __init__(self, sender, sp, note=None, lease=None, rekey_to=None):
+        KeyregTxn.__init__(self, sender, sp, None, None, None, None, None,
+                           note=note, lease=lease, rekey_to=rekey_to,
+                           nonpart=True)
+        if not sp.flat_fee:
+            self.fee = max(self.estimate_size() * self.fee,
+                           constants.min_txn_fee)
+
+    def dictify(self):
```
jasonpaulos, line 569: Likewise, this can also be removed.

jasonpaulos (changes requested):
Hi @jkschin, after some internal discussion we realized there should actually be 3 `KeyregTxn` types: online, offline, and nonparticipating.
* Online should have non-None values for `votekey`, `selkey`, `votefst`, `votelst`, and `votekd`, and `nonpart` should be False.
* Offline should have None values for `votekey`, `selkey`, `votefst`, `votelst`, and `votekd`, and `nonpart` should be False.
* Nonparticipating should have None values for `votekey`, `selkey`, `votefst`, `votelst`, and `votekd`, and `nonpart` should be True.

Which means the current `OfflineKeyregTxn` is actually the nonparticipating case.

Additionally, we would prefer if the new classes had `Keyreg` at the beginning, e.g. `KeyregOnlineTxn`.

I also left some comments for minor changes. Once these and the above changes are made, I'd be happy to approve this. Thank you again for working on this PR!

jkschin:
@jasonpaulos thanks for the review and sorry for my late response. Will work on this over the next 48 hours and ping you when this is ready!

algosdk/future/transaction.py
jkschin: fixed.

algosdk/future/transaction.py
jkschin: Yup that's correct. Thanks!

algosdk/future/transaction.py
jkschin: removed.

algosdk/error.py
```diff
@@ -135,6 +135,11 @@ def __init__(self):
             "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ")
 
 
+class KeyregOnlineTxnInitError(Exception):
```
jkschin, line 138: @jasonpaulos I added this error to check that all the attributes are not `None`.

algosdk/future/transaction.py
```diff
@@ -434,47 +446,196 @@ def dictify(self):
...
+                 votelst, votekd, note=None,
+                 lease=None, rekey_to=None):
+        KeyregTxn.__init__(self, sender, sp, votekey, selkey, votefst, votelst,
+                           votekd, note, lease, rekey_to, nonpart=False)
+        self.votepk = votekey
+        self.selkey = selkey
+        self.votefst = votefst
+        self.votelst = votelst
+        self.votekd = votekd
+        if votekey is None:
```
jkschin, line 511: @jasonpaulos I added checks here to ensure that a `KeyregOnlineTxn` cannot be initialized when some of the required values are `None`.

algosdk/future/transaction.py
```diff
@@ -434,47 +446,196 @@ def dictify(self):
...
+        votefst = d["votefst"]
+        votelst = d["votelst"]
+        votekd = d["votekd"]
         args = {
             "votekey": votekey,
             "selkey": selkey,
             "votefst": votefst,
             "votelst": votelst,
             "votekd": votekd,
-            "nonpart": nonpart
```
jkschin, line 463: If you're curious why `nonpart` is removed here, it's because it already exists in the dictionary.

test_unit.py
```diff
@@ -403,6 +405,166 @@ def test_serialize_keyreg_nonpart(self):
...
+            fee, 322575, 323575, gh, flat_fee=True)
+        txn = transaction.KeyregOnlineTxn(
+            pk, sp, votepk, selpk, votefirst, votelast, votedilution)
+        path = "/tmp/%s" % uuid.uuid4()
+        transaction.write_to_file([txn], path)
+        txnr = transaction.retrieve_from_file(path)[0]
+        os.remove(path)
+        self.assertEqual(txn, txnr)
+
+    def test_init_keyregonlinetxn_with_none_values(self):
```
jkschin, line 463: @jasonpaulos I added this test to check that an error is thrown when a `None` value is in one of the 5 required variables.

jkschin:
@jasonpaulos sorry for the delay - this looks good for a review now!

test_unit.py
jasonpaulos: I think the change to this test should be reverted, since now we have online, offline, and nonparticipating keyreg txns, so the original implementation is correct again.

jasonpaulos (commented):
@jkschin thanks for making the requested changes! We really appreciate the contribution. This looks really good, just one test needs to be changed before merging.

test_unit.py
jkschin: This has been reverted and the test passes, but I had to make a change to another part of the code which I will tag you on and explain it.

algosdk/future/transaction.py
```diff
@@ -219,11 +219,18 @@ def undictify(d):
                 args.update(KeyregNonparticipatingTxn._undictify(d))
                 txn = KeyregNonparticipatingTxn(**args)
             else:
-                if (d["votekey"] is None and
-                    d["selkey"] is None and
-                    d["votefst"] is None and
-                    d["votelst"] is None and
-                    d["votekd"] is None):
+                if ("votekey" not in d and
```
jkschin, line 222: @jasonpaulos I had to add this section here in an `if`-`elif` block. There are two reasons:
1. I wrote a test `test_write_read_keyregofflinetxn`. This test initializes a `KeyregOfflineTxn` and then calls `transaction.write_to_file`. I then try `transaction.retrieve_from_file` and verify that the two objects are identical. In this instance, when `write_to_file` is called, the original `KeyregTxn` class method `dictify` is called, and it asserts that the dictionary is populated with the keys `selkey`, `votefst` ... `nonpart`, even if they are `None`. As such, empty values are written to the file as well (which arguably might be an inefficiency?)
2. On the other hand, for `test_keyreg_txn_offline`, which is the original test that we reverted to, it strips empty values and thus a decoded string without `votekey`, `selkey`, etc. is assumed to be a `KeyregOfflineTxn`.

Given this, I decided to first check that the parameters are __not__ in the dictionary, which would imply that it might come from a string generated from `msgpack_encode`. If that first check passes, we then check if these values are `None`. If that fails, then it has to be a `KeyregOnlineTxn`.

jkschin: A more elegant solution would be to refactor the `dictify` method in `KeyregTxn`, but I'm not sure if the team wants to keep that untouched. The change would involve only dictifying `selkey`, `votefst` etc. iff they won't be `None`, which I think we have some guarantees on. That way, the classes would implicitly match the behavior of `msgpack_encode`, which strips `None` values.

jasonpaulos: Thank you for noticing this! In this case I believe the better solution would be to modify the `KeyregTxn.dictify` to only add the fields that are not None. This is more in line with the behavior of the other transaction types.

Then it should only be necessary to check if `votekey` and other parameters are not present in the dictionary in order to call `KeyregOfflineTxn._undictify`.

jkschin: I've made the recommended fixes! This works now, and is probably the best solution since `KeyregTxn` is a class that might be used in the wild. A nicer solution would be to write `dictify` in `KeyregOnlineTxn` and `KeyregOfflineTxn` and `KeyregNonparticipatingTxn` and convert `KeyregTxn` to be an internal class that cannot be instantiated.

jasonpaulos (approved):
Thank you for the great work @jkschin!

PR #210, opened 2021-06-27
Fix typo in min/max balance indexer & make clearer

fabrice102:
Fix a typo and make it clearer that results with an amount equal to the value are excluded.
This is because usually "min" and "max" include the value


jasonpaulos (approved):
Looks good

Issue #169, opened 2021-01-15
Requests need to merge headers


mxmauro:
Reason: CF countermeasures returns 403 for any request with user agent containing the text `Python-urllib`.

Just try: `curl -v -A "Python-urllib" https://www.algorand.com/`

Regards.

fabrice102:
For those affected by the issue, as a temporary workaround, you can manually change the User Agent.
E.g.,
```python
import algosdk.algod as algod
c = algod.AlgodClient('', 'https://api.algoexplorer.io', headers={'User-Agent': 'DoYouLoveMe?'})
print(c.status())
```

But hopefully CloudFlare will soon fix this issue.

danmurphy1217:
> For those affected by the issue, as a temporary workaround, you can manually change the User Agent.
> E.g.,
> 
> ```python
> import algosdk.algod as algod
> c = algod.AlgodClient('', 'https://api.algoexplorer.io', headers={'User-Agent': 'DoYouLoveMe?'})
> print(c.status())
> ```
> 
> But hopefully CloudFlare will soon fix this issue.

Just an FYI that this does not completely solve the problem. It only serves as a solution if the Algorand SDK is not adding their own headers to HTTP requests later in the program. For example, I can define a user-agent early on as you suggest, but the `send_raw_transaction` function in **algosdk.v2client.algod** ALSO defines headers. Therefore, when I try to send a transaction I get a:

`TypeError: algosdk.v2client.algod.AlgodClient.algod_request() got multiple values for keyword argument 'headers'`

error since the initial headers that I set are passed as kwargs (and two arguments of the same name are not allowed).

I think this can and should be solved on the Algorand side, and I'm happy to write a patch.

mxmauro:
@danmurphy1217  [this code](https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/algod.py#L55) should merge the common headers added by the constructor with the ones on the `send_raw_transaction` call.

danmurphy1217:
> @danmurphy1217 [this code](https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/algod.py#L55) should merge the common headers added in the constructor with the ones on the `send_raw_transaction` call.

Let me provide you with some more context. I am calling the `client.send_transaction` method as follows:

```python
client.send_transaction(signed_transaction, headers={'User-Agent': '[value]'}
```

[...]

PR #211, opened 2021-07-09
Solve Issue #169: Merge Request headers in `algod.py`

danmurphy1217:
Previously, under certain circumstances, if a developer passed their own `headers` to the `send_raw_transaction` method, the following error would occur:

```
TypeError: algosdk.v2client.algod.AlgodClient.algod_request() got multiple values for keyword argument 'headers'
```

This is because the headers defined in `send_raw_transaction` are passed along to `self.algod_request` with the headers that the Algorand SDK automatically passes. This creates duplicate header kwargs, resulting in the above `TypeError`. This PR attempts to solve this problem.

More context is provided in #169. CC: @jasonpaulos

### DIFF
--- a/algosdk/util.py
+++ b/algosdk/util.py
@@ -85,6 +85,6 @@ def build_headers_from(kwarg_headers: Dict[str, Any], additional_headers: Dict[s
     if kwarg_headers:
         kwarg_headers.update(additional_headers)
     else:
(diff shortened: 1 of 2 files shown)


jasonpaulos (approved):
Looks good, thanks for working on this!

Closes #169

danmurphy1217:
@jasonpaulos happy to help! Are there any other outstanding issues that you could currently use some help on?

PR #215, opened 2021-07-19
modified: constants.py to match python.org PEP 8 style guidelines

Fluffycup:
I was talking with a developer about writing code in python, and I too was familiar seeing constants across other SDKs as all uppercase. Thought it would be best to keep in line with general standards. Link to PEP 8 below:
https://www.python.org/dev/peps/pep-0008/#constants

### DIFF
--- a/algosdk/constants.py
+++ b/algosdk/constants.py
@@ -1,89 +1,125 @@
 """
 Contains useful constants.
 """
-
-kmd_auth_header = "X-KMD-API-Token"
+KMD_AUTH_HEADER = "X-KMD-API-Token"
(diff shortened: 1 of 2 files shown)


Fluffycup:
Yep. Don't know why I didn't think that would work with just updating the constants.py file. Working to search and replace everything for a successful build.

Fluffycup:
Alright updated to have constants follow PEP 8 guidelines and also be backwards compatible with original constants. If accepted, assumption going forward is that new constants will follow PEP 8 standard.

jasonpaulos (approved):
This looks great, thanks for doing this.

PR #221, opened 2021-08-05
Python < 3.7 bug fixed

sepandhaghighi:
The current version of `py-algorand-sdk` doesn't work on Python < 3.7.
f-Strings was introduced in Python 3.7, so I changed it to a more general method (`str.format`) to be backward compatible.

## Current version
SS

## This PR
SS2


jasonpaulos (approved):
Thanks for noticing and fixing this

PR #224, opened 2021-08-11
Add black as a code formatter to py-algorand-sdk

algochoi:
This adds a Travis check to check that commits are formatted using `black` and formats existing files. 

When reviewing, it will be helpful to review against [this commit](https://github.com/algorand/py-algorand-sdk/pull/224/commits/6627bd6816af7d41e8d87cab6cdd373a1a214e2b) and [this commit](https://github.com/algorand/py-algorand-sdk/pull/224/commits/e1abbf6c2158e4555d22fed424d4ef4a6afc9833), because there are lots of formatting changes that aren't very relevant to the human eye.


setup.py
```diff
@@ -16,9 +16,9 @@
     project_urls={
         "Source": "https://github.com/algorand/py-algorand-sdk",
     },
-    install_requires=["pynacl", "pycryptodomex>=3.6.0", "msgpack"],
+    install_requires=["pynacl", "pycryptodomex>=3.6.0", "msgpack", "black"],
```
jasonpaulos, line 19: Why did you revert your changes to the `requirements.txt` file and place black here instead? I believe this is the list of runtime dependencies, and if black is only needed during development (like most linters) `requirements.txt` is a better place for it.

algochoi: Ah okay, I will put it back to `requirements.txt`.

bricerisingalgorand:
When I ran the tests locally I saw these two errors. I just restarted the job so hopefully we'll either see it pass or the same errors from travis 🤞 

```
Failing scenarios:
  test/features/integration/applications.feature:83  local test - Use every applications feature! -- @1.1 
  test/features/integration/applications.feature:84  global test - Use every applications feature! -- @1.2 
```

examples/asset_create_example.py
```diff
@@ -3,30 +3,60 @@
 from algosdk import account
 from algosdk.future import transaction
 
-private_key, address = account.generate_account() # creator
-_, freeze = account.generate_account() # account that can freeze other accounts for this asset
-_, manager = account.generate_account() # account able to update asset configuration
-_, clawback = account.generate_account() # account allowed to take this asset from any other account
-_, reserve = account.generate_account() # account that holds reserves for this asset
+private_key, address = account.generate_account()  # creator
```
jasonpaulos, line 6: This seems unnecessarily complicated. I think if you moved the comments from the same line to the line before, the formatted code might be simpler?

examples/asset_update_example.py
```diff
@@ -1,26 +1,55 @@
...
 
 from algosdk import account
 from algosdk.future import transaction
 
-manager_private_key, manager_address = account.generate_account() # this transaction must be sent from the manager's account
-_, new_freeze = account.generate_account() # account that can freeze other accounts for this asset
-_, new_manager = account.generate_account() # account able to update asset configuration
-_, new_clawback = account.generate_account() # account allowed to take this asset from any other account
-_, new_reserve = account.generate_account() # account that holds reserves for this asset
+(
```
jasonpaulos, line 6: Same thing here

jasonpaulos (commented):
Thank you for this! I just had a few small comments

examples/asset_create_example.py
algochoi: Thanks for catching this! Moving the comment does the trick, i.e. 
```
# account that can freeze other accounts for this asset
_, freeze = account.generate_account()
```

algosdk/error.py
```diff
@@ -1,7 +1,8 @@
 class BadTxnSenderError(Exception):
     def __init__(self):
-        Exception.__init__(self, "transaction sender does not match multisig "
-                                 "parameters")
+        Exception.__init__(
+            self, "transaction sender does not match multisig " "parameters"
```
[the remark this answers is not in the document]
algochoi: I'll try to use some regex to catch these

jasonpaulos (approved):
Looks good!

PR #236, opened 2021-09-27
Support AVM 1.0

### LINKED ISSUE #235: Support AVM 1.0
## Summary
Update this SDK to work with features from AVM 1.0.

## Scope
Specifically:
* Update TEAL langspec & add corresponding tests
* Make sure algod REST endpoints can support returning logs and inner transactions (might not need to do anything for this SDK, since it directly returns a dictionary of the response).
* Add a way to get app address from app ID (and test it).

## Urgency
AVM 1.0 just got enabled in BetaNet, so good to do this as soon as possible.

jasonpaulos:
Update TEAL langspec to v5 and add function `get_application_address` to compute the address of an application.

Since this SDK does not do any processing on the return values from the algod REST endpoints, no changes are needed in order for it to return logs and inner transactions.

Closes #235.


test/steps/v2_steps.py
```diff
@@ -858,6 +858,23 @@ def lookup_application(context, app_id):
...
+)
+def lookup_application_logs(
+    context, app_id, limit, min_round, max_round, next_token, sender, txid
+):
+    context.response = context.icl.application_logs(
+        int(app_id),
+        limit=int(limit),
+        min_round=int(min_round),
+        max_round=int(max_round),
+        next_page=next_token,
```
shiqizng, line 872: should we use next_token instead of next_page? next_page seems to imply an int value and next_token would be more consistent with the naming in go sdk.

shiqizng (approved):
Looks good.

test/steps/v2_steps.py
jasonpaulos: Other methods in this SDK also use `next_page`, so I did the same for consistency (but maybe we should change this in v2).

PR #238, opened 2021-09-28
ABI Type encoding support

### LINKED ISSUE #228: Add ABI encoding support
## Summary
Add support for encoding and decoding ABI values, as described in https://github.com/algorandfoundation/ARCs/pull/7

## Scope
Add ability to:
* Parse and serialize ABI type strings, e.g. `(uint64, byte[], address)`
* Encode and decode ABI values, using type information

algochoi:
This PR adds ABI encoding support for the Python SDK:

- Adds `abi` support for types, e.g. `from algosdk.abi import UintType`
- Adds ABI encoding and decoding functions
- Adds unit tests largely from the original [goal PR that implemented ABI encoding support](https://github.com/algorand/go-algorand/pull/2807) 

Closes #228


algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        )
+
+    @staticmethod
+    def make_string_type():
+        return Type(BaseType.String)
+
+    @staticmethod
+    def make_tuple_type(arg_types):
+        if len(arg_types) >= 2 ** 16:
+            raise error.ABITypeError(f"tuple args exceed 2^16")
```
ahangsu, line 165: more specifically, should say something like tuple element number larger than uint16 max.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+                else:
+                    raise error.ABITypeError(
+                        f"cannot convert {tup} to an ABI type"
+                    )
+
+            return Type.make_tuple_type(tuple_list)
+        else:
+            raise error.ABITypeError(f"cannot convert {s} to an ABI type")
+
+    def parse_tuple(s):
```
ahangsu, line 228: I think this should have a static method decorator?

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        elif self.abi_type_id == BaseType.Address:
+            return f"address"
+        elif self.abi_type_id == BaseType.ArrayDynamic:
+            return f"{str(self.child_type)}[]"
+        elif self.abi_type_id == BaseType.String:
+            return f"string"
+        elif self.abi_type_id == BaseType.Tuple:
+            if not self.child_types:
+                return f"()"
+            return f"({','.join(str(t) for t in self.child_types)})"
```
ahangsu, line 76: It seems, if no `child_types` passed in construction, then it is an empty list. I guess it still works without the if statement checking if child types is none?

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        bit_size (int)
+        precision (int)
+        static_length (int)
+    """
+
+    def __init__(
+        self,
+        type_id,
+        child_type=None,
+        child_types=list(),
```
ahangsu, line 44: Not sure if it is the best practice of passing list() to default argument, but if it were me, I would do something like set it to be None as default, and check in constructor that: if `child_type` is None, then set it to be a list()

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            or self.abi_type_id == BaseType.Ufixed
+        ):
+            return self.bit_size // 8
+        elif self.abi_type_id == BaseType.ArrayStatic:
+            if self.child_type.abi_type_id == BaseType.Bool:
+                # 8 Boolean values can be encoded into 1 byte
+                byte_length = self.static_length // 8
+                if self.static_length % 8 != 0:
+                    byte_length += 1
+                return byte_length
```
ahangsu, line 319: There should be an easy way to do: use math.ceil, and calculate the byte_length = ceil(self.static_length / 8).
It will evaluate self.static_length / 8 to be float and ceil the result back to int.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            self.abi_type_id == BaseType.ArrayDynamic
+            or self.abi_type_id == BaseType.String
+        ):
+            return True
+        for child in self.child_types:
+            if child.is_dynamic():
+                return True
+        return False
+
+    def find_bool(type_list, index, delta):
```
ahangsu, line 282: I think this is also a static method.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+                    )
+
+            return Type.make_tuple_type(tuple_list)
+        else:
+            raise error.ABITypeError(f"cannot convert {s} to an ABI type")
+
+    def parse_tuple(s):
+        """
+        Given a tuple string, parses one layer of the tuple.
+        i.e. '(x,(y,(z)))' -> ['x', '(y,(z))']
```
ahangsu, line 231: Actually, I think the string argument passed in is pre-processed, i.e., it has form 'x,(y,z)' rather than form '(x,(y,z))'. See line 210. As a result, minor change needed in comment here :)

test_unit.py
```diff
@@ -3570,6 +3571,262 @@ def test_local_state(self):
...
+            with self.assertRaises(error.ABITypeError) as e:
+                abi.Type.make_ufixed_type(8, j)
+            self.assertIn(
+                f"unsupported ufixed precision: {j}", str(e.exception)
+            )
+
+    def test_type_from_string_invalid(self):
+        test_cases = (
+            # uint
+            # Whitespaces are allowed, e.g. "uint8 " or "uint 8" can be parsed
```
ahangsu, line 3713: I am not so sure if it is allowed, please refer to ARC#4.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            matches = re.search(static_array_regex, s)
+            try:
+                static_length = int(matches.group(2))
+                array_type = Type.type_from_string(matches.group(1))
+                return Type.make_static_array_type(array_type, static_length)
+            except:
+                raise error.ABITypeError(f"malformed static array string: {s}")
+        elif s.startswith("uint"):
+            try:
+                type_size = int(s[4:])
```
ahangsu, line 187: I think some kind of check need to be performed over s[4:], say eliminating the space.
In golang, the integer parsing from string does not allow spaces, it will return err. In this case, this can allow spaces in head/tail.

test_unit.py
algochoi: Ah yes, you are correct -- I left this as a note to myself but forgot about it. I think `isdecimal()` can be used to check whether the string literal is a valid integer size.

algosdk/abi.py
aldur: I suggest the `None` as well since passing a list as a default argument could have astonishing effects :) 

https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

algochoi: Thanks! Will fix this

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
+from enum import IntEnum
+import re
+from unittest.mock import Base
```
jasonpaulos, line 3: This import seems unnecessary

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        self.abi_type_id = type_id
+        self.child_type = child_type  # Used for arrays
+        self.child_types = child_types  # Used for tuples
+        self.bit_size = bit_size
+        self.precision = precision
+        self.static_length = static_length
+
+    def __str__(self):
+        if self.abi_type_id == BaseType.Uint:
+            return f"uint{str(self.bit_size)}"
```
jasonpaulos, line 58: Unfortunately this SDK currently targets Python 3.5+, so f-strings may not be available. Perhaps we should drop support for 3.5 at this point?

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        if (
+            not isinstance(other, Type)
+            or self.abi_type_id != other.abi_type_id
+            or self.precision != other.precision
+            or self.bit_size != other.bit_size
+            or self.static_length != other.static_length
+            or len(self.child_types) != len(other.child_types)
+        ):
+            return False
+        for i in range(len(self.child_types)):
```
jasonpaulos, line 90: I believe you can directly check `self.child_types == other.child_types`, since the list `==` operator only returns true if all values are equal

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        child_types (list)
+        bit_size (int)
+        precision (int)
+        static_length (int)
+    """
+
+    def __init__(
+        self,
+        type_id,
+        child_type=None,
```
jasonpaulos, line 43: It might be a good idea to enforce that all arguments after `type_id` must be provided with keywords. You [can do this with `*`](https://stackoverflow.com/a/14298976), i.e. `__init__(self, type_id, *, child_type, ...`).

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            return Type.make_dynamic_array_type(array_arg_type)
+        elif s.endswith("]"):
+            static_array_regex = "^([a-z\d\[\](),]+)\[([1-9][\d]*)]$"
+            matches = re.search(static_array_regex, s)
+            try:
+                static_length = int(matches.group(2))
+                array_type = Type.type_from_string(matches.group(1))
+                return Type.make_static_array_type(array_type, static_length)
+            except:
+                raise error.ABITypeError(f"malformed static array string: {s}")
```
jasonpaulos, line 184:
```suggestion
            except Exception as e:
                raise error.ABITypeError(f"malformed static array string: {s}") from e
```
Since you are raising an error only as a result of another error, it's a good idea to use `raise ... from ...` to preserve this relationship. More info here: https://stefan.sofa-rockers.org/2020/10/28/raise-from/

There are other places below that would benefit from this as well.

algosdk/abi.py
jasonpaulos: This is a private helper function, right? Then it should probably be a normal function outside of the `Type` class, so it will remain private.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            return Type.make_address_type()
+        elif s == "string":
+            return Type.make_string_type()
+        elif len(s) >= 2 and s[0] == "(" and s[-1] == ")":
+            # Recursively parse parentheses from a tuple string
+            tuples = Type.parse_tuple(s[1:-1])
+            tuple_list = []
+            for tup in tuples:
+                if isinstance(tup, str):
+                    tt = Type.type_from_string(str(tup))
```
jasonpaulos, line 214: nit: no need to convert to `str` since you just checked it it was a `str`

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+                    tuple_list.append(tt)
+                elif isinstance(tup, list):
+                    tts = list(map(lambda t_: Type.type_from_string(t_), tup))
```
jasonpaulos, line 217: nit: is this the same as `tts = [Type.type_from_string(t_) for t_ in tup]`? If so, the list comprehension seems simpler to understand

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+
+    def is_dynamic(self):
+        """
+        Returns whether the current type or any of its child types is a
+        dynamic type or not.
+        """
+        if (
+            self.abi_type_id == BaseType.ArrayDynamic
+            or self.abi_type_id == BaseType.String
+        ):
```
jasonpaulos, line 275: super nit: `self.abi_type_id in (BaseType.ArrayDynamic, BaseType.String)` is more concise.

Similarly, for below: `return any(child.is_dynamic() for child in self.child_types)`

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+    """
+
+    def __init__(self, abi_type, val) -> None:
+        self.abi_type = abi_type
+        self.value = val
+
+    @staticmethod
+    def make_uint(value, size):
+        assert isinstance(value, int)
+        if value >= (2 ** size):
```
jasonpaulos, line 364: Also worthwhile to check if value is negative

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        return Value(Type.make_string_type, value)
+
+    @staticmethod
+    def make_byte(value):
+        assert isinstance(value, bytes)
+        return Value(Type.make_byte_type, value)
+
+    @staticmethod
+    def make_address(value):
+        assert len(value) == 32
```
jasonpaulos, line 411: Is this meant to accept a base32 encoded address + checksum as input, or the decoded raw 32-byte public key?

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        return Value(Type.make_bool_type, value)
+
+    @staticmethod
+    def make_string(value):
+        assert isinstance(value, str)
+        return Value(Type.make_string_type, value)
+
+    @staticmethod
+    def make_byte(value):
+        assert isinstance(value, bytes)
```
jasonpaulos, line 406: `bytes` is actually the string byte array type. So you should probably check if the input is an int between 0 and 255 inclusive.

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+        return Value(Type.make_byte_type, value)
+
+    @staticmethod
+    def make_address(value):
+        assert len(value) == 32
+        return Value(Type.make_address_type, value)
+
+    @staticmethod
+    def make_dynamic_array(values, element_type):
+        assert isinstance(values, list())
```
jasonpaulos, line 416:
```suggestion
        assert isinstance(values, list)
```

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+                    f"dynamic array type mismatch: {val.abi_type} != {values[0].abi_type}"
+                )
+        return Value(
+            Type.make_static_array_type(values[0].abi_type, len(values)),
+            values,
+        )
+
+    @staticmethod
+    def make_tuple(values):
+        assert isinstance(values, list())
```
jasonpaulos, line 450: Same thing here

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+            )
+        tuple_types = []
+        for val in values:
+            assert isinstance(val, Value)
+            tuple_types.append(val.abi_type)
+
+        return Value(Type.make_tuple_type(tuple_types), values)
+
+    def get_uint(self):
+        if not isinstance(self.abi_type.abi_type_id, BaseType.Uint):
```
jasonpaulos, line 463:
```suggestion
        if self.abi_type.abi_type_id != BaseType.Uint:
```
I think you mean this?

algosdk/abi.py
```diff
@@ -0,0 +1,570 @@
...
+
+    def get_uint64(self):
+        if (
+            self.abi_type.abi_type_id != BaseType.Uint
+            or self.abi_type.bit_size != 64
+        ):
+            raise error.ABIValueError(
+                f"value type {self.abi_type.abi_type_id} is not a uint64"
+            )
+        return self.get_uint()
```
jasonpaulos, line 511: I don't think these functions are useful in python, since there is only one int type and it's essentially a bigint, so it can hold all values.

jasonpaulos (changes requested):

algosdk/abi.py
algochoi: This is a good observation -- in the original Go implementation, it was just an alias for a 32-length byte array. I'm not sure if it should be kept that way or do something a bit more user-friendly.

algosdk/abi/type.py
```diff
@@ -0,0 +1,1026 @@
...
+        bit_size=None,
+        precision=None,
+        static_length=None,
+    ) -> None:
+        self.abi_type_id = type_id
+        self.child_type = child_type  # Used for arrays
+        if not child_types:
+            self.child_types = list()
+        else:
+            self.child_types = child_types  # Used for tuples
```
ahangsu, line 58: this can be simplified:
```suggestion
        self.child_type = list() if not child_types else child_types
```

algosdk/abi/type.py
```diff
@@ -0,0 +1,1026 @@
...
+
+    def encode(self, value):
+        assert isinstance(value, int)
+        if value >= (2 ** self.bit_size) or value < 0:
+            raise error.ABIEncodingError(
+                "value {} is negative or is too big to fit in size {}".format(
+                    value, self.bit_size
+                )
+            )
+        return (value).to_bytes(self.bit_size // 8, byteorder="big")
```
ahangsu, line 151: nit: redundant space around `value`

algosdk/abi/type.py
```diff
@@ -0,0 +1,1026 @@
...
+            return UintType(type_size)
+        except Exception as e:
+            raise error.ABITypeError(
+                "malformed uint string: {}".format(s)
+            ) from e
+    elif s == "byte":
+        return ByteType()
+    elif s.startswith("ufixed"):
+        ufixed_regex = "^ufixed([1-9][\d]*)x([1-9][\d]*)$"
+        matches = re.search(ufixed_regex, s)
```
ahangsu, line 993: nit: can move `ufixed_regex` and `static_array_regex` to global, and match symbols with compiled regex exprs. 
There's some minor benchmark for doing so in python: https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

algosdk/abi/type.py
```diff
@@ -0,0 +1,1026 @@
...
+    def is_dynamic(self):
+        return False
+
+    def encode(self, value):
+        """
+        Encode a single byte or a uint8
+        """
+        # Enforce ByteType to only accept bytes or int values
+        if isinstance(value, bytes):
+            value = int.from_bytes(value, byteorder="big")
```
jasonpaulos, line 198: I don't think there's a need to support `bytes` as an encodable value for this type, since the type represents just a single byte.

But I do think there's a lot of value in supporting `bytes` as an encodable type for dynamic/static ABI arrays of bytes. I believe that's already possible with your existing code, since it appears `TypleType.encode` can accept `bytes` as an input. Though I think you should add a check there to ensure an input type of `bytes` or `bytearray` is only accepted if all of the tuple child types are `ByteType`.

test_unit.py
```diff
@@ -3600,6 +3613,537 @@ def test_local_state(self):
...
+            StringType().decode((0).to_bytes(1, byteorder="big"))
+        with self.assertRaises(error.ABIEncodingError) as e:
+            StringType().decode((1).to_bytes(2, byteorder="big"))
+
+    def test_array_static_encoding(self):
+        test_cases = [
+            (
+                ArrayStaticType(BoolType(), 3),
+                [True, True, False],
+                bytes.fromhex("C0"),
```
jasonpaulos, line 4018: nit: a format like `bytes([0b11000000])` would be easier to understand for these boolean test cases

algosdk/abi/type.py
jasonpaulos: > I thought it might be useful to encode characters like `b"a"`

I personally think this is a little misleading, since `b"a"` is really a 1 element byte array. But python is a bit weird about string characters in general (`"a"[0] == "a"`!), so if you think this is useful, it's fine with me. But I would suggest you add an explicit check that the length is 1, so we fail early if `b"ab"` is passed.

> Also the current String encoding relies on the `ByteType.encode()` to encode [characters byte by byte](https://github.com/algorand/py-algorand-sdk/blob/e4fc085c28357dafda29d05fb7d5b308609bfd83/algosdk/abi/string_type.py#L37).

True but it just as easily could do `value_array.append(val)`

> For the second point, do you mean like: if you have `(bytes,bytes)`, the input can either be like `bytes([1, 2])` or `[b"\x1", b"\x2"]`?

Yes, it would be very convent to be able to directly pass in `b"\x01\x02\x03"` instead of `[1, 2, 3]` (or `[b"\x01", b"\x02", b"\x03"]` if we allow treating length 1 `bytes` as a single byte)

algosdk/abi/string_type.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def is_dynamic(self):
+        return True
+
+    def encode(self, string_val):
+        """
+        Encode a value into a String ABI bytestring.
+
+        Args:
+            value (str | bytes): value to be encoded. It can be either a base32
```
jasonpaulos, line 34: This docstring is wrong

algosdk/abi/tuple_type.py
```diff
@@ -0,0 +1,329 @@
...
+                    # Pack bytes to heads and tails
+                    if before % 8 != 0:
+                        raise error.ABIEncodingError(
+                            "expected before index should have number of bool mod 8 equal 0"
+                        )
+                    after = min(7, after)
+                    compressed_int = TupleType._compress_multiple_bool(
+                        values[i : i + after + 1]
+                    )
+                    # For converting one byte, the byteorder should not matter
```
jasonpaulos, line 177: nit: comment no longer necessary

algosdk/abi/tuple_type.py
```diff
@@ -0,0 +1,329 @@
...
+                    heads.append(encoded_tuple_element)
+                is_dynamic_index.append(False)
+                tails.append(b"")
+            i += 1
+
+        # Adjust heads for dynamic types
+        head_length = 0
+        for head_element in heads:
+            # If the element is not a placeholder, append the length of the element
+            if head_element:
```
jasonpaulos, line 191: This branch still does not differentiate between a 0 length head string that should be allowed (e.g. an empty tuple) and a head string that belongs to a dynamic type.

I would suggest either checking the `is_dynamic_index` list, or using `None` to represent dynamic heads and specifically checking `head_element is None` here.

algosdk/abi/tuple_type.py
```diff
@@ -0,0 +1,329 @@
...
+                    raise error.ABIEncodingError(
+                        "byte length {} should not exceed a uint16".format(
+                            head_value
+                        )
+                    )
+                heads[i] = head_value.to_bytes(
+                    ABI_LENGTH_SIZE, byteorder="big"
+                )
+            if tails[i]:
+                tail_curr_length += len(tails[i])
```
jasonpaulos, line 212: nit: no need to have this code behind an if statement

algosdk/abi/tuple_type.py
```diff
@@ -0,0 +1,329 @@
...
+                        "malformed value: dynamically typed values must contain a two-byte length specifier"
+                    )
+                # Decode the size of the dynamic element
+                dynamic_index = int.from_bytes(
+                    tuple_string[array_index : array_index + ABI_LENGTH_SIZE],
+                    byteorder="big",
+                    signed=False,
+                )
+                if len(dynamic_segments) > 0:
+                    dynamic_segments[len(dynamic_segments) - 1][
```
jasonpaulos, line 256: nit: `dynamic_segments[-1][1] = dynamic_index` would be shorter

algosdk/abi/tuple_type.py
```diff
@@ -0,0 +1,329 @@
...
+
+                    if before % 8 != 0:
+                        raise error.ABIEncodingError(
+                            "expected before index should have number of bool mod 8 equal 0"
+                        )
+                    after = min(7, after)
+                    # Parse bool values into multiple byte strings
+                    for bool_i in range(after + 1):
+                        mask = 128 >> bool_i
+                        bit = int.from_bytes(
```
jasonpaulos, line 281: nit: you could compuate `bit` before entering the loop

algosdk/abi/bool_type.py
```diff
@@ -32,41 +32,41 @@ def encode(self, value):
...
         return b"\x00"
 
-    def decode(self, bool_string):
+    def decode(self, bytestring):
         """
         Decodes a bytestring to a single boolean.
 
         Args:
-            bool_string (bytes | bytearray): bytestring to be decoded that contains a single boolean, i.e. \x80 or \x00
+            bytestring (bytes | bytearray): bytestring to be decoded that contains a single boolean, i.e. \x80 or \x00
```
jasonpaulos, line 48: nit: also here
```suggestion
            bytestring (bytes | bytearray): bytestring to be decoded that contains a single boolean, i.e. "0x80" or "0x00"
```

algosdk/abi/tuple_type.py
```diff
@@ -275,13 +266,13 @@ def decode(self, tuple_string):
                             "expected before index should have number of bool mod 8 equal 0"
                         )
                     after = min(7, after)
+                    bit = int.from_bytes(
+                        bytestring[array_index : array_index + 1],
+                        byteorder="big",
+                    )
```
jasonpaulos, line 272: nit: actually since you're only extracting 1 byte, `bit = bytestring[array_index]` would be sufficient. `bits` might be a better name though

jasonpaulos (approved):
Looks good! I left a few nits which you can choose to accept or not. Pending tests, this looks good to merge

PR #243, opened 2021-10-25
Change __init__.py to include v2client import

StylishTriangles:
I propose this change to help language servers with their work. Unless there is a good reason to keep v2 out of the \_\_init\_\_ file, I think the v2client should be added too.


StylishTriangles:
Can we get this merged?

jasonpaulos (approved):
Looks like a good change, I apologize for the delay

PR #244, opened 2021-10-25
New keyreg txn field

shiqizng:
This PR adds a new field, stateProofKey, to key reg txn to match the changes in algod.


algosdk/future/transaction.py
```diff
@@ -460,6 +483,7 @@ def __init__(
         lease=None,
         rekey_to=None,
         nonpart=None,
+        state_proof_ID=None,
```
jasonpaulos, line 486: nit: this SDK uses the msgpack name of the fields in most places, so `sprfkey` would be more appropriate

algosdk/future/transaction.py
```diff
@@ -489,6 +515,8 @@ def dictify(self):
             d["votelst"] = self.votelst
         if self.nonpart is not None:
             d["nonpart"] = self.nonpart
+        if self.state_proof_ID is not None:
+            d["sprfkey"] = self.state_proof_ID
```
jasonpaulos, line 519: Can you treat this as a base64 encoded string, similar to `selkey` and `votepk`?

algosdk/future/transaction.py
```diff
@@ -588,6 +622,8 @@ def __init__(
             raise error.KeyregOnlineTxnInitError("votelst")
         if votekd is None:
             raise error.KeyregOnlineTxnInitError("votekd")
+        if state_proof_ID is None:
+            raise error.KeyregOnlineTxnInitError("state_proof_ID")
```
jasonpaulos, line 626: In order to maintain compatibility before state proof keys are required, I think this should allow the state proof key to be None.

algosdk/future/transaction.py
```diff
@@ -600,12 +636,15 @@ def _undictify(d):
         votefst = d["votefst"]
         votelst = d["votelst"]
         votekd = d["votekd"]
+        sprfID = d["sprfkey"]
```
jasonpaulos, line 639: Similar thing here. If you attempted to parse an old keyreg txn before state proof keys were introduced, this would raise a KeyError.

jasonpaulos (commented):
Just some comments on the model file, which I don't think is used anyway

jasonpaulos (commented):
Looks good. Though once stateproofs are merged into master, I would like to do a quick review again

winder (approved):

PR #247, opened 2021-10-27
ABI Interaction Support for Python SDK

### LINKED ISSUE #242: ABI Interaction - off-chain programs invoking on-chain contracts
As described in [this design document](https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a), we build support in the SDK for off-chain programs invoking on-chain contracts.

algochoi:
This PR enables developers to invoke on-chain contract methods using off-chain applications, as defined by ARC-0004.

- [x] Define an object for an ABI `method` description and expose the signature and selector
- [x] Define an object for an ABI `interface` description
- [x] Define an object for an ABI `contract` description
- [x] Define and implement `AtomicTransactionComposer`
- [x] Merge signatures for `TransactionSigners`
- [x] Parse logs for return arguments
- [x] Add cucumber tests and add tags to the `Makefile`
- [ ] Support foreign objects as ABI arguments (Will require follow up ticket)

Closes #242

### DIFF
--- a/algosdk/abi/base_type.py
+++ b/algosdk/abi/base_type.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from enum import IntEnum
 
 # Globals
 ABI_LENGTH_SIZE = 2  # We use 2 bytes to encode the length of a dynamic element
--- a/algosdk/abi/base_type.py
+++ b/algosdk/abi/base_type.py
@@ -44,7 +44,7 @@ class Type(ABC):
         pass
 
     @abstractmethod
-    def decode(self, value_string):
+    def decode(self, bytestring):
         """
         Deserialize the ABI type and value from a byte string using ABI encoding rules.
         """
--- a/algosdk/abi/base_type.py
+++ b/algosdk/abi/base_type.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 ABI_LENGTH_SIZE = 2  # We use 2 bytes to encode the length of a dynamic element
 
 
-class Type(ABC):
+class ABIType(ABC):
     """
     Represents an ABI Type for encoding.
     """
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -21,7 +21,11 @@ class Contract:
     def __eq__(self, o: object) -> bool:
         if not isinstance(o, Contract):
             return False
-        return self.name == o.name and self.app_id == o.app_id and self.methods == o.methods
+        return (
+            self.name == o.name
+            and self.app_id == o.app_id
+            and self.methods == o.methods
+        )
 
     @staticmethod
     def from_json(resp):
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -14,6 +14,7 @@ class Contract:
         self.app_id = app_id
         self.methods = methods
 
+    @staticmethod
     def from_json(resp):
         d = json.loads(resp)
         name = d["name"]
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -1,7 +1,6 @@
 import json
 
-from .method import Method
-from .. import error
+from algosdk.abi.method import Method
 
 
 class Contract:
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -6,6 +6,11 @@ from algosdk.abi.method import Method
 class Contract:
     """
     Represents a ABI contract description.
+
+    Args:
+        name (string): name of the contract
+        app_id (int): application id associated with the contract
+        methods (list): list of Method objects
     """
 
     def __init__(self, name, app_id, methods) -> None:
@@ -16,6 +21,17 @@ class Contract:
     @staticmethod
     def from_json(resp):
         d = json.loads(resp)
+        return Contract.undictify(d)
+
+    def dictify(self):
+        d = {}
+        d["name"] = self.name
+        d["app_id"] = self.app_id
+        d["methods"] = [m.dictify() for m in self.methods]
+        return d
+
+    @staticmethod
+    def undictify(d):
         name = d["name"]
         app_id = d["app_id"]
         method_list = [Method.undictify(method) for method in d["methods"]]
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -15,8 +15,13 @@ class Contract:
 
     def __init__(self, name, app_id, methods) -> None:
         self.name = name
-        self.app_id = app_id
+        self.app_id = int(app_id)
         self.methods = methods
+
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Contract):
+            return False
+        return self.name == o.name and self.app_id == o.app_id and self.methods == o.methods
 
     @staticmethod
     def from_json(resp):
@@ -26,13 +31,13 @@ class Contract:
     def dictify(self):
         d = {}
         d["name"] = self.name
-        d["app_id"] = self.app_id
+        d["appId"] = self.app_id
         d["methods"] = [m.dictify() for m in self.methods]
         return d
 
     @staticmethod
     def undictify(d):
         name = d["name"]
-        app_id = d["app_id"]
+        app_id = d["appId"]
         method_list = [Method.undictify(method) for method in d["methods"]]
         return Contract(name=name, app_id=app_id, methods=method_list)
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -18,7 +18,7 @@ class Contract:
         self.app_id = int(app_id)
         self.methods = methods
 
-    def __eq__(self, o: object) -> bool:
+    def __eq__(self, o) -> bool:
         if not isinstance(o, Contract):
             return False
         return (
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -1,0 +1,22 @@
+import json
+
+from .method import Method
+from .. import error
+
+
+class Contract:
+    """
+    Represents a ABI contract description.
+    """
+
+    def __init__(self, name, app_id, methods) -> None:
+        self.name = name
+        self.app_id = app_id
+        self.methods = methods
+
+    def from_json(resp):
+        d = json.loads(resp)
+        name = d["name"]
+        app_id = d["app_id"]
+        method_list = [Method.undictify(method) for method in d["methods"]]
+        return Contract(name=name, app_id=app_id, methods=method_list)
--- a/algosdk/abi/interface.py
+++ b/algosdk/abi/interface.py
@@ -1,0 +1,21 @@
+import json
+
+from .method import Method
+from .. import error
+
+
+class Interface:
+    """
+    Represents a ABI interface description.
+    """
+
+    def __init__(self, name, methods) -> None:
+        self.name = name
+        self.methods = methods
+
+    @staticmethod
+    def from_json(resp):
+        d = json.loads(resp)
+        name = d["name"]
+        method_list = [Method.undictify(method) for method in d["methods"]]
+        return Interface(name=name, methods=method_list)
--- a/algosdk/abi/interface.py
+++ b/algosdk/abi/interface.py
@@ -12,11 +12,11 @@ class Interface:
         methods (list): list of Method objects
     """
 
-    def __init__(self, name, methods) -> None:
+    def __init__(self, name, methods):
         self.name = name
         self.methods = methods
 
-    def __eq__(self, o: object) -> bool:
+    def __eq__(self, o):
         if not isinstance(o, Interface):
             return False
         return self.name == o.name and self.methods == o.methods
--- a/algosdk/abi/interface.py
+++ b/algosdk/abi/interface.py
@@ -1,7 +1,6 @@
 import json
 
-from .method import Method
-from .. import error
+from algosdk.abi.method import Method
 
 
 class Interface:
@@ -17,5 +16,6 @@ class Interface:
     def from_json(resp):
         d = json.loads(resp)
         name = d["name"]
+        # TODO: Should we check that method name does not begin with an underscore here?
         method_list = [Method.undictify(method) for method in d["methods"]]
         return Interface(name=name, methods=method_list)
--- a/algosdk/abi/interface.py
+++ b/algosdk/abi/interface.py
@@ -16,6 +16,11 @@ class Interface:
         self.name = name
         self.methods = methods
 
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Interface):
+            return False
+        return self.name == o.name and self.methods == o.methods
+
     @staticmethod
     def from_json(resp):
         d = json.loads(resp)
--- a/algosdk/abi/interface.py
+++ b/algosdk/abi/interface.py
@@ -6,6 +6,10 @@ from algosdk.abi.method import Method
 class Interface:
     """
     Represents a ABI interface description.
+
+    Args:
+        name (string): name of the interface
+        methods (list): list of Method objects
     """
 
     def __init__(self, name, methods) -> None:
@@ -15,7 +19,16 @@ class Interface:
     @staticmethod
     def from_json(resp):
         d = json.loads(resp)
+        return Interface.undictify(d)
+
+    def dictify(self):
+        d = {}
+        d["name"] = self.name
+        d["methods"] = [m.dictify() for m in self.methods]
+        return d
+
+    @staticmethod
+    def undictify(d):
         name = d["name"]
-        # TODO: Should we check that method name does not begin with an underscore here?
         method_list = [Method.undictify(method) for method in d["methods"]]
         return Interface(name=name, methods=method_list)
--- a/algosdk/abi/method.py
+++ b/algosdk/abi/method.py
@@ -2,10 +2,7 @@ import json
 
 from Cryptodome.Hash import SHA512
 
-from algosdk.abi.tuple_type import TupleType
-from algosdk.abi.util import type_from_string
-from algosdk import error
-from algosdk import constants
+from algosdk import abi, constants, error
 
 
 TRANSACTION_ARGS = (
@@ -108,7 +105,9 @@ class Method:
         # the second token should be the arguments as a tuple,
         # and the last token should be the return type (or void).
         tokens = Method._parse_string(s)
-        argument_list = [Argument(t) for t in TupleType.parse_tuple(tokens[1])]
+        argument_list = [
+            Argument(t) for t in abi.TupleType.parse_tuple(tokens[1])
+        ]
         return_type = Returns(tokens[-1])
         return Method(name=tokens[0], args=argument_list, returns=return_type)
 
@@ -148,7 +147,7 @@ class Argument:
             self.type = arg_type
         else:
             # If the type cannot be parsed into an ABI type, it will error
-            self.type = type_from_string(arg_type)
+            self.type = abi.util.type_from_string(arg_type)
         self.name = name
         self.desc = desc
 
@@ -194,7 +193,7 @@ class Returns:
             self.type = arg_type
         else:
             # If the type cannot be parsed into an ABI type, it will error
-            self.type = type_from_string(arg_type)
+            self.type = abi.util.type_from_string(arg_type)
         self.desc = desc
 
     def __eq__(self, o: object) -> bool:
--- a/algosdk/abi/method.py
+++ b/algosdk/abi/method.py
@@ -9,7 +9,7 @@ from algosdk import constants
 
 
 # Globals
-TRANSACTION_ARG = (
+TRANSACTION_ARGS = (
     constants.PAYMENT_TXN,
     constants.KEYREG_TXN,
     constants.ASSETCONFIG_TXN,
@@ -42,7 +42,7 @@ class Method:
         # add one for this method call itself.
         txn_count = 1
         for arg in self.args:
-            if arg.type in TRANSACTION_ARG:
+            if arg.type in TRANSACTION_ARGS:
                 txn_count += 1
         self.txn_calls = txn_count
 
@@ -127,7 +127,7 @@ class Argument:
     """
 
     def __init__(self, arg_type, name=None, desc=None) -> None:
-        if arg_type in TRANSACTION_ARG:
+        if arg_type in TRANSACTION_ARGS:
             self.type = arg_type
         else:
             # If the type cannot be parsed into an ABI type, it will error
--- a/algosdk/abi/method.py
+++ b/algosdk/abi/method.py
@@ -8,7 +8,6 @@ from algosdk import error
 from algosdk import constants
 
 
-# Globals
 TRANSACTION_ARGS = (
     constants.PAYMENT_TXN,
     constants.KEYREG_TXN,
@@ -27,11 +26,11 @@ class Method:
         name (string): name of the method
         args (list): list of Argument objects with type, name, and optional
         description
-        returns (Returns): a Returns object with a type and optional description
+        returns (Returns, optional): a Returns object with a type and optional description
         desc (string, optional): optional description of the method
     """
 
-    def __init__(self, name, args, returns, desc=None) -> None:
+    def __init__(self, name, args, returns=None, desc=None) -> None:
         self.name = name
         self.args = args
         self.desc = desc
@@ -70,6 +69,25 @@ class Method:
         return self.txn_calls
 
     @staticmethod
+    def _parse_string(s):
+        # Parses a method signature into three tokens, returned as a list:
+        # e.g. 'a(b,c)d' -> ['a', 'b,c', 'd']
+        stack = []
+        for i, char in enumerate(s):
+            if char == "(":
+                stack.append(i)
+            elif char == ")":
+                if not stack:
+                    break
+                left_index = stack.pop()
+                if not stack:
+                    return (s[:left_index], s[left_index + 1 : i], s[i + 1 :])
+
+        raise error.ABIEncodingError(
+            "ABI method string has mismatched parentheses: {}".format(s)
+        )
+
+    @staticmethod
     def from_json(resp):
         method_dict = json.loads(resp)
         return Method.undictify(method_dict)
@@ -85,6 +103,16 @@ class Method:
         return_type = Returns(tokens[-1])
         return Method(name=tokens[0], args=argument_list, returns=return_type)
 
+    def dictify(self):
+        d = {}
+        d["name"] = self.name
+        d["args"] = [arg.dictify() for arg in self.args]
+        if self.returns:
+            d["returns"] = self.returns.dictify()
+        if self.desc:
+            d["desc"] = self.desc
+        return d
+
     @staticmethod
     def undictify(d):
         name = d["name"]
@@ -95,33 +123,13 @@ class Method:
         desc = d["desc"] if "desc" in d else None
         return Method(name=name, args=arg_list, returns=return_obj, desc=desc)
 
-    @staticmethod
-    def _parse_string(s):
-        # Parses a method signature into three tokens, returned as a list:
-        # e.g. 'a(b,c)d' -> ['a', 'b,c', 'd']
-        stack = []
-        for i, char in enumerate(s):
-            if char == "(":
-                stack.append(i)
-            elif char == ")":
-                if not stack:
-                    break
-                left_index = stack[-1]
-                stack.pop()
-                if not stack:
-                    return (s[:left_index], s[left_index + 1 : i], s[i + 1 :])
-
-        raise error.ABIEncodingError(
-            "ABI method string has mismatched parentheses: {}".format(s)
-        )
-
 
 class Argument:
     """
     Represents an argument for a ABI method
 
     Args:
-        arg_type (string | Type): ABI type or transaction of the method argument
+        arg_type (string): ABI type or transaction string of the method argument
         name (string, optional): name of this method argument
         desc (string, optional): description of this method argument
     """
@@ -137,6 +145,15 @@ class Argument:
 
     def __str__(self):
         return str(self.type)
+
+    def dictify(self):
+        d = {}
+        d["type"] = str(self.type)
+        if self.name:
+            d["name"] = self.name
+        if self.desc:
+            d["desc"] = self.desc
+        return d
 
     @staticmethod
     def undictify(d):
@@ -166,6 +183,13 @@ class Returns:
 
     def __str__(self):
         return str(self.type)
+
+    def dictify(self):
+        d = {}
+        d["type"] = str(self.type)
+        if self.desc:
+            d["desc"] = self.desc
+        return d
 
     @staticmethod
     def undictify(d):
--- a/algosdk/abi/method.py
+++ b/algosdk/abi/method.py
@@ -34,9 +34,7 @@ class Method:
         self.name = name
         self.args = args
         self.desc = desc
-        self.returns = (
-            returns if (returns and returns.type != "void") else None
-        )
+        self.returns = returns
         # Calculate number of method calls passed in as arguments and
         # add one for this method call itself.
         txn_count = 1
@@ -44,6 +42,17 @@ class Method:
             if arg.type in TRANSACTION_ARGS:
                 txn_count += 1
         self.txn_calls = txn_count
+
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Method):
+            return False
+        return (
+            self.name == o.name
+            and self.args == o.args
+            and self.returns == o.returns
+            and self.desc == o.desc
+            and self.txn_calls == o.txn_calls
+        )
 
     def get_signature(self):
         arg_string = ",".join([str(arg.type) for arg in self.args])
@@ -143,6 +152,13 @@ class Argument:
         self.name = name
         self.desc = desc
 
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Argument):
+            return False
+        return (
+            self.name == o.name and self.type == o.type and self.desc == o.desc
+        )
+
     def __str__(self):
         return str(self.type)
 
@@ -159,7 +175,7 @@ class Argument:
     def undictify(d):
         return Argument(
             arg_type=d["type"],
-            name=d["name"],
+            name=d["name"] if "name" in d else None,
             desc=d["desc"] if "desc" in d else None,
         )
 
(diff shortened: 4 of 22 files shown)


algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        return len(self.txn_list)
+
+    def clone(self):
+        """
+        Creates a new composer with the same underlying transactions.
+        The new composer's status will be BUILDING, so additional transactions
+        may be added to it.
+        """
+        cloned = copy.deepcopy(self)
```
algoidurovic, line 71: Shouldn't you remove the groupid if it's set for the transactions?

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+            list[SignedTransactions]: list of signed transactions
+        """
+        if self.status >= AtomicTransactionComposerStatus.SIGNED:
+            # Return cached versions of the signatures
+            return self.signed_txns
+
+        txn_list = self.build_group()
+        for txn_with_signer in txn_list:
+            unsigned_txn = txn_with_signer.txn
+            stxn = txn_with_signer.signer.sign_txn(unsigned_txn)
```
algoidurovic, line 240: I think you want to verify that the signing was successful

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+        The composer's status must be SUBMITTED or lower before calling this method.
+        If submission is successful, this composer's status will update to SUBMITTED.
+
+        Note: a group can only be submitted again if it fails.
+
+        Returns:
+            list[Transaction]: list of submitted transactions
+        """
+        if self.status <= AtomicTransactionComposerStatus.SUBMITTED:
+            self.build_group()
```
algoidurovic, line 259: nit: gather_signatures() already calls build_group()

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+                results for each method call transaction in this group. If a
+                method has no return value (void), then the method results array
+                will contain None for that method's return value.
+        """
+        if self.status > AtomicTransactionComposerStatus.SUBMITTED:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposerStatus must be submitted or lower to execute a group"
+            )
+
+        client.submit(client)
```
algoidurovic, line 293: This might resend an already sent tx, if I'm understanding this correctly

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+        client.wait_for_confirmation(client, self.tx_ids[0])
+
+        resp = client.pending_transaction_info(self.tx_ids[0])
+        resp_dict = json.loads(resp)
+
+        confirmed_round = resp_dict["confirmed-round"]
+        tx_ids = self.tx_ids
+        results = resp_dict["logs"] if "logs" in resp_dict else None
```
algoidurovic, line 301: From Jason: "We have not finalized this 100% yet, but the plan is to have the contract log the return value as encoded bytes, likely with a special 4-byte prefix (maybe all 0s). So you can use the pending txn endpoint to see the logged values, then iterate over them to find an entry that starts with 0x00000000, and decode the rest of it to get the ABI return value"

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,373 @@
...
+    Creates a TransactionSigner for a basic account or LogicSigAccount
+    """
+    return TransactionSigner(private_key)
+
+
+def make_multisig_account_transaction_signer(msig, sks):
+    """
+    Creates a TransactionSigner for a MultiSig account
+    """
+    return TransactionSigner(private_key=sks, msig=msig)
```
ahangsu, line 373: nit: I guess you can put them into `TransactionSigner` as a class method? See: https://stackoverflow.com/questions/682504/what-is-a-clean-pythonic-way-to-have-multiple-constructors-in-python

algosdk/future/atomic_transaction_composer.py
ahangsu: I think the solution can add if condition to `submit` and check if equal to `SUBMITTED` then just return txid, which eliminates the duplicate sending.

algosdk/future/atomic_transaction_composer.py
algochoi: Thanks! Wasn't aware of this -- I'll change this

jasonpaulos (commented):
I guess the spec doesn't say anything about this currently, but could you add methods to re-encode the `Method`, `Interface`, and `Contract` objects to JSON? For symmetry this would be good to have.

algosdk/abi/method.py
```diff
@@ -0,0 +1,174 @@
...
+            "ABI method string has mismatched parentheses: {}".format(s)
+        )
+
+
+class Argument:
+    """
+    Represents an argument for a ABI method
+
+    Args:
+        arg_type (string | Type): ABI type or transaction of the method argument
```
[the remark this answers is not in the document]
algochoi: Ah, good catch. The argument itself will always be a string.

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,496 @@
...
+        An error will be thrown if signing any of the transactions fails.
+
+        Returns:
+            list[SignedTransactions]: list of signed transactions
+        """
+        if self.status >= AtomicTransactionComposerStatus.SIGNED:
+            # Return cached versions of the signatures
+            return self.signed_txns
+
+        stxn_list = [None] * len(self.txn_list)
```
algochoi, line 255: The approach I took to gathering signatures was:
1. Loop through the transaction list and make a map that maps a unique signer to a list of transaction indexes.
2. Loop through the map and call the `sign` method on the transaction list and indexes.
3. After receiving the signed transactions for the indexes, assign it to a fixed array of signed transactions to return.
Open to discussing more elegant solutions.

algosdk/abi/method.py
```diff
@@ -0,0 +1,218 @@
+import json
+
+from Cryptodome.Hash import SHA512
+
+from algosdk import abi, constants, error
+
+
+TRANSACTION_ARGS = (
+    constants.PAYMENT_TXN,
```
jasonpaulos, line 9: According to the latest ARC-4, [`txn` should also be acceptable](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#transaction-types)

algosdk/abi/tuple_type.py
```diff
@@ -71,7 +71,7 @@ def _find_bool(type_list, index, delta):
         return until
 
     @staticmethod
-    def _parse_tuple(s):
```
jasonpaulos, line 74: nit: I think this can still be an underscore method, since it's only supposed to be used internally

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,512 @@
...
+                        return_value=None,
+                        decode_error=None,
+                    )
+                )
+                continue
+
+            # Parse log for ABI method return value
+            resp = client.pending_transaction_info(tx_id)
+            confirmed_round = resp["confirmed-round"]
+            logs = resp["logs"] if "logs" in resp else None
```
jasonpaulos, line 344: Use `[]` instead of `None` here? Otherwise the for loop below won't work properly

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,512 @@
...
+                        tx_id=tx_id,
+                        raw_value=None,
+                        return_value=None,
+                        decode_error=None,
+                    )
+                )
+                continue
+
+            # Parse log for ABI method return value
+            resp = client.pending_transaction_info(tx_id)
```
jasonpaulos, line 342: It would better if you can expand the `try` block to include this call too, since it might error due to network issues.

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,512 @@
...
+                if result_bytes[:4] != bytes.fromhex(ABI_RETURN_HASH):
+                    continue
+                try:
+                    raw_value = result_bytes[4:]
+                    return_value = self.method_dict[i].returns.type.decode(
+                        raw_value
+                    )
+                    decode_error = None
+                except Exception as e:
+                    raw_value = None
```
jasonpaulos, line 361: No need to erase `raw_value` if an exception happens, since if the type decoding fails, it's still worthwhile to return the raw value

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,512 @@
...
+
+class TransactionSigner(ABC):
+    """
+    Represents an object which can sign transactions from an atomic transaction group.
+    """
+
+    def __init__(self) -> None:
+        pass
+
+    def sign(self, txn_group, indexes):
```
jasonpaulos, line 393: This should use `@abstractmethod` so all subclasses are forced to override it

Also, how about a name like `sign_transactions` instead? At some point we may want to introduce signers for other objects, and if they all use `sign` it won't be possible for subclasses to implement multiple interfaces.

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,512 @@
...
+        corresponds to the signed transaction from txnGroup[indexesToSign[i]].
+
+        Args:
+            txn_group (list[Transaction]): atomic group of transactions
+            indexes (list[int]): array of indexes in the atomic transaction group that should be signed
+        """
+        stxns = []
+        for i in indexes:
+            for sk in self.sks:
+                stxn = txn_group[i].sign(sk)
```
jasonpaulos, line 489: This should actuall use `transaction.MultisigTransaction` to sign. See an example here: https://github.com/algorand/py-algorand-sdk/blob/8cd3bb615ca3c1c15846bd9eb7ec180ac4267fc4/examples/multisig_example.py#L26-L31

test/steps/v2_steps.py
```diff
@@ -1688,6 +1711,27 @@ def split_and_process_app_args(in_args):
...
+    if receiver == "transient":
+        receiver = context.transient_pk
+    if not close_remainder_to:
+        close_remainder_to = None
+    context.transaction = transaction.PaymentTxn(
+        sender=sender,
+        sp=context.suggested_params,
+        receiver=receiver,
+        amt=int(amount),
+        # close_remainder_to=close_remainder_to,
```
jasonpaulos, line 1731: Uncomment this?

test/steps/v2_steps.py
```diff
@@ -1799,10 +1843,25 @@ def sign_transaction_with_signing_account(context):
     context.signed_transaction = context.transaction.sign(private_key)
 
 
+@then('the base64 encoded signed transactions should equal "{goldens}"')
+def compare_stxns_array_to_base64_golden(context, goldens):
+    golden_strings = goldens.split(",")
```
jasonpaulos, line 1848: Probably a good idea to compare lengths as well

Or just convert `context.signed_transactions` to an encoded array and compare that to `golden_strings`

test/steps/v2_steps.py
```diff
@@ -2264,3 +2340,322 @@ class TestCase(DryrunTestCaseMixin, unittest.TestCase):
...
+def append_txn_to_method_args(context):
+    context.method_args.append(context.transaction_with_signer)
+
+
+@when(
+    'I append the encoded arguments "{method_args:MaybeString}" to the method arguments array.'
+)
+def append_app_args_to_method_args(context, method_args):
+    # Returns a list of ABI method arguments
+    app_args = split_and_process_abi_args(context.abi_method, method_args)
```
jasonpaulos, line 2418: This works for now so it doesn't need to be changed in this PR, but I'd like this step to not assume that `method_args` match the arguments from `context.abi_method`.

Meaning it should be possible to call this step many times to append arguments, and then only in the `I add a method call...` step should the arguments be decoded. This will give us more flexibility to mix transaction args and normal args, like `(uint64,pay,uint64,axfer,uint64)`.

This is how I did in the JS SDK: https://github.com/algorand/js-algorand-sdk/blob/6cb22886b089544349448bc1097164dc70a085b8/tests/cucumber/steps/steps.js#L4750-L4791

algosdk/abi/tuple_type.py
algochoi: I decided to expose it so it can be used in the `Method` class to parse the tuple string: https://github.com/algorand/py-algorand-sdk/blob/d63b4d132a0b8c3aeb2b9001a3e87f596eb34423/algosdk/abi/method.py#L110

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Args:
+        name (string): name of the method
+        args (list): list of Argument objects with type, name, and optional
+        description
+        returns (Returns, optional): a Returns object with a type and optional description
+        desc (string, optional): optional description of the method
+    """
+
+    def __init__(self, name, args, returns=None, desc=None) -> None:
```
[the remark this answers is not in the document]
algochoi: Good point!

algosdk/abi/tuple_type.py
jasonpaulos: Right it can still be used there, but unless we want consumer of the library to use that method, it's still probably best for it to remain underscored.

algochoi: Good point, I agree with that

algosdk/future/atomic_transaction_composer.py
algochoi: Sounds good!

test/steps/v2_steps.py
algochoi: That's a good point, we should make the step as general as possible.

algosdk/future/atomic_transaction_composer.py
algochoi: Oops -- updated and manually tested the  multisig signatures again 😓

algosdk/atomic_transaction_composer.py
```diff
@@ -317,51 +332,53 @@ def execute(self, client, wait_rounds):
...
 
         for i, tx_id in enumerate(self.tx_ids):
+            raw_value = None
+            return_value = None
+            decode_error = None
+
             if i not in self.method_dict:
                 continue
             # Return is void
             if self.method_dict[i].returns is None:
```
jasonpaulos, line 351: Is this check still right?

jasonpaulos (commented):
Looks good, just one question

algosdk/atomic_transaction_composer.py
algochoi: Good catch, this should be changed as `returns` now has `void` and is always required. I will also strengthen the unit tests in the next commit.

jasonpaulos (approved):

algosdk/abi/base_type.py
```diff
@@ -45,7 +44,7 @@ def encode(self, value):
         pass
 
     @abstractmethod
-    def decode(self, value_string):
+    def decode(self, bytestring):
```
tzaffi, line 47: Sometimes we are using type hints, and sometimes we aren't. What is our roadmap in this regards, or is just ad-hoc at the moment?

algosdk/abi/interface.py
```diff
@@ -0,0 +1,39 @@
...
+class Interface:
+    """
+    Represents a ABI interface description.
+
+    Args:
+        name (string): name of the interface
+        methods (list): list of Method objects
+    """
+
+    def __init__(self, name, methods) -> None:
```
tzaffi, line 15: Continuing with my first comment, If we are gradually transitioning to using type hints, this new class presents an opportunity for that. If not, I'm not sure the purpose of having partial type hints. `__eq__` is a lonely outlier.

algosdk/abi/interface.py
```diff
@@ -0,0 +1,39 @@
...
+        self.name = name
+        self.methods = methods
+
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Interface):
+            return False
+        return self.name == o.name and self.methods == o.methods
```
tzaffi, line 22:
```suggestion
        return self.id == o.id or (self.name == o.name and self.methods == o.methods)
```
Not really required, just a handy trick that speeds up evaluation `x == x` (especially for nested structures). But it depends on the usage pattern as it does add an extra evaluation in other cases.

algosdk/abi/contract.py
```diff
@@ -0,0 +1,47 @@
+import json
```
tzaffi, line 1: ...similar suggestions as for `interface.py`

algosdk/abi/interface.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @staticmethod
```
tzaffi, line 24: This, as well as, `undictify()`, is a good candidate to become a `@classmethod`. That way you could write more generic versions that are a bit more robust to copy-pasta and refactorings. EG (assuming no type hints):

```python
@classmethod
def from_json(cls, resp):
    return cls.undictify(json.loads(resp))

...

@classmethod
def undictify(cls, resp):
    name = d["name"]
    method_list = [Method.undictify(method) for method in d["methods"]]
    return cls(name=name, methods=method_list)
```

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+        return (
+            self.name == o.name
+            and self.args == o.args
+            and self.returns == o.returns
+            and self.desc == o.desc
+            and self.txn_calls == o.txn_calls
+        )
+
+    def get_signature(self):
+        arg_string = ",".join([str(arg.type) for arg in self.args])
```
tzaffi, line 56:
```suggestion
        arg_string = ",".join(str(arg.type) for arg in self.args)
```
this works as well

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+    def get_selector(self):
+        """
+        Returns the ABI method signature, which is the first four bytes of the
+        SHA-512/256 hash of the method signature.
+
+        Returns:
+            bytes: first four bytes of the method signature hash
+        """
+        hash = SHA512.new(truncate="256")
+        hash.update((self.get_signature()).encode("utf-8"))
```
tzaffi, line 69:
```suggestion
        hash.update(self.get_signature().encode("utf-8"))
```

algosdk/abi/base_type.py
algochoi: Good observation, I think PyTeal uses more type hints (it also has `mypy` checks), but I don't think our SDK does it right now. I think it would be nice for the future though.

algosdk/abi/interface.py
algochoi: Agree with your point, I will clean this up

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
+import json
```
tzaffi, line 1: Similar remarks regarding type hints and `@classmethod` vs. `@staticmethod` as previous.

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+    def get_txn_calls(self):
+        """
+        Returns the number of transactions needed to invoke this ABI method.
+        """
+        return self.txn_calls
+
+    @staticmethod
+    def _parse_string(s):
+        # Parses a method signature into three tokens, returned as a list:
+        # e.g. 'a(b,c)d' -> ['a', 'b,c', 'd']
```
tzaffi, line 81: Will we ever get nested parens?

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+        stack = []
+        for i, char in enumerate(s):
+            if char == "(":
+                stack.append(i)
```
tzaffi, line 85: If no nesting, can avoid the `stack`

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+                    break
+                left_index = stack.pop()
+                if not stack:
+                    return (s[:left_index], s[left_index + 1 : i], s[i + 1 :])
+
+        raise error.ABIEncodingError(
+            "ABI method string has mismatched parentheses: {}".format(s)
+        )
+
+    @staticmethod
```
tzaffi, line 97: `@classmethod` instead?

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        stxns = []
+        for i in indexes:
+            mtxn = transaction.MultisigTransaction(txn_group[i], self.msig)
+            for sk in self.sks:
+                mtxn.sign(sk)
+            stxns.append(mtxn)
+        return stxns
+
+
```
tzaffi, line 510: These method-less classes are good candidates for becoming [data classes](https://realpython.com/python-data-classes/). Just a thought.

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        Creates a new composer with the same underlying transactions.
+        The new composer's status will be BUILDING, so additional transactions
+        may be added to it.
+        """
+        cloned = AtomicTransactionComposer()
+        cloned.method_dict = copy.deepcopy(self.method_dict)
+        cloned.txn_list = copy.deepcopy(self.txn_list)
+        for t in cloned.txn_list:
+            t.txn.group = None
```
tzaffi, line 78: To me the name `clone` suggests an exact copy of the cloned object with all of its current state, but here the transaction groups are being nulled out. So maybe a more apropos name for the method would be something like `clone_skeleton()` or maybe just a dunder version that's not meant to be used externally such as `_clone()` ?

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        Args:
+            txn_and_signer (TransactionWithSigner)
+        """
+        if self.status != AtomicTransactionComposerStatus.BUILDING:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposer must be in BUILDING state for a transaction to be added"
+            )
+        if len(self.txn_list) == self.MAX_GROUP_SIZE:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposer cannot exceed MAX_GROUP_SIZE transactions"
```
tzaffi, line 99:
```suggestion
                f"AtomicTransactionComposer cannot exceed MAX_GROUP_SIZE ({MAX_GROUP_SIZE}) transactions"
```
could be helpful to add the actual max in an `f-string`

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        cloned = AtomicTransactionComposer()
+        cloned.method_dict = copy.deepcopy(self.method_dict)
+        cloned.txn_list = copy.deepcopy(self.txn_list)
+        for t in cloned.txn_list:
+            t.txn.group = None
+        cloned.status = AtomicTransactionComposerStatus.BUILDING
+        return cloned
+
+    def add_transaction(self, txn_and_signer):
```
tzaffi, line 82: Random thought... if we want to enable a "builder pattern" type approach, this `add_XYZ` methods return `self`. So you could do something like:
```python
atc = AtomicTransactionComposer()
atc.add_transaction(... a transaction ...)
     .add_method(... a method ...)
     .add_method(... another method ...)
```

etc.

But the builder pattern is a controversial one...

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+            rekey_to (str, optional): additionally rekey the sender to this address
+
+        """
+        if self.status != AtomicTransactionComposerStatus.BUILDING:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposer must be in BUILDING state for a transaction to be added"
+            )
+        if len(self.txn_list) + method.get_txn_calls() > self.MAX_GROUP_SIZE:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposer cannot exceed MAX_GROUP_SIZE transactions"
```
tzaffi, line 151:
```suggestion
               f"AtomicTransactionComposer cannot exceed MAX_GROUP_SIZE ({MAX_GROUP_SIZE}) transactions"
```

f-string version?

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        for i, arg in enumerate(method.args):
+            if arg.type in abi.method.TRANSACTION_ARGS:
+                if not isinstance(method_args[i], TransactionWithSigner):
+                    raise error.AtomicTransactionComposerError(
+                        "expected TransactionWithSigner as method argument, but received: {}".format(
+                            method_args[i]
+                        )
+                    )
+                txn_list.append(method_args[i])
+            elif len(app_args) > 14:
```
tzaffi, line 186: 14 is a good candidate for a constant.

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        app_args.append(method.get_selector())
+        # Iterate through the method arguments and either pack a transaction
+        # or encode a ABI value.
+        for i, arg in enumerate(method.args):
+            if arg.type in abi.method.TRANSACTION_ARGS:
+                if not isinstance(method_args[i], TransactionWithSigner):
+                    raise error.AtomicTransactionComposerError(
+                        "expected TransactionWithSigner as method argument, but received: {}".format(
+                            method_args[i]
+                        )
```
tzaffi, line 183:
```suggestion
                        "expected TransactionWithSigner as method argument, but received: {arg}"
```
f-string version

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        app_args = []
+        # For more than 14 args, including the selector, compact them into a tuple
+        additional_args = []
+        additional_types = []
+        txn_list = []
+        # First app arg must be the selector of the method
+        app_args.append(method.get_selector())
+        # Iterate through the method arguments and either pack a transaction
+        # or encode a ABI value.
+        for i, arg in enumerate(method.args):
```
tzaffi, line 177:
```suggestion
        for arg in method.args:
```
don't really need `i` (but you might actually want to use it, in which case just ignore this suggestion)

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        txns = [t.txn for t in self.txn_list]
+        for signer, indexes in signer_indexes.items():
+            stxns = signer.sign_transactions(txns, indexes)
+            for i, stxn in enumerate(stxns):
+                index = indexes[i]
+                stxn_list[index] = stxn
+
+        if None in stxn_list:
+            raise error.AtomicTransactionComposerError(
+                "missing signatures, got {}".format(stxn_list)
```
tzaffi, line 274:
```suggestion
                f"missing signatures, got {stxn_list}"
```
f-string version

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+        else:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposerStatus must be submitted or lower to submit a group"
+            )
+
+        client.send_transactions(self.signed_txns)
+        self.status = AtomicTransactionComposerStatus.SUBMITTED
+        return self.tx_ids
+
+    def execute(self, client, wait_rounds):
```
tzaffi, line 307: Do we want to have a default number of `wait_rounds`. I suspect the answer is "no" but it's worth your 2-cents.

algosdk/abi/method.py
```diff
@@ -0,0 +1,219 @@
...
+class Returns:
+    """
+    Represents a return type for a ABI method
+
+    Args:
+        arg_type (string): ABI type of this return argument
+        desc (string, optional): description of this return argument
+    """
+
+    def __init__(self, arg_type, desc=None) -> None:
```
tzaffi, line 192: I believe "void" deserves it's own constant -at least- so it'll be easier for the linter to pick up any errors and avoid the hard-coded string "void" in other locations. For example, you could have something like
```python
class Returns:
    """ ... 
    """
    VOID = "void"
```
and below:
```python
        if arg_type == "void":
            self.type = self.VOID
```

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,529 @@
...
+
+        for i, tx_id in enumerate(self.tx_ids):
+            raw_value = None
+            return_value = None
+            decode_error = None
+
+            if i not in self.method_dict:
+                continue
+            # Return is void
+            if self.method_dict[i].returns.type == "void":
```
tzaffi, line 351:
```suggestion
            if self.method_dict[i].returns.type == RETURNS.VOID:
```

Assuming you took my `RETURNS.VOID` suggestion above....

tzaffi (commented):
Left various suggestions. All basically minor improvements or ignoreable.

algosdk/abi/method.py
algochoi: There could be nested tuples for arguments such as: https://github.com/algorand/py-algorand-sdk/blob/155fc9002f1b606843de10a50224f817f792327e/test_unit.py#L4207

As an aside, transaction arguments (`pay`, `appl`, etc.) or foreign objects (`apfa` ...) cannot be nested inside a parentheses, and the ABI encoding/typing method will throw an exception if someone tries to do so.

algosdk/atomic_transaction_composer.py
algochoi: I believe we have to (unfortunately) support Python 3.5 in this SDK: https://github.com/algorand/py-algorand-sdk/blob/8cd3bb615ca3c1c15846bd9eb7ec180ac4267fc4/setup.py#L21

This is also why we couldn't use f-strings (just one version short).

algosdk/atomic_transaction_composer.py
algochoi: I think `wait_rounds=0` could be nice (no timeout), but I just left it as is so it could be consistent with the other SDKs.

tzaffi: 👍

algosdk/atomic_transaction_composer.py
tzaffi: FYI, I was [conflating the Builder pattern with a Fluent Interface](https://stackoverflow.com/questions/12521712/what-is-the-pattern-used-in-java-and-elsewhere-to-set-multiple-properties-when/12521762)

algosdk/atomic_transaction_composer.py
jasonpaulos: I think `wait_rounds=0` is actually something you should never do in practice. algod currently has a bug where it will never return an error for a failed transaction in some cases (see https://github.com/algorand/go-algorand-internal/issues/1759), meaning your code would be blocked if you were unlucky enough to encounter that bug.

Besides, `wait_rounds=1000` is the maximum you should ever need to wait, since that's a max lifetime of a transaction. This _might_ be a good candidate for a default value if we want one, but it's still pretty large.

algosdk/atomic_transaction_composer.py
jasonpaulos: This seems like a useful improvement, since the functions currently don't return anything

algosdk/atomic_transaction_composer.py
algochoi: Interesting, I didn't know about that first point. I think `wait_for_confirmation()` in this SDK actually has a default value for `wait_rounds=0`: https://github.com/algochoi/py-algorand-sdk/blob/e444328616d99c88b830366dedcbf75e9795dcb3/algosdk/future/transaction.py#L3013

algosdk/atomic_transaction_composer.py
algochoi: I just need it for `method_args[i]`; `arg` is for `method.args`. I realize it's a bit visually confusing (originally, `method` was called `method_call` for some differentiation, but the name was misleading for a `Method` object).

algosdk/abi/method.py
tzaffi: I don't think the _arguments_ are nested as these are just `"(string,uint16),bool"`

algosdk/atomic_transaction_composer.py
tzaffi: ahhh dang it! 

So no `f-strings` either (only [since 3.6](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) )

algosdk/atomic_transaction_composer.py
tzaffi: Doesn't work now that I know we have to support 3.5. Will bring this up in standup.

algosdk/abi/method.py
algochoi: Fair, it could look something more like: https://github.com/algorand/py-algorand-sdk/blob/8cd3bb615ca3c1c15846bd9eb7ec180ac4267fc4/test_unit.py#L3720 

Funnily, we can use `f-strings` in the unit tests because they do not need to be externally exposed 😅

algosdk/atomic_transaction_composer.py
tzaffi: 👍

tzaffi:
Looks like you shouldn't follow my `f-string` or `@dataclass` suggestions since we are stuck on Python 3.5. I'll bring this up for discussion during standup.

algosdk/abi/method.py
algochoi: That look nice, I'll do that thanks

algosdk/abi/interface.py
algochoi: Good point, I think I will leave it as is for this PR - the existing `dictify` and `undictify` methods (in `Transaction` for instance) follow this so we can consider refactoring this in a later PR all at once.

algosdk/atomic_transaction_composer.py
algochoi: I can see your point. I think we chose `clone` for the name so users can "clone" the transactions so they can build on top of it. You can't actually re-send the exact copy of the transactions because the blockchain will reject them while a transaction with the same hash is still in the pool (every transaction hash should be unique in the blockchain anyway). So I think the `clone` here sufficiently implies that it is being "cloned" for building, because cloning everything wouldn't have much use cases, because you'd have to break them down again and build them back up.

tzaffi: 👍

tzaffi (approved):
LGTM

PR #249, opened 2021-11-03
Update AlgodClient

arfaoui47:


algoanne:
Thank you for the contribution, and sorry for missing it at the time of submission. 

We are currently thinking about redesigning our response types and this will feed into that conversation.

PR #251, opened 2021-11-17
Support Foreign objects as ABI arguments and address ARC-4 changes

### LINKED ISSUE #250: Support foreign objects as ABI arguments
## Summary
This is a follow up to #247, which added support for invoking ABI methods.

In addition to standard ABI type arguments, the SDK should support the special foreign object argument types `account`, `asset`, and `application`, [as described in ARC-4](https://github.com/algorandfoundation/ARCs/blob/6a9f91aed8068bbace872483a64cbf22d0e1c975/ARCs/arc-0004.md#foreign-apps-asas-and-accounts).

## Scope
Extend the `AtomicTransactionComposer.add_method_call` function to support these new method argument types. At a high level it should:

1. Recognize when an argument is a special type or a standard ABI type. The special argument types are only valid as top-level arguments, so direct string comparison should be sufficient.
2. If a special type is present, insert the argument value into the appropriate foreign array, and use the array index of the value as the app call argument, treated as an ABI `uint8`.

algochoi:
This PR adds support for Foreign accounts, apps, and assets as ABI arguments. In addition, it includes some 

- [x] Accept foreign objects in method signature and convert values from method argument array
- [x] Check foreign array limits
- [x] Implicitly convert duplicate or special (i.e. sender address or its own app id) values
- [x] Accept foreign arrays in `addMethodCall`
- [x] Changes to the ABI `Interface` and `Contract` from: https://github.com/algorandfoundation/ARCs/pull/55
- [x] Changes from ARC-0004: Change how contract creation and OnCompletion actions work algorandfoundation/ARCs#57
- [x] Testing

Closes #250


algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,533 @@
...
+
+        Args:
+            txn_group (list[Transaction]): atomic group of transactions
+            indexes (list[int]): array of indexes in the atomic transaction group that should be signed
+        """
+        stxns = []
+        for i in indexes:
+            for sk in self.sks:
+                stxn = txn_group[i].sign(sk)
+            stxns.append(stxn)
```
barnjamin, line 511: MultisigTransaction returns None from the sign call. 

Believe this should be
```py
      for sk in self.sks:
          txn_group[i].sign(sk)
      stxns.append(txn_group[i])
```

algosdk/future/atomic_transaction_composer.py
```diff
@@ -0,0 +1,533 @@
...
+                        foreign_apps.append(method_args[i])
+                else:
+                    current_type = arg.type
+                    current_arg = method_args[i]
+                if len(app_args) > 14:
+                    # Pack the remaining values as a tuple
+                    additional_types.append(current_type)
+                    additional_args.append(current_arg)
+                else:
+                    encoded_arg = arg.type.encode(current_arg)
```
barnjamin, line 206: Getting `TypeError: encode() argument 'encoding' must be str, not int`  here for an 'account' type. 

Looks like this attribute is a `str` not one of the ABI types  https://github.com/algorand/py-algorand-sdk/blob/algochoi/foreign-types/algosdk/abi/method.py#L149

algosdk/future/atomic_transaction_composer.py
algochoi: Thanks! I will change this in the other branch and merge the changes here soon....

algosdk/future/atomic_transaction_composer.py
algochoi: My bad, that should be: 
`encoded_arg = current_type.encode(current_arg)` 

I will push this change in the short term and add some more tests after the other branch is merged.

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,541 @@
...
+                        "expected TransactionWithSigner as method argument, but received: {}".format(
+                            method_args[i]
+                        )
+                    )
+                txn_list.append(method_args[i])
+            else:
+                if arg.type in abi.method.FOREIGN_ARRAY_ARGS:
+                    current_type = abi.UintType(8)
+                    if arg.type == "account":
+                        current_arg = len(foreign_accounts)
```
barnjamin, line 189: think this should be `len(foreign_accounts)+1` since the `sender` is implicitly the 0th

algosdk/atomic_transaction_composer.py
```diff
@@ -0,0 +1,541 @@
...
+                        foreign_accounts.append(method_args[i])
+                    elif arg.type == "asset":
+                        current_arg = len(foreign_assets)
+                        foreign_assets.append(method_args[i])
+                    else:
+                        current_arg = len(foreign_apps)
```
barnjamin, line 195: this should be `len(foreign_apps)+1` since current app id is implicitly 0th

barnjamin (changes requested):

algosdk/atomic_transaction_composer.py
algochoi: Good point

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
+                            foreign_assets.append(asset_arg)
+                    elif arg.type == abi.ABIReferenceType.APPLICATION:
+                        app_arg = int(method_args[i])
+                        if app_arg == app_id:
+                            current_arg = 0
+                        elif app_arg in foreign_apps:
+                            current_arg = foreign_apps.index(app_arg) + 1
+                        else:
+                            current_arg = len(foreign_apps) + 1
+                            foreign_apps.append(app_arg)
```
ahangsu, line 278: nit: would be nice to consider `populateForeignArray`-like method, since the code is somewhat redundant (but it is optional)

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
+                )
+            )
+
+        # Compact the arguments into a single tuple, if there are more than
+        # 15 arguments excluding the selector, into the last app arg slot.
+        if len(raw_types) > self.MAX_APP_ARG_LIMIT - 1:
+            additional_types = raw_types[14:]
+            additional_values = raw_values[14:]
+            raw_types = raw_types[:14]
+            raw_values = raw_values[:14]
```
ahangsu, line 316: nit: consider using `MAX_APP_ARG_LIMIT` rather than numeric constant

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
 
-        if additional_args:
-            remainder_args = abi.TupleType(additional_types).encode(
-                additional_args
+                raw_types.append(current_type)
+                raw_values.append(current_arg)
+
+        if len(accounts) > self.FOREIGN_ACCOUNT_LIMIT:
+            raise error.AtomicTransactionComposerError(
+                "foreign accounts exceed app limit {}".format(
```
ahangsu, line 296:
```suggestion
                "foreign accounts exceed account number limit {}".format(
```

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
+                    self.FOREIGN_ACCOUNT_LIMIT
+                )
             )
-            app_args.append(remainder_args)
+        if (
+            len(accounts) + len(foreign_apps) + len(foreign_assets)
+            > self.FOREIGN_ARRAY_LIMIT
+        ):
+            raise error.AtomicTransactionComposerError(
+                "foreign object array exceeds app limit {}".format(
```
ahangsu, line 305:
```suggestion
                "foreign object array exceeds foreign object number limit {}".format(
```

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
         # or encode a ABI value.
         for i, arg in enumerate(method.args):
-            if arg.type in abi.method.TRANSACTION_ARGS:
+            if abi.is_abi_transaction_type(arg.type):
                 if not isinstance(method_args[i], TransactionWithSigner):
                     raise error.AtomicTransactionComposerError(
                         "expected TransactionWithSigner as method argument, but received: {}".format(
                             method_args[i]
                         )
                     )
```
ahangsu, line 249: nit: I remembered in @jasonpaulos implementation, we also need to check the `TransactionWithSigner`'s transaction match with ABI method signature's arg type, e.g.,
```js
      if (abiTypeIsTransaction(argType)) {
        if (
          !isTransactionWithSigner(argValue) ||
          !abiCheckTransactionType(argType, argValue.txn)
        ) {
```
in https://github.com/algorand/js-algorand-sdk/pull/482/files#diff-534b9601ad1de4bb83d07dc0cb5af85424ccf602bcd8f4c0a2d3b79def06e8aeR319

ahangsu (commented):
Looks largely good to me, just some polishing here and there

barnjamin (approved):

ahangsu (commented):
Look largely good to me.

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
-                # Pack the remaining values as a tuple
-                additional_types.append(arg.type)
-                additional_args.append(method_args[i])
             else:
-                encoded_arg = arg.type.encode(method_args[i])
-                app_args.append(encoded_arg)
+                if abi.is_abi_reference_type(arg.type):
+                    current_type = abi.UintType(8)
+                    if arg.type == abi.ABIReferenceType.ACCOUNT:
+                        account_arg = str(method_args[i])
```
jasonpaulos, line 255: As we discussed, this should "normalize" the input before examinging it, since the input could be a 32-length byte array or a base32 address string

algosdk/abi/contract.py
```diff
@@ -10,22 +10,32 @@ class Contract:
...
+            such as app-id.
     """
 
-    def __init__(self, name: str, app_id: int, methods: List[Method]) -> None:
+    def __init__(
+        self,
+        name: str,
+        methods: List[Method],
+        desc: str = None,
+        networks: Dict[str, dict] = None,
```
jasonpaulos, line 24: I'd recommend making a new class to hold the network-specific values instead of using a dictionary, since it's more type safe. E.g. `networks: Dict[str, NetworkInfo]` or similar

algosdk/atomic_transaction_composer.py
```diff
@@ -165,67 +188,156 @@ def add_method_call(
...
-                "application create call not supported"
-            )
+
+        # Initialize foreign object maps
+        if not accounts:
+            accounts = []
+        if not foreign_apps:
+            foreign_apps = []
+        if not foreign_assets:
+            foreign_assets = []
```
jasonpaulos, line 232: It would be best to make a copy of these input arrays always, since this function may append to them.

algosdk/atomic_transaction_composer.py
algochoi: I will delete this check entirely because I think it is slightly better to have this check on a lower level (e.g. `algod`) as discussed

algosdk/abi/reference.py
```diff
@@ -0,0 +1,17 @@
...
+    # Asset reference type
+    ASSET = "asset"
+
+
+def is_abi_reference_type(t: "ABIReferenceType") -> bool:
+    return t in (
+        ABIReferenceType.ACCOUNT,
+        ABIReferenceType.APPLICATION,
+        ABIReferenceType.ASSET,
+    )
```
ahangsu, line 17: nit: I think my suggestion on this one is kinda optional: https://docs.python.org/3/library/enum.html
It feels to me that you can take advantage of `Enum` class, and see if an element is abi_reference type can use a simple `in` operation.
`is_abi_reference_type` can actually accept  any variable type, or according to the suggestion of type hint, the boolean function is vacuous true.

algosdk/abi/transaction.py
```diff
@@ -0,0 +1,45 @@
...
+    ACFG = constants.ASSETCONFIG_TXN
+
+    # Asset transfer transaction type
+    AXFER = constants.ASSETTRANSFER_TXN
+
+    # Asset freeze transaction type
+    AFRZ = constants.ASSETFREEZE_TXN
+
+    # Application transaction type
+    APPL = constants.APPCALL_TXN
```
ahangsu, line 25: nit: same as ref type, seems okay to use `Enum` (totally optional)

algosdk/abi/reference.py
algochoi: That's a good point, I think the type hint should be changed to `Any`. I think Python's regular `Enum` has some issues with string comparison so I opted for this comparison.

algosdk/atomic_transaction_composer.py
```diff
@@ -33,6 +34,38 @@ class AtomicTransactionComposerStatus(IntEnum):
     COMMITTED = 4
 
 
+def populate_foreign_array(
+    value_to_add: Any, foreign_array: List[Any], zero_value: Any = None
```
jasonpaulos, line 38: nit: you could use generics here: https://mypy.readthedocs.io/en/stable/generics.html#generic-functions

algosdk/abi/contract.py
```diff
@@ -36,13 +46,43 @@ def from_json(resp: Union[str, bytes, bytearray]) -> "Contract":
     def dictify(self) -> dict:
         d = {}
         d["name"] = self.name
-        d["appId"] = self.app_id
+        d["desc"] = self.desc
+        d["networks"] = self.networks
```
jasonpaulos, line 50: This should call `dictify` on every value in the `self.networks` dict

algosdk/atomic_transaction_composer.py
```diff
@@ -50,7 +82,10 @@ class AtomicTransactionComposer:
     # The maximum size of an atomic transaction group.
     MAX_GROUP_SIZE = 16
     # The maximum number of app-args that can be individually packed for ABIs
-    MAX_ABI_APP_ARG_LIMIT = 14
+    MAX_APP_ARG_LIMIT = 16
+    # The maximum number of foreign objects that can be associated per app
+    FOREIGN_ACCOUNT_LIMIT = 4
+    FOREIGN_ARRAY_LIMIT = 8
```
jasonpaulos, line 88: These are no longer needed?

algosdk/abi/contract.py
```diff
@@ -10,22 +10,32 @@ class Contract:
...
+        name: str,
+        methods: List[Method],
+        desc: str = None,
+        networks: Dict[str, "NetworkInfo"] = None,
+    ) -> None:
         self.name = name
-        self.app_id = int(app_id)
         self.methods = methods
+        self.desc = desc
+        self.networks = networks
```
jasonpaulos, line 29: Also, `networks` might be None here, perhaps do `self.networks = networks if networks else {}` instead?

jasonpaulos (approved):
Pending tests, looks good!

ahangsu (approved):

PR #253, opened 2021-11-24
Add type hints and clean up ABI code

algochoi:
This PR aims to do some refactoring and clean up so the ABI code has type hints, change some relative imports to absolute imports and makes some minor documentation changes to reflect the newer design.

I also opted to use [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) in some places because we are still on `3.5` for now.


algosdk/abi/address_type.py
```diff
@@ -62,7 +64,7 @@ def encode(self, value):
             )
         return bytes(value)
 
-    def decode(self, bytestring):
+    def decode(self, bytestring: Union[bytearray, bytes]) -> bytes:
```
jasonpaulos, line 67: I think this actually returns a string

algosdk/abi/address_type.py
```diff
@@ -1,7 +1,9 @@
-from .base_type import ABIType
```
tzaffi, line 1: Nice!!!

algosdk/abi/util.py
```diff
@@ -1,23 +1,24 @@
...
+from algosdk import error
 
 
 # Globals
 UFIXED_REGEX = r"^ufixed([1-9][\d]*)x([1-9][\d]*)$"
 STATIC_ARRAY_REGEX = r"^([a-z\d\[\](),]+)\[([1-9][\d]*)]$"
 
 
-def type_from_string(s):
+def type_from_string(s: str) -> ABIType:
```
jasonpaulos, line 21: Not directly related to this PR, but in the sprit of improving ABI support, I realized it is possible to implement this as a static method on the `ABIType` class, like `ABIType.of()` or `ABIType.from_string()` (not sure what the best name would be for Python).

You can do it putting all the subclass imports that would otherwise be recursive inside the method body. Something like:

```python
@staticmethod
def of(s: str) -> "ABIType":
   from algosdk.abi.uint_type import UintType
   from algosdk.abi.uint_type import UintType
   ...
```

This avoids the circular import issue.

jasonpaulos (commented):
This is a nice improvement, but please consider replacing instances of `list` with `List[<the element type>]` for further type support. I believe a plain `list` is equal to `List[Any]`.

You can import `List` from the `typing` module.

algosdk/abi/array_dynamic_type.py
```diff
@@ -1,45 +1,47 @@
...
         if not isinstance(other, ArrayDynamicType):
             return False
         return self.child_type == other.child_type
 
-    def __str__(self):
+    def __str__(self) -> str:
         return "{}[]".format(self.child_type)
 
-    def byte_len(self):
+    def byte_len(self) -> NoReturn:
```
tzaffi, line 32: 👍

algosdk/abi/array_static_type.py
```diff
@@ -1,62 +1,63 @@
...
+        child_type (ABIType)
         static_length (int)
     """
 
-    def __init__(self, arg_type, array_len) -> None:
+    def __init__(self, arg_type: ABIType, array_len: int) -> None:
         if array_len < 1:
             raise error.ABITypeError(
                 "static array length must be a positive integer: {}".format(
-                    len(array_len)
```
tzaffi, line 27: Nice catch!

algosdk/abi/contract.py
```diff
@@ -28,19 +29,19 @@ def __eq__(self, o) -> bool:
         )
 
     @staticmethod
-    def from_json(resp):
+    def from_json(resp: Union[str, bytes, bytearray]) -> "Contract":
```
tzaffi, line 32: I agree with this approach for our SDK, rather than self-referencing `Contract` (with a `from __future__ import annotations` above). That's because the [future may never arrive](https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class).

algosdk/abi/method.py
```diff
@@ -69,14 +72,14 @@ def get_selector(self):
...
-    def get_txn_calls(self):
+    def get_txn_calls(self) -> int:
         """
         Returns the number of transactions needed to invoke this ABI method.
         """
         return self.txn_calls
 
     @staticmethod
-    def _parse_string(s):
+    def _parse_string(s: str) -> list:
```
tzaffi, line 82:
```suggestion
    def _parse_string(s: str) -> List[str]:
```

Optional with a`from typing import List` above. I didn't suggest that in the `encod()` type hints, cause it was getting kind of verbose and I wasn't 100% what the `list` was made up of. But here it might make more sense

algosdk/abi/method.py
```diff
@@ -88,19 +91,19 @@ def _parse_string(s):
                     break
                 left_index = stack.pop()
                 if not stack:
-                    return (s[:left_index], s[left_index + 1 : i], s[i + 1 :])
+                    return [s[:left_index], s[left_index + 1 : i], s[i + 1 :]]
```
tzaffi, line 94: Good catch

algosdk/abi/__init__.py
```diff
@@ -1,14 +1,14 @@
...
-from .ufixed_type import UfixedType
-from .base_type import ABIType
-from .bool_type import BoolType
-from .byte_type import ByteType
-from .address_type import AddressType
-from .string_type import StringType
-from .array_dynamic_type import ArrayDynamicType
-from .array_static_type import ArrayStaticType
-from .tuple_type import TupleType
+from algosdk.abi.uint_type import UintType
```
tzaffi, line 2: 👍

algosdk/abi/tuple_type.py
```diff
@@ -128,7 +130,7 @@ def _compress_multiple_bool(value_list):
                 result |= 1 << (7 - i)
         return result
 
-    def encode(self, values):
+    def encode(self, values: list) -> bytes:
```
tzaffi, line 133: Is this `encode` method more restrictive than the others? I.e., it doesn't accept `bytes` or `bytearray` ?

algosdk/abi/ufixed_type.py
```diff
@@ -1,21 +1,23 @@
...
 
 
 class UfixedType(ABIType):
     """
     Represents an Ufixed ABI Type for encoding.
 
     Args:
-        bit_size (int, optional): size of a ufixed type.
-        precision (int, optional): number of precision for a ufixed type.
+        type_size (int): size of a ufixed type.
```
tzaffi, line 12: 👍

algosdk/abi/util.py
algochoi: Amazing 🙌  I like it

algosdk/abi/tuple_type.py
algochoi: For tuples, `encode()` can basically take a list of any `ABIType`, so if I had a tuple like `(bool,byte,uint8)`, it would take a list like: `[True, b"\x01", 24]` and encode that into `bytes` so we can send it off to an ABI compliant app. It could take `bytes` object inside the list if needed. Similarly, `decode()` should convert the `bytes` or `bytearray` back to the list of `ABITypes`.

algosdk/abi/util.py
tzaffi: +1 on `from_string(F)`

algosdk/abi/method.py
algochoi: Will do 👍

algosdk/abi/tuple_type.py
tzaffi: so maybe `values: List[ABIType]) ...`. ?

tzaffi (approved):
I am approving the PR as I don't have any beef with it being merged as is. None of my suggestions are mandatory.

algosdk/abi/tuple_type.py
algochoi: My bad, it should be list of any "ABIValues". We didn't define that yet so I'm incline to use a `List[Any]` for this PR.

algochoi: Actually good catch, I forgot that `TupleType` can accept `bytes` or `bytearray` if the child types are `ByteType` -- will fix this!

PR #256, opened 2021-11-25
FOR REVIEW ONLY - py-algorand-sdk v1.9.0

onetechnical:
# Changelog

### Added

- ABI Interaction Support for Python SDK (#247)
- ABI Type encoding support (#238)
- Add type hints and clean up ABI code (#253)
- Add CircleCI configs to the Python SDK repo (#246)


jasonpaulos:
Can we release this as a beta instead? Making the version `1.9.0b1` should work, according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-releases).

onetechnical:
> Can we release this as a beta instead? Making the version `1.9.0b1` should work, according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-releases).

Yes, opening #257 instead.

Issue #258, opened 2021-12-02
Dryrun from SDK


algoanne:
## What are you looking for? 

A way to perform a transaction/transaction group dryrun using the SDK. 

## Why is it useful? 

Dryrun simulates the submission of transactions to the network and outputs the result of the submission. Thus it can be used to test TEAL programs without having them on the network. Dryrun can currently be run through the CLI using goal (see https://developer.algorand.org/docs/clis/goal/clerk/dryrun/). 

It would be natural to be able to use dryrun from the SDKs for developers who build out transactions using the SDKs. This avoids needing to use goal and run a node to do this testing.  

## Some functional details

- inputs to the dryrun fuction: the transaction/transaction group file, an algod client
- output using JSON format
- use the current protocol version

## Notes 
the goal version of dryrun can take in protocol version to use and output format, we keep these out of scope for this initial implementation.

PR #259, opened 2021-12-03
Create dryrun

barnjamin:
Initial stab at providing utility function to create DryrunRequest object from transaction list given a client.

Probably needs tests but I need guidance on what tests and where they should go

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -3044,20 +3044,24 @@ def wait_for_confirmation(algod_client, txid, wait_rounds=0, **kwargs):
         current_round += 1
 
 
-
 def create_dryrun(
-    client: algod.AlgodClient, txns: List[Union[transaction.SignedTransaction,  transaction.LogicSigTransaction]],
-    protocol_version=None, latest_timestamp=None, round=None
-)->models.DryrunRequest:
+    client: algod.AlgodClient,
+    txns: List[
+        Union[transaction.SignedTransaction, transaction.LogicSigTransaction]
+    ],
+    protocol_version=None,
+    latest_timestamp=None,
+    round=None,
+) -> models.DryrunRequest:
     """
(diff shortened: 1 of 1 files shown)


barnjamin:
addresses https://github.com/algorand/py-algorand-sdk/issues/258

jasonpaulos:
This seems like a good addition. My one suggestion would be that maybe the `AtomicTransactionComposer` should have a method to generate a dryrun request as well?

barnjamin:
> My one suggestion would be that maybe the AtomicTransactionComposer should have a method to generate a dryrun request as well

Do you mean add a method in the AtomicTransactionComposer class that creates the dryrun (and submits it?). I was thinking users may just call build then pass the result to this method.

Does it make sense to also add a method to write to a file? Essentially https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/testing/dryrun.py#L820

jasonpaulos:
> Do you mean add a method in the AtomicTransactionComposer class that creates the dryrun (and submits it?). I was thinking users may just call build then pass the result to this method.

Yeah I was suggesting adding a method to the composer to streamline the process. I hadn't considered having the composer also submitting the dryrun request, but that seems like potentially a good idea too.
 
> Does it make sense to also add a method to write to a file? Essentially https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/testing/dryrun.py#L820

Personally I think the SDKs should _only_ provide a way to encode and decode objects to bytes for storage, not specific IO for files. The reason being it requires minimal additional code to read/write encoded bytes to files, and if we focus only on file IO we can't address any other storage mechanisms, like DBs or sending objects over the network.

barnjamin:
> Yeah I was suggesting adding a method to the composer to streamline the process.

I'd like to add this in a separate PR if that is ok? Maybe use this method to construct it internally

jasonpaulos (approved):

Issue #237, opened 2021-09-27
Algoexplorer API 1010 status code check


barnjamin:
## Problem

I've seen several folks in the discord report issues getting started when they try to use the python SDK.  Trying to call against the algoexplorer api returns a status code 1010 from the User Agent check but it presents itself as a json decoder error. 

I believe this may be the line triggering the error but I haven't dug into it. https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/algod.py#L89

Mostly posting this as a reminder to add a check

## Solution

Add a check for http status code and if 1010, print out an error message that suggests adding `User-Agent` header field

barnjamin:
Would it be better to set a header like `User-Agent: 'py-algorand-sdk'` by default?

algochoi:
Closed by #260

PR #260, opened 2021-12-03
Default header

barnjamin:
Adding a default User-Agent header to the v2 algod client to address #237


algochoi:
One more thing: Do we also want to do this for the `indexer`?

barnjamin:
Good point, I'll add there as well

algochoi:
:shipit:

PR #261, opened 2021-12-07
Change behave tag list to use commas

barnjamin:
s/ or /,//g


barnjamin:
`python -m pip install -U git+https://github.com/behave/behave`

behave                   1.2.7.dev2

PR #262, opened 2021-12-07
Add requirement to fetch behave source code and update readme

algochoi:
This PR makes a README change to add details to the development section and specify the `requirements.txt` to fetch the latest `behave` for tag support.

Thanks to @barnjamin for this suggestion 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,6 @@ Install dependencies
 
 Run tests
 * `make docker-test`
-* `make unit` (only unit tests)
-* `make integration` (only integration tests)
 
 Format code:
 * `black .`
--- a/README.md
+++ b/README.md
@@ -14,7 +14,16 @@ Alternatively, choose a [distribution file](https://pypi.org/project/py-algorand
 
 ## SDK Development
 
-Run tests with `make docker-test`
+Install dependencies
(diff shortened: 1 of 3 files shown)


barnjamin (approved):
Thanks!

README.md
```diff
@@ -14,7 +14,16 @@ Alternatively, choose a [distribution file](https://pypi.org/project/py-algorand
...
 ## SDK Development
 
-Run tests with `make docker-test`
+Install dependencies
+* `pip install -r requirements.txt`
+
+Run tests
+* `make docker-test`
+* `make unit` (only unit tests)
+* `make integration` (only integration tests)
```
jasonpaulos, line 23: This is a little misleading because these commands only work if the testing harness is running.

jasonpaulos (commented):
Since it's probably best to specify dependencies only in one place, could also you modify line 10 of `Dockerfile` to use `pip install -r requirements.txt -q` instead?

README.md
algochoi: Good point -- I think I'll just revert this back to `make docker-test`

jasonpaulos (approved):

PR #263, opened 2021-12-07
Fix wait for confirmation function

jasonpaulos:
Fixes the following issues with the wait for confirmation function:
1. `pool-error` was being ignored
2. If the call to `pending_transaction_info ` raises an HTTP error, that error should be ignored and the function should continue retrying if `wait_rounds ` allows it.
3. Removed the special value 0 for `wait_rounds ` which allows indefinitely waiting. This is a dangerous feature, since your code might actually wait forever if it never sees the transaction fail. Instead a value of 1000 (the maximum txn life) is used to implement the same thing, and in reality is the longest you will ever have to wait. (Because v1.8.0 of this SDK was already released with this function, we can't get rid of the default value altogether without breaking the API).
4. Remove f-string use, which is not available in Python 3.5

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -9,6 +9,7 @@ from .. import encoding
 from .. import error
 from .. import logic
 from .. import transaction
+from ..v2client import algod
 from nacl.signing import SigningKey, VerifyKey
 from nacl.exceptions import BadSignatureError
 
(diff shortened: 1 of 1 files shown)


algosdk/future/transaction.py
```diff
@@ -3010,31 +3011,51 @@ def assign_group_id(txns, address=None):
     return result
 
 
-def wait_for_confirmation(algod_client, txid, wait_rounds=0, **kwargs):
+def wait_for_confirmation(
+    algod_client: algod.AlgodClient, txid: str, wait_rounds: int = 0, **kwargs
```
winder, line 3015: Set the default value here instead of checking for 0 and overriding it.

A smaller default like 5 would probably be better default than 1000.

jasonpaulos: This is my attempt at maintaining backwards compatibility, since we already shipped a version with this function (unlike the other SDKs).

There might be existing code that passes 0 for `wait_rounds` and expects the function to wait indefinitely, so I though 1000 was the best replacement for it.

winder (approved):

PR #268, opened 2021-12-23
C2C Feature and Testing

tzaffi:
# Handling Contract to Contract Applications in the SDK

This PR aims to show that the critical capabilities for handling C2C functionality are in place. This is accomplished by implementing and passing Cucumber integration tests that are defined in `algorand-sdk-testing` [PR #156](https://github.com/algorand/algorand-sdk-testing/pull/156).

### Slight Enhancement to Function Obtaining Application Account Address
In a Contract-to-Contract call, the application that is making the call is responsible for paying the associated transaction fees. This means that the application needs to be funded and that its account address should be readily available for funding. `logic` already contains `get_application_address()`, and I'm just adding a type-checker for it.

### Add void ABI Method `appl` Transactions into the `List[ABIResult]` Member of `AtomicTransactionResponse`
This is mainly for completeness when trying to understand all trasactions (including nested) and all side-effects that resulted from ABI method calls in an atomic transactions group.

## New Cucumber Tests
To pass the tests in [testing's PR #156](https://github.com/algorand/algorand-sdk-testing/pull/156) (and mainly `integration/c2s.feature`) functions have been refactored with new steps added inside of `test/steps/`

### Transaction call trace from an Atomic Transaction Composer ABI Execution

Allow [json-like traces](https://gist.github.com/tzaffi/6166662066920ca321b598bc56389c05) of ABI calls via `transactions_trace(atc: AtomicTransactionComposer, results: List[ABIResult])` in `v2_steps.py`

## Summary of changes
* `Makefile`: re-org the cucumber tags into variables for easier running of selected tests
* `algosdk/atomic_transaction_composer.py`: adding non-void ABI method `appl` transactions to the `ABIResult`s inside of an `AtomicTransactionResponse`
* `algosdk/future/template.py`: deprecation comments added
* `algosdk/logic.py`: type assertion for `get_application_address(appID: int)`
* `algosdk/v2client/algod.py`: re-org'ed the imports ONLY
* `requirements.txt`: added `glom` and `pyteset` which are needed for testing (but not any running code)
* `run_integration.sh`: temporarily pointing to C2C sdk-testing branch
* `test/environment.py`: small bugfix
* new file `test/steps/__init__.py`: helped me run my tests locally in a debugger
* `test/steps/steps.py`: remove `Templates` related steps
* `test/steps/v2_steps.py`: modifications for passing the cucumber tests
* `test_unit.py`: 
  * introduced a new test case in a pre-existing test
  * removed `Templates` related unit tests

## Links
[...]

PR #278, opened 2022-01-20
Enhancement: ATC dryrun

barnjamin:
Adding `dryrun` here to see how it feels, added cost to abi result as well but maybe should move that elsewhere?

I copied most of the body from `execute` method

If this looks reasonable I can add to other sdks

api is:
```py
        signer = AccountTransactionSigner(pk)

        sp = client.suggested_params()
        # Create atc to handle method calling for us
        atc = AtomicTransactionComposer()
        # add a method call to "call" method, pass the second app id so we can dispatch a call
        atc.add_method_call(
            app_id, get_method(c, "acct_param"), addr, sp, signer, method_args=[addr]
        )

        result = atc.dryrun(client)

        for i, res in enumerate(result.abi_results):
            print(
                "Result of '{}': {} (cost {})".format(
                    atc.method_dict[i].name, res.return_value, res.cost
                )
            )
```

Contains:
```py
dryrun_response: dict,
tx_ids: List[str],
results: DryrunABIResult,
traces: List[DryrunResponse],
```


algochoi:
We have this [PR open for C2C testing](https://github.com/algorand/py-algorand-sdk/pull/268/files) that includes the `PnedingTransactionInfo` into the `ABIResult`, which might make checking for cost either. Might be relevant: https://github.com/algorand/py-algorand-sdk/blob/e1aa79c23dd5f97b67181cb88f4053d2efe12ec4/algosdk/atomic_transaction_composer.py#L694

barnjamin:
@algochoi I think only the DryrunTxnResult contains the cost field

https://github.com/algorand/go-algorand/blob/8a335bb2b10ea3cb6e9471526701bcaa4089b53d/daemon/algod/api/server/v2/generated/types.go#L303

What do you think about abstracting the log parsing into a self contained method that gets re-used?

algochoi:
> @algochoi I think only the DryrunTxnResult contains the cost field
> 
> https://github.com/algorand/go-algorand/blob/8a335bb2b10ea3cb6e9471526701bcaa4089b53d/daemon/algod/api/server/v2/generated/types.go#L303
> 
> What do you think about abstracting the log parsing into a self contained method that gets re-used?

Ahh okay, that makes sense. So the `ABIResult` object will only have a `cost` field for the dryrun object (or maybe we can create/inherit a separate dryrun response object)? I think the log parsing method looks good to me, but others can chime in on this.

jasonpaulos (commented):
I like the idea of this, but it might be clearer to have a different response type for the `dryrun` method instead of reusing `AtomicTransactionResponse`.

barnjamin:
@jasonpaulos thats fair, can I keep the log parsing as is?

jasonpaulos:
> @jasonpaulos thats fair, can I keep the log parsing as is?

To clarify, I think the dryrun response should not use `AtomicTransactionResponse` or `ABIResult`. It should likely have its own response type(s) that contain only the fields it will populate.

Though I agree that the same log parsing code should be used to find the return values for both dryrun and normal execute. It just probably should not return `ABIResult`s, instead maybe a tuple of the raw information that can be used to create an `ABIResult` or the dryrun equivalent.

algosdk/atomic_transaction_composer.py
```diff
@@ -703,3 +747,31 @@ def __init__(
         self.confirmed_round = confirmed_round
         self.tx_ids = tx_ids
         self.abi_results = results
+
+
+class DryrunABIResult:
```
algochoi, line 752: How about we inherit this from the `ABIResult` class? This constructor can call `super()` and set the cost.

algochoi (commented):
Looks largely good to me, thanks for doing this. I would also suggest merging the `tzaffi/c2c` branch to here because there may be some conflicting changes (e.g. we added a `tx_info` field to the `ABIResult` class)

barnjamin:
>I would also suggest merging the tzaffi/c2c branch to here because there may be some conflicting changes

Merged, probably should wait to review this again until that one is merged to master though?

jasonpaulos:
@barnjamin if can merge develop back into this branch, I can give it a final review

barnjamin:
@jasonpaulos does it make sense to wait for the dryrun stack printer to be merged in first? then the result of calling atc.dryrun also includes the DryrunRequest object?

barnjamin:
@algochoi 
> Are we going to add this to the other SDKs and add tests like https://github.com/algorand/algorand-sdk-testing/pull/161/files

I expect it will need to be added to other SDKs as well but I think before adding tests we need to find a nice way to test the dryrun request creation bits using some live-ish ledger.

Alternatively we could just create some DryrunResponse json objects and make sure it parses out the abi results correctly?

barnjamin:
@jasonpaulos @ahangsu can I get another peek at this please? 

Other SDKs will replicate the behavior, most of the value here is in decoding the ABI results

PR #283, opened 2022-01-31
Dryrun response

barnjamin:
API should look something like:

```py
    group = atc.gather_signatures()
    drr = create_dryrun(client, group)
    dr = DryrunResponse(client.dryrun(drr))
    for txn in dr.txns:
        print(txn.app_trace())
```

and output
```
#pragma version 6       []
intcblock 1 0 10        []
bytecblock 0xd87699ef   []
txn ApplicationID       [57]
intc_1 // 0             [57, 0]
==                      [0]
bnz label1              []
txn OnCompletion        [0]
pushint 5               [0, 5]
==                      [0]
bnz label2              []
txn OnCompletion        [0]
pushint 4               [0, 4]
...
```


algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
+from typing import List
+
+
+class DryrunResponse:
+    def __init__(self, drrjson: dict):
+        # These are all required fields
```
tzaffi, line 6: If these are required, why not just add them to the init?

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
...
+        self.error = drrjson["error"]
+        self.protocol = drrjson["protocol-version"]
+        self.txns = [DryrunTransactionResult(txn) for txn in drrjson["txns"]]
+
+
+class DryrunTransactionResult:
+    def __init__(self, dr):
+        # Only required field
+        self.disassembly = dr["disassembly"]
```
tzaffi, line 15: I also recommend an assertion similar to the above suggestion here.

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
...
+        self.pc = tl["pc"]
+        self.stack = [DryrunStackValue(sv) for sv in tl["stack"]]
+
+    def trace_line(self):
+        return (self.line, [sv.__str__() for sv in self.stack])
+
+
+class DryrunStackValue:
+    def __init__(self, v):
+        self.type = v["type"]
```
tzaffi, line 78: similar assertions here

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
...
+        self.protocol = drrjson["protocol-version"]
+        self.txns = [DryrunTransactionResult(txn) for txn in drrjson["txns"]]
+
+
+class DryrunTransactionResult:
+    def __init__(self, dr):
+        # Only required field
+        self.disassembly = dr["disassembly"]
+
+        # These are all optional
```
tzaffi, line 17: Nit. with so many keys, it's probably easier to read and maintain if you do something like:
```python
fields =["disassembly", "app-call-messages", .... ]
for field in fields:
    if field in dr:
        setattr(self, field, dr[field]) 
```
then later, you can re-set the special fields `app-call-trace` and `logic-sig-trace`

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
...
+        if "cost" in dr:
+            self.app_call_cost = dr["cost"]
+        if "logic-sig-messages" in dr:
+            self.logic_sig_messages = dr["logic-sig-messages"]
+        if "logic-sig-trace" in dr:
+            self.logic_sig_trace = DryrunTrace(dr["logic-sig-trace"])
+        if "logs" in dr:
+            self.logs = dr["logs"]
+
+    def app_trace(self, trace_spaces: int = 16):
```
tzaffi, line 35: Is there a reason for the default of 16? In any case, I recommend making this "magic number" a class constant.
Additionally, in the spirit of D.R.Y. (don't repeat yourself) I recommend adding a new class method which does the string formatting and is called by each of the `*_trace()` instance methods:
Something like:
```python
class DryRunTransactionResult:
    DEFAULT_TRACE_SPACES = 16
... etc ...
    @classmethod
    trace(cls, dr_trace: DryRunTrace, spaces: int = None):
        if spaces is None:
            spaces = cls.DEFAULT_TRACE_SPACES
        for line in dr_trace.get_trace():
            ... etc ...

 
        def app_trace(self, spaces: int = None):
            return self.trace(self.app_call_trace, spaces=spaces)

        .... similarly for lsig_trace() ...

File: algosdk/dryrun_results.py [hunk 6]
```diff
@@ -0,0 +1,85 @@
...
+
+
+class DryrunTraceLine:
+    def __init__(self, tl):
+        self.line = tl["line"]
+        self.pc = tl["pc"]
+        self.stack = [DryrunStackValue(sv) for sv in tl["stack"]]
+
+    def trace_line(self):
+        return (self.line, [sv.__str__() for sv in self.stack])
```
tzaffi, line 73: Best practice is to use the builtin version of a dunder... i.e. prefer `str(sv)` over `sv.__str__()`

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,85 @@
...
+class DryrunStackValue:
+    def __init__(self, v):
+        self.type = v["type"]
+        self.bytes = v["bytes"]
+        self.int = v["uint"]
+
+    def __str__(self):
+        if self.type == 1:
+            return self.bytes
+        return self.int
```
tzaffi, line 85: Nit.

We have a strictly linear class composition hierarchy here:
```
DryrunTrace has a DryrunTraceLine has a DryrunStackValue
```
To me it suggests that we can probably do without the last two classes. So you could do something like:

```python
class DryRunTrace:
    def __init__(self, trace: List[dict]):  # is that really the correct type-hint?
        self._trace = trace

    def get_trace(self) -> List[Tuple....]:   # I'm not sure about the type-hint
        return [self.trace_line(line) for line in self._trace]

    @classmethod
    def trace_line(cls, line: dict) -> Tuple[str, list]:
        return (line["line"], [cls.stack_value(sv) for sv in line["stack"]])

    @classmethod
    def stack_value(cls, sv: dict) -> str:
        .... etc ...
```

After doing all this refactoring, I'm not convinced it's actually worth it. A lot of effort, for just a little gained.

tzaffi (commented):
Looks very useful.

I have a couple of suggestions to shorten the code, and some nit-picks that aren't as crucial.

A general question that we should bring up at standup as well: How do we feel about SDK's getting out of sync? I don't believe that this sort of helper functionality exists in any of the other SDK's, and is that ok? I think it should be ok. We may need to signal that this is special to python by having it live in a directory such as `extras` or `goodies` ...

barnjamin:
@tzaffi thanks for the review!

> I don't believe that this sort of helper functionality exists in any of the other SDK's, and is that ok?

I expect to add it to the rest of them. I like using python as a testing ground for these to see if it makes sense and find a good pattern.

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,112 @@
...
+        for field in optionals:
+            if field in dr:
+                setattr(self, field.replace("-", "_"), dr[field])
+
+        traces = ["app-call-trace", "logic-sig-trace"]
+        for trace_field in traces:
+            if trace_field in dr:
+                setattr(
+                    self,
+                    trace_field.replace("-", "_"),
```
tzaffi, line 45: ahh, ya had to replace the hyphens!

tzaffi (approved):
LGTM

jasonpaulos (commented):
Looks like a nice improvement, though it would be great if there were at least one unit test.

The code seems very testable, since it does not rely on any REST API calls

ahangsu:
looks pretty neat to me, waiting on the test cases

ahangsu (commented):
looks pretty neat to me (not sure if we want to ensure fields in `DryrunTraceLine` or `DryrunStackValue`), waiting on the test cases

barnjamin:
Before merging, I want to raise a few more things that may be useful to include.

1. Some indication in the trace for _where_ the issue originated. Often when looking at a trace I try to find the first line the 0 that fails the program is put on the stack. If we can do that automatically and add a `<-- This line G`, it might help.
2. Parsing out the deltas for each Transaction result. At the moment we don't have anything that does this automatically to b64 decode keys/values. 
3. A trivial `succeeded` type method on the Transaction result that can inspect the `app-call-messages` or `logic-sig-messages` for the relevant `APPROVED` or `REJECTED` 
4. At the moment the `disassembly` is ambiguous and if an app call is signed by logic, the app call disassembly will be returned https://github.com/algorand/go-algorand/issues/3563

barnjamin:
I changed the formatting a bit, to add line number, unquote stack values/convert to hex for bytes, and if 0 is passed for spaces use the longest line length.

Curious to hear thoughts on this format instead

```
30  | store 0                                                                                             [99]
32  | load 0                                                                                              []
33  | pushint 97                                                                                          [99]
34  | >                                                                                                   [99, 97]
35  | bnz label7                                                                                          [1]
39  | load 0                                                                                              []
40  | callsub label8                                                                                      [99]
109 | store 1                                                                                             [99]
110 | load 1                                                                                              []
111 | callsub label14                                                                                     [99]
75  | store 2                                                                                             [99]
76  | load 2                                                                                              []
77  | intc_1 // 0                                                                                         [99]
78  | ==                                                                                                  [99, 0]
79  | bnz label10                                                                                         [0]
80  | load 2                                                                                              []
81  | intc_2 // 10                                                                                        [99]
82  | /                                                                                                   [99, 10]
83  | intc_1 // 0                                                                                         [9]
84  | >                                                                                                   [9, 0]
85  | bnz label11                                                                                         [1]```

**Comment by barnjamin** [comment]
Tweaked this again to also include PC 
```
pc# line# source                                                                                                    stack
   1    1 intcblock 1 0 10                                                                                          []
   6    2 bytecblock 0x20426f74746c6573206f662062656572206f6e207468652077616c6c                                     []
  37    3 txn ApplicationID                                                                                         []
  39    4 intc_1 // 0                                                                                               [1]
  40    5 ==                                                                                                        [1, 0]
  41    6 bnz label1                                                                                                [0]
  44    7 txn OnCompletion                                                                                          []
  46    8 pushint 5                                                                                                 [0]
  48    9 ==                                                                                                        [0, 5]
  49   10 bnz label2                                                                                                [0]
  52   11 txn OnCompletion                                                                                          []
  54   12 pushint 4                                                                                                 [0]
  56   13 ==                                                                                                        [0, 4]
  57   14 bnz label3                                                                                                [0]
  60   15 txn OnCompletion                                                                                          []
  62   16 pushint 2                                                                                                 [0]
  64   17 ==                                                                                                        [0, 2]
 ```

jasonpaulos:
I like the new format! I see you also have a cucumber PR as well? https://github.com/algorand/algorand-sdk-testing/pull/161

Are these ready for final review or are they out of sync? I don't think CI is running the cucumber test yet. To test with the cucumber PR, you could temporarily change this line to point to your branch in the algorand-sdk-testing repo: https://github.com/algorand/py-algorand-sdk/blob/0bdd947ada1610f9a2820eae8378df508ee748d7/run_integration.sh#L10

algosdk/__init__.py
```diff
@@ -15,5 +15,6 @@
 from . import v2client
 from . import wallet
 from . import wordlist
+from . import dryrun_results
```
tzaffi, line 18: nit: can we keep this in alphabetical order?

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # 16 for length of the header up to spaces
+        lines = ["pc# line# source" + " " * (spaces - 16) + "stack"]
+
+        for line, pc, stack in dr_trace.get_trace():
+            # Pad to 4 spaces since we don't expect programs to have > 9999 lines
+            line_number_padding = " " * (4 - len(str(line)))
+            pc_number_padding = " " * (4 - len(str(pc)))
+
+            src_line = "{}{} {}{} {}".format(
```
tzaffi, line 89: I recommend you use the `tabulate` package (or something sophisticated like it) to handle the formatting. This will make sure to properly align all the columns and make the results more readable (especially the stack column). That's what I was using in my [PyTeal source map POC](https://github.com/algorand/pyteal/blob/b598fa949586f51f62e99f46224fdfeb44a97c26/pyteal/compiler/source_map.py#L24)

tzaffi: Looking at the tabulate link on PyPi above, it seems like adding `tablefmt="presto"` might be worthwhile as well. That way, an automated process such as a unit-test, might have an easier time parsing the human-readable dry-run results and validating certain cells.

tzaffi: Also, not sure if the logs (or at least the final log) deserve to make into the trace, given their importance in ABI apps. But there's a counter-argument that they should just be left where they are, and that the "ATC" version of the dry-run response will know how to handle them appropriately (I see you have another PR up for that)

tzaffi: Sorry to keep piling on while you're on vacay... but better to jot this down while it's fresh on my mind:

1. Nice to have: a column for scratchvar's. E.g., consider this [recent dryrun](https://gist.github.com/tzaffi/16f72b39faea1c3b45492caf9a7a8dfa#file-dr-json-L50)
2. I'm not sure where we are regarding running **op costs**. (Looping in @algochoi). But if that's coming soon into dry-run, it would be nice to have columns for opcode cost, and running cost. If that's not coming to dry-run, I still think it's worthwhile having this information, but there are downside, as we would need to maintain a dictionary of opcode costs and keep track of these by consensus version.

tzaffi: I created this issue in go-algorand regarding more detailed cost reporting in go-algorand's dry-run: https://github.com/algorand/go-algorand/issues/3714

Makefile
```diff
@@ -1,6 +1,6 @@
-UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring"
+UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring or @unit.dryrun.trace.application"
 unit:
-	behave --tags=$(UNITS) test -f progress2
+	behave --tags=$(UNITS) test -f progress2 -s -v
```
tzaffi, line 3: Do we want this verbosity? we probably do but worth discussing

barnjamin: nope, that was a mistake to include, thanks

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,228 @@
+from typing import List
+
+import tabulate
+
+tabulate.MIN_PADDING = 0
+
+from tabulate import tabulate, TableFormat, DataRow
+import base64
```
tzaffi, line 8:
```suggestion
import base64
from typing import List

import tabulate
from tabulate import tabulate, TableFormat, DataRow


tabulate.MIN_PADDING = 0
```

barnjamin: I dont think i can do it like that, It has to be import tabulate then change the global. The thing i import from tabulate is also called tabulate

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+            "logic-sig-messages",
+            "logic-sig-disassembly",
+            "logs",
+        ]
+        for field in optionals:
+            if field in dr:
+                setattr(self, field.replace("-", "_"), dr[field])
+            else:
+                setattr(self, field.replace("-", "_"), None)
+
```
tzaffi, line 53:
```suggestion
        def attrname(field):
            return field.replace("-","_")
            
        for field in optionals:
            setattr(self, attrname(field), dr.get(field))

```

tzaffi: minor improvement... and you can use the `attrname()` lambda elsewhere here

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+        traces = ["app-call-trace", "logic-sig-trace"]
+        for trace_field in traces:
+            if trace_field in dr:
+                setattr(
+                    self,
+                    trace_field.replace("-", "_"),
```
tzaffi, line 59:
```suggestion
                    attrname(trace_field)
```

assuming you added `attrname()`

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+                    DryrunTrace(dr[trace_field]),
+                )
+
+    def app_call_rejected(self) -> bool:
+        if self.app_call_messages is not None:
+            return "REJECT" in self.app_call_messages
+        return False
```
tzaffi, line 66:
```suggestion
        return False if self.app_call_messages is None else "REJECT" in self.app_call_messages
```

Not clearly better... but putting it out there

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def __eq__(self, other: "DryrunStackValue"):
+        return (
+            self.type == other.type
+            and self.bytes == other.bytes
+            and self.int == other.int
+        )
+
+
+def truncate(s: str, max_width: int) -> str:
```
tzaffi, line 197: looks like the column width is really `3 + max_width`. Is that the intention? Consistency across the API's trumps any other issues here.

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+    if len(s) > max_width and max_width > 0:
```
tzaffi, line 198: There is a weird edge case here when `max_width < len(s) < max_with + 3`. You get less data and more space usage.

So maybe it would be better to have the condition be:
```python
if len(s) >= max_width + 3 and max_width > 0:
```

But that could break the test, unfortunately.

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+                    truncate(stack, spc.max_width),
+                ]
+            )
+
+        return (
+            tabulate(
+                lines,
+                headers,
+                disable_numparse=True,
+                tablefmt=TableFormat(
```
tzaffi, line 113: NICE!!!!

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+            new_idx = idx
+            continue
+
+        if curr_scratch[idx] != prev_scratch[idx]:
+            new_idx = idx
+
+    if new_idx == None:
+        return ""
+
+    return "{} = {}".format(new_idx, str(curr_scratch[new_idx]))
```
tzaffi, line 221:
```suggestion
    return "{} = {}".format(new_idx, curr_scratch[new_idx])
```
I believe `format()` will apply `str()` for you, but you should verify

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+        return self.trace(self.app_call_trace, self.disassembly, spc=spc)
+
+    def lsig_trace(self, spc: StackPrinterConfig = None) -> str:
+        if not hasattr(self, "logic_sig_trace"):
+            return ""
+
+        if (
+            not hasattr(self, "logic_sig_disassembly")
+            or self.logic_sig_disassembly is None
+        ):
```
tzaffi, line 143:
```suggestion
        if getattr(self, "logic_sig_disassembly", None) is None:
```

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+            return ""
+
+        if spc == None:
```
tzaffi, line 146:
```suggestion
        if spc is None:
```

algosdk/dryrun_results.py
```diff
@@ -0,0 +1,227 @@
...
+    if len(s) > max_width and max_width > 0:
+        return s[:max_width] + "..."
+    return s
+
+
+def scratch_to_string(
+    prev_scratch: List[DryrunStackValue], curr_scratch: List[DryrunStackValue]
+) -> str:
+    if len(curr_scratch) == 0:
+        return ""
```
tzaffi, line 207:
```suggestion
    if not curr_scratch:
        return ""
```
slightlhy more pythoni

test/steps/v2_steps.py
```diff
@@ -3082,6 +3083,51 @@ def deserialize_json_to_contract(context):
...
 
+@given(
+    'a dryrun response file "{dryrun_response_file}" and a transaction at index "{txn_id}"'
+)
+def parse_dryrun_response_object(context, dryrun_response_file, txn_id):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    dir_path = os.path.dirname(os.path.dirname(dir_path))
+    with open(
+        dir_path + "/test/features/resources/" + dryrun_response_file, "r"
+    ) as f:
```
tzaffi, line 3094:
```suggestion
    file = Path.cwd() /  "test" / "features" / "resources" /  dryrun_response_file
    with open(file) as f:
```
(assuming a `from pathlib import Path` at the top)
... but I'm a `pathlib` fanboy

tzaffi: there's also the `load_resource()` function which is probably better than my previous suggetion.

I.e.:

```python
drr_dict = json.loads(load_resource(dryrun_response_file, is_binary=False))
```

test/steps/v2_steps.py
```diff
@@ -3082,6 +3083,51 @@ def deserialize_json_to_contract(context):
...
+
+
+@then('calling app trace produces "{app_trace_file}"')
+def dryrun_compare_golden(context, app_trace_file):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    dir_path = os.path.dirname(os.path.dirname(dir_path))
+    with open(
+        dir_path + "/test/features/resources/" + app_trace_file, "r"
+    ) as f:
+        trace_expected = f.read()
```
tzaffi, line 3110:
```suggestion
    trace_expected = load_resource(app_trace_file, is_binary=False)
```

tzaffi (commented):
Several minor suggestions, the most controversial being around how to handle `max_width`.

tzaffi:
Question for @jasonpaulos . This PR (and #298 as well) adds a new package dependency `tabulate`. Should this be added to our `setup.py` here:

https://github.com/algorand/py-algorand-sdk/blob/7c0b8a4bccfbf6a512c690b6f07707f9cc2d9b71/setup.py#L19

Makefile
```diff
@@ -1,4 +1,4 @@
-UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring"
+UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring or @unit.dryrun.trace.application"
```
tzaffi, line 1:
```suggestion
UNITS = "@unit.abijson or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets"
```

suggest alphabetizing

algosdk/dryrun_results.py
barnjamin: All of the SDKs have the same type of logic where we truncate at the max width, then add the 3 period ellipses. I went back and forth on this but I think having max width describe where we truncate the real value then adding ellipses to denote that the value is truncated is the right way to go. It is tough though so if you feel strongly I can change it.

algosdk/dryrun_results.py
barnjamin: Maybe I should change the flag from max_width to max_value_width to prevent confusion on this?

tzaffi: Seems like a good compromise

jasonpaulos:
> Question for @jasonpaulos . This PR (and #298 as well) adds a new package dependency `tabulate`. Should this be added to our `setup.py` here?
> 
> https://github.com/algorand/py-algorand-sdk/blob/7c0b8a4bccfbf6a512c690b6f07707f9cc2d9b71/setup.py#L19

Yes, @barnjamin did this correctly in 8ded7ec1b3cef82dca0989787c2a91db147c8734

tzaffi (approved):
LGTM again

PR #285, opened 2022-02-08
Add CircleCI build step to generate docsets

aldur:
This PR adds a CircleCI build step to generate docsets for [Kapeli's Dash](https://kapeli.com/dash) and [zeal](https://zealdocs.org). The result is uploaded as a CircleCI artifact.

I was doing these steps manually as I tend to access the docs quite often, so thought to automate it. If not useful/interesting we can just close this. :)

Looks like this:
Screenshot 2022-02-08 at 11 57 01


cusma:
@aldur I think we should consider adding this to `pyteal` too, if/when CircleCI is added to that repo.

PR #290, opened 2022-02-14
Partially fix types for atomic transaction composer

aldur:
`decode_error` is now an optional Exception.

Was getting the following error from `mypy`:

> py-algorand-sdk/algosdk/atomic_transaction_composer.py|693 col 23-28 error| Module cannot be used as a type

Also, this was creating trouble when generating documentation (see https://github.com/algorand/py-algorand-sdk/pull/289)

Note that there are other warnings/issues, but decided to start small and only fix this one as it was blocking for docs.

### DIFF
--- a/algosdk/atomic_transaction_composer.py
+++ b/algosdk/atomic_transaction_composer.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 import base64
 import copy
 from enum import IntEnum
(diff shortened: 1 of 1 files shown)


algochoi (approved):
LGTM, thanks for this fix!

barnjamin (approved):

PR #295, opened 2022-02-17
Add account asset/app info

nicholasguoalgorand:
Add the following endpoints to the python sdk: `account_asset_info` and `account_application_info`

### DIFF
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -123,6 +123,28 @@ class AlgodClient:
         req = "/applications/" + str(application_id)
         return self.algod_request("GET", req, **kwargs)
 
+    def account_asset_info(self, address, asset_id, **kwargs):
+        """
+        Return asset information for a specific account.
+
+        Args:
+            address (str): account public key
+            asset_id (int): The ID of the asset to look up.
+        """
+        req = "/accounts/" + address + "/assets/" + str(asset_id)
+        return self.algod_request("GET", req, **kwargs)
+
+    def account_application_info(self, address, application_id, **kwargs):
+        """
+        Return application information for a specific account.
+
+        Args:
+            address (str): account public key
+            application_id (int): The ID of the application to look up.
+        """
+        req = "/accounts/" + address + "/applications/" + str(application_id)
+        return self.algod_request("GET", req, **kwargs)
+
     def pending_transactions_by_address(
         self, address, limit=0, response_format="json", **kwargs
     ):


jasonpaulos:
@nicholasguoalgorand are you (or anyone else) planning to add cucumber tests as well for these new endpoints? I would recommend it

nicholasguoalgorand:
yes, there is a ticket for that next.

algosdk/v2client/indexer.py
```diff
@@ -250,6 +250,176 @@ def account_info(
...
+        if limit:
+            query["limit"] = limit
+        if next_page:
+            query["next"] = next_page
+        if asset_id:
+            query["asset-id"] = asset_id
+
+        return self.indexer_request("GET", req, query, **kwargs)
+
+    def account_asset_by_creator(
```
winder, line 296: how did you come up with this name? The spec calls this function `lookupAccountAssets`, so I would expect the name to somehow be derived from that.

algosdk/v2client/indexer.py
```diff
@@ -250,6 +250,176 @@ def account_info(
...
+        if limit:
+            query["limit"] = limit
+        if next_page:
+            query["next"] = next_page
+        if asset_id:
+            query["asset-id"] = asset_id
+
+        return self.indexer_request("GET", req, query, **kwargs)
+
+    def account_application_local_state(
```
winder, line 339: Same with this one, the spec name is `lookupAccountAppLocalStates`

winder (commented):
I just have a question about the function names. They don't seem to be derived from the spec.

algosdk/v2client/indexer.py
```diff
@@ -250,6 +250,176 @@ def account_info(
...
+    def account_asset_info(
+        self,
+        address,
+        limit=None,
+        next_page=None,
+        asset_id=None,
+        block=None,
+        round_num=None,
+        include_all=False,
+        **kwargs
```
winder, line 262: the exclude parameter is missing, I believe it is needed for several of these.

winder (changes requested):
Actually, there are a few other things that seem to be missing:

`round` parameter removed from the `v2/assets/{id}/balances` function.
`exclude` parameter (it is an array) seems to be missing also.

winder:
The new tests need to be enabled: `@unit.responses.unlimited_assets,@unit.indexer.ledger_refactoring,@unit.algod.ledger_refactoring`

winder (approved):

PR #298, opened 2022-03-04
Teal blackbox

tzaffi:
# TEAL Blackbox Toolkit: Program Reporting and Testing via Dry Runs

### Testing

```sh
pytest x/blackbox
```

## TODO's (not in this PR)

* handle statefulness and non-determinism

## FORMERLY TODO's

* [x]  Figure out how to integrate this (if at all) with the existing [DryRunContext](https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/testing/dryrun.py#L45). See usage examples [here](https://github.com/algorand/docs/blob/bbd379df193399f82686e9f6d5c2bcb9d676d2d7/docs/features/asc1/teal_test.md#basic-setup-and-simple-tests)
  * **UPSHOT**: ditches DryRunContext. It's too heavy.
* [x]  Allow CSV outputs?
  * **UPSHOT**: yes!
* [x]  Don't require logic sigs?
  * **UPSHOT**: yes, not needed. However, the ability to _test_ logic sigs is crucial, and so logic-sigs are considered as first-class citizens.
* [x]  See if we can simplify the roundtrips using the approach taken in PytealUtils:
  * [assert_output](https://github.com/algorand/pyteal-utils/blob/main/pytealutils/strings/test_string.py#L17)
  * [execute_app](https://github.com/algorand/pyteal-utils/blob/550b998e33985896607512992aef0e9c53680ea9/tests/helpers.py#L312)
  * [get_stats_from_dryrun](https://github.com/algorand/pyteal-utils/blob/550b998e33985896607512992aef0e9c53680ea9/tests/helpers.py#L360)
  * **UPSHOT**: no round trips at all


algosdk/dryrun_results.py
```diff
@@ -4,6 +4,7 @@
 
 class DryrunResponse:
     def __init__(self, drrjson: dict):
+        self.raw_response: dict = drrjson
```
tzaffi, line 7:
```suggestion
```

tzaffi:
# TODO's
1. Figure out how to integrate this (if at all) with the existing [DryRunContext](https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/testing/dryrun.py#L45). See usage examples [here](https://github.com/algorand/docs/blob/bbd379df193399f82686e9f6d5c2bcb9d676d2d7/docs/features/asc1/teal_test.md#basic-setup-and-simple-tests)
2. Allow CSV outputs?
3. Don't require logic sigs?
4. See if we can simplify the roundtrips using the approach taken in PytealUtils:
  * [assert_output](https://github.com/algorand/pyteal-utils/blob/main/pytealutils/strings/test_string.py#L17)
  * [execute_app](https://github.com/algorand/pyteal-utils/blob/550b998e33985896607512992aef0e9c53680ea9/tests/helpers.py#L312)
  * [get_stats_from_dryrun](https://github.com/algorand/pyteal-utils/blob/550b998e33985896607512992aef0e9c53680ea9/tests/helpers.py#L360)

Makefile
```diff
@@ -1,10 +1,18 @@
...
 	behave --tags=$(UNITS) test -f progress2
 
-INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send"
-integration:
+# INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send"
+INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @kmd or @rekey or @send.keyregtxn or @send"
+integration: dryrun
 	behave --tags=$(INTEGRATIONS) test -f progress2
 
+dryrun:
```
tzaffi, line 11: If [this is TMI ](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/297/workflows/b54effa4-aef5-43b3-baea-e082cc4b9fb3/jobs/462/parallel-runs/0/steps/0-102) on circle, I can remove the `-sv` flag

.gitignore
```diff
@@ -122,3 +122,6 @@ py-algorand-sdk-*
 
 # Pycharm
 .idea
+
+# Comma Seperated Value report files
```
tzaffi, line 126:
```suggestion
# Comma Separated Value report files
```

x/blackbox/dryrun_mixin_docs_test.py
```diff
@@ -0,0 +1,612 @@
+"""
```
tzaffi, line 1: New test, based on old test in docs. Helps ensure that refactoring won't break origin `DryrunTestCaseMixin`

Makefile
tzaffi: I ended up removing the flag

Makefile
```diff
@@ -1,10 +1,15 @@
-UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment"
```
tzaffi, line 1: Temporarily commented out. Indexer tests were failing on me (should be completely independent to my changes), but this issue needs to be resolved before merging.

algosdk/testing/dryrun.py
```diff
@@ -42,6 +43,188 @@ class App:
     global_state: List[TealKeyValue] = None
 
 
+#### LIGHTWEIGHT ASSERTIONS FOR RE-USE ####
```
tzaffi, line 46: Refactor `DryrunTestCaseMixin` so can have common assertions with `teal_blackbox.py`

algosdk/testing/dryrun.py
```diff
@@ -485,23 +522,33 @@ def _checked_request(
...
-        try:
-            self.fail(msg)
-        except AttributeError:
-            raise AssertionError(msg)
-
 
 class Helper:
     """Utility functions for dryrun"""
 
+    @classmethod
```
tzaffi, line 536: These are used to build "Singleton" dry run requests.

algosdk/testing/dryrun.py
tzaffi: These changes have unit tests to back them and I've also added the new file `x/blackbox/dryrun_mixin_docs_test.py` for further confidence.

Makefile
michaeldiamant: @tzaffi In case helpful:  I _think_ https://github.com/algorand/py-algorand-sdk/pull/301/ may fix the issue you faced.  I see the branch does _not_ contain the fix provided by https://github.com/algorand/py-algorand-sdk/pull/301/.

So, it's possible merging in `develop` fixes the problem.

tzaffi:
:+1

Makefile
tzaffi: that seems to have done the trick. 🙏

Makefile
```diff
@@ -3,8 +3,11 @@ unit:
...
-integration:
+integration: blackbox
 	behave --tags=$(INTEGRATIONS) test -f progress2
 
+blackbox:
+	pytest x/blackbox
+
 docker-test:
-	./run_integration.sh
+	./run_integration.sh
```
tzaffi, line 13:
```suggestion
	./run_integration.sh

```

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,476 @@
+from collections import Counter
```
michaeldiamant, line 1: @tzaffi Friendly pointer to unused import.  Perhaps one day _soon_, a linter will help us here.  😅 

```suggestion

```

x/testnet.py
```diff
@@ -0,0 +1,35 @@
...
+DEVNET_TOKEN = "a" * 64
+ALGOD_PORT = 60000
+KMD_PORT = 60001
+INDEXER_PORTS = range(59_996, 60_000)
+
+LIVE_ALGOD: AlgodClient = None
+
+
+def get_algod(force_retry: bool = False) -> AlgodClient:
+    global LIVE_ALGOD
```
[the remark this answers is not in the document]
tzaffi: You make a good point... There's not a sufficient upside to having the global.

x/blackbox/blackbox_test.py
tzaffi: good catch

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,476 @@
...
+-------"""
+    )
+
+    # 2. Build the Dryrun requests:
+    drbuilder = DryRunHelper.build_simple_app_request
+    dryrun_reqs = list(
+        map(lambda a: drbuilder(teal, lightly_encode_args(a)), inputs)
+    )
+
+    # 3. Run the requests to obtain sequence of Dryrun resonses:
```
michaeldiamant, line 253: @tzaffi Friendly pointer to minor typo.

```suggestion
    # 3. Run the requests to obtain sequence of Dryrun responses:
```

x/testnet.py
```diff
@@ -0,0 +1,24 @@
...
+from algosdk.kmd import KMDClient
+from algosdk.v2client.indexer import IndexerClient
+
+DEVNET_TOKEN = "a" * 64
+ALGOD_PORT = 60000
+KMD_PORT = 60001
+INDEXER_PORTS = range(59_996, 60_000)
+
+
+def get_algod(force_retry: bool = False) -> AlgodClient:
```
michaeldiamant, line 11: @tzaffi Minor:  Following https://github.com/algorand/py-algorand-sdk/pull/298/commits/bad794f92942457f2a6e1f511a0a59a2247ba892, I think it's safe to remove the `force_retry` parameter.
```suggestion
def get_algod() -> AlgodClient:
```

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,720 @@
...
+    * simple python types - these are useful in the case of _constant_ assertions.
+        For example, if you want to assert that the `maxStackHeight` is 3, just use `3`.
+    * dictionaries of type Dict[Tuple, Any] - these are useful when you just want to assert
+        a discrete set of input-output pairs.
+        For example, if you have 4 inputs that you want to assert are being squared,
+        you could use `{(2,): 4, (7,): 49, (13,): 169, (11,): 121}`
+    * functions which take a single variable. These are useful when you have a python "simulator"
+        for the assertions.
+        In the square example you could use `lambda args: args[0]**2`
+    * functions which take _two_ variables. These are useuful when your assertion is more
```
michaeldiamant, line 329: @tzaffi Friendly pointer to a minor typo.

```suggestion
    * functions which take _two_ variables. These are useful when your assertion is more
```

algosdk/testing/dryrun.py
```diff
@@ -610,6 +634,89 @@ def build_dryrun_request(
...
+                        )
+                    )
+                else:
+                    accounts.append(acc)
+            app.accounts = accounts
+
+        return app
+
+    @classmethod
+    def _prepare_lsig(cls, lsig, sender):
```
michaeldiamant, line 677: @tzaffi Minor:  `sender` parameter is unused.  Is it by-design?  If so, can it be removed?

The remark applies to `_prepare_app` too.

algosdk/testing/dryrun.py
```diff
@@ -485,23 +522,33 @@ def _checked_request(
...
-            self.fail(msg)
-        except AttributeError:
-            raise AssertionError(msg)
-
 
 class Helper:
     """Utility functions for dryrun"""
 
+    @classmethod
+    def build_simple_logicsig_request(
```
michaeldiamant, line 537: @tzaffi Optional - Feel welcomed to close if you prefer as is.

I typically prefer to avoid including the term _simple_ in method names.  In my experience, the convention creates confusion when alternative factories appear in the future.

Alternatively, I'd name it according to its functionality _like_ `build_dryrun_request_for_logicsig` to follow the convention of `build_dryrun_request`.

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,475 @@
...
+    with open(csvpath, "w") as f:
+        f.write(csv_from_dryruns(inputs, dryrun_resps))
+
+    # 5. Sequential assertions (if provided any)
+    for i, type_n_assertion in enumerate(assertions.items()):
+        assert_type, assertion = type_n_assertion
+
+        assert mode_has_assertion(
+            mode, assert_type
+        ), f"assert_type {assert_type} is not applicable for {mode}. Please REMOVE of MODIFY"
```
michaeldiamant, line 267: @tzaffi Friendly pointer to what appears to be a minor typo.

```suggestion
        ), f"assert_type {assert_type} is not applicable for {mode}. Please REMOVE or MODIFY"
```

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,720 @@
...
+    assert (
+        len(txns) == 1
+    ), f"expecting exactly 1 transaction but got {len(txns)}"
+    txn = txns[0]
+    mode = (
+        ExecutionMode.Signature
+        if "logic-sig-messages" in txn
+        else ExecutionMode.Application
+    )
+    is_app = mode == ExecutionMode.Application
```
michaeldiamant, line 455: @tzaffi Friendly pointer to what appears to be a duplicate `is_app` definition.  It's defined again several lines below within the same lexical scope.  I _assume_ 1 of 2 the definitions can be safely removed.

```suggestion

```

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,720 @@
...
+        return ok or msg
+
+    if assert_type == DryRunAssertionType.noError:
+        ok, msg = assert_no_error(dryrun_resp, enforce=False)
+        return ok or msg
+
+    raise Exception(f"Unknown assert_type {assert_type}")
+
+
+def extract_logs(txn, decode_logs: bool = True):
```
michaeldiamant, line 512: @tzaffi Minor:  It seems `decode_logs` is _not_ used within the method definition.  Is there logic worth adding or is it safe to remove the parameter?

algosdk/testing/dryrun.py
tzaffi: 👍

algosdk/testing/teal_blackbox.py
tzaffi: 👍

algosdk/testing/dryrun.py
tzaffi: I wasn't sure about the wording, but I wanted to make it clear that there is something _simpler_ about this request. I've gone with the following for now. Let me know your thoughts:
* `singleton_logicsig_request()`
* `singleton_app_request)`
Though I'm dropping `dryrun` from the name, in context it should still be clear (if you import the library). EG:
```python
from algosdk.testing.dryrun import Helper as DryRunHelper

...
req = DryRunHelper.singleton_logicsig_request(...)
```

algosdk/testing/dryrun.py
michaeldiamant: Closing the loop - Thanks for bringing in the feedback via https://github.com/algorand/py-algorand-sdk/pull/298/commits/3c75c626513800983607a022ee8db6e097f39c6b.

algosdk/testing/dryrun.py
tzaffi: Resol.ving this zombie conversation. "simple" nomenclature has long been removed

michaeldiamant: @tzaffi I think I misunderstood the scope of _simple_ on my initial review.  I focused on the app vs logic specification rather than what I think you intended:  `List[bytes]` specification.

Having made another pass, here's my thinking:
* I realize `build_dryrun_request` predates the PR.  I think it's surprising the method supports invocation _without_ specifying execution mode (lsig vs app).  What do folks think about requiring an execution mode? 
  * It's unclear to me what backward compatibility guarantees _need_ to be maintained.  And perhaps I'm missing a use case. 
* I agree `dryrun` can be safely removed from the name.  Usage is qualified by the import.
* I'd consider one of the following options for `singleton` factory methods:
  * Do nothing and leave as is.
    * Personally, I'd prefer _not_ to include terms like _singleton_  in the method definition.  Though I don't feel strongly enough to pursue further.
  * If you're open to further renaming, I'd consider naming something specific to the functionality like:   `app_request_with_args`.
  * Delete the _singleton_ factory methods in favor of `build_dryrun_request`, _and_ add `args` factory methods to `App` and `LSig`.
    * If given a preference, I'd lean here.  
      * I think there's a large number of possible factory methods that _may_ exist for building a dryrun request.  I am thinking moving the factories one level up minimizes the surface area needing changes later.
      * Though I don't have enough experience using the repo to be confident about my thinking.

tzaffi: I don't feel very strongly about naming. I do think it's convenient to have separate helper functions for **app** versus **logicsig** because it makes it very apparent to the end-user that they need to think this through. These helper functions are easy enough to maintain, as they route immediately to the `build_dryrun_request()` workhorse. It's worth soliciting more feedback/opinions.

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,995 @@
...
+from algosdk.testing.dryrun import (
+    ZERO_ADDRESS,
+    assert_error,
+    assert_no_error,
+    Helper as DryRunHelper,
+)
+
+ExecutionMode = Enum("ExecutionMode", "Signature Application")
+
+DryRunProperty = Enum(
```
michaeldiamant, line 20: @tzaffi Beyond avoiding the declaration of enum indices, what other advantages does the _functional API_ (https://docs.python.org/3/library/enum.html#functional-api) provide for enum declaration?

* I also prefer the functional API's conciseness, but it seems to be mostly incompatible with at least PyCharm.  PyCharm seems unable to parse specific enumeration values from these strings.
* It's unclear if the limitation extends to mypy.  I made a minor change in pyteal where I referred to a non-existent enumeration value.  mypy did _not_ report an error.  https://mypy.readthedocs.io/en/stable/literal_types.html#enums suggests mypy does work with other forms of enum definitions.
* The IDE limitation feels hampering.  I think it'd make refactoring (e.g. finding usages) quite difficult.
* I'm wondering if we can get most of the benefit in a way that's more compatible with the tooling?
  * For example, https://docs.python.org/3/library/enum.html#using-automatic-values offers a middle ground on the manual index definition concern.

algosdk/testing/dryrun.py
```diff
@@ -527,80 +575,57 @@ def build_dryrun_request(
...
-                accounts = app.accounts
-                # app idx must match in sources and in apps arrays so dryrun find apps sources
-                source.app_index = application.id
-                rnd = app.round
-            else:
-                txns = [cls._build_logicsig_txn(program, txn, lsig)]
-            sources = [source]
+        if run_mode != "lsig":
+            txns = [cls._build_appcall_signed_txn(txn, app_or_lsig)]
+            application = cls.sample_app(sender, app_or_lsig, program)
```
michaeldiamant, line 623: @tzaffi For my edification:  `sample_app` supports supplying `program` optionally.  What does it mean to submit a dryrun request _without_ a program?

My intuition is that the request will fail.  I didn't go deep enough to vet my assumption.  And I'm asking in case you know off-hand.

algosdk/testing/teal_blackbox.py
tzaffi: 👍 - ok, I'll go with `auto()`

algosdk/testing/dryrun.py
tzaffi: You can provide program bytes instead, but I'm not supporting that functionality in the Blackbox Toolkit.

algosdk/testing/dryrun.py
```diff
@@ -527,80 +575,57 @@ def build_dryrun_request(
...
-        else:
-            if lsig is not None:
-                if not isinstance(lsig, LSig) and not isinstance(lsig, dict):
-                    raise ValueError("lsig must be a dict or LSig")
-                if isinstance(lsig, dict):
-                    lsig = LSig(**lsig)
-            else:
-                lsig = LSig()
+        del app
+        del lsig
```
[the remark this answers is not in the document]
tzaffi: By deleting `app` and `lsig` I'm asserting that you can't attempt to use that variable later in in the flow. That seemed very error prone to me, when I wanted to enforce that you have all you need in the variable `app_or_lsig`.

algosdk/testing/dryrun.py
michaeldiamant: @tzaffi Ah right, I keep forgetting a different path exists via program bytes.  Thanks for the explanation.  Resolving.

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,610 @@
...
+            DRProp.lastLog: Encoder.hex(2 ** 10),
+            # dicts have a special meaning as assertions. So in the case of "finalScratch"
+            # which is supposed to _ALSO_ output a dict, we need to use a lambda as a work-around
+            DRProp.finalScratch: lambda _: {0: 2 ** 10},
+            DRProp.stackTop: 2 ** 10,
+            DRProp.maxStackHeight: 2,
+            DRProp.status: "PASS",
+            DRProp.passed: True,
+            DRProp.rejected: False,
+            DRProp.noError: True,
```
jasonpaulos, line 196: Why would you assert on all of `status`, `passed`, `rejected`, and `noError`? It seems needlessly repetitive

x/blackbox/dryrun_mixin_docs_test.py
jasonpaulos: To confirm, this file does not have anything to do with the new blackbox testing?

Makefile
```diff
@@ -3,8 +3,11 @@ unit:
 	behave --tags=$(UNITS) test -f progress2
 
 INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send"
-integration:
+integration: blackbox
 	behave --tags=$(INTEGRATIONS) test -f progress2
 
+blackbox:
+	pytest x/blackbox
```
jasonpaulos, line 10: Instead of just running blackbox tests, it would be more useful to make an entry to run all local Python tests, which include the blackbox tests.

jasonpaulos (commented):
I've not yet looked into https://github.com/algorand/pyteal/pull/249, but here are my comments/questions so far from looking at this PR in isolation.

High level questions:
1. Is the `x` directory name temporary? If not, what is the justification? Logically it would make the most sense if this directory were grouped with the existing files `test_unit.py` and `test_integration.py` which also provide local (non-cucumber) tests.

2. It seems to me that `SequenceAssertion` is an opinionated and higher level way of verifying dryrun results. For that reason I argue it should not be included in this PR to the Python SDK, since the SDK should contain the minimal amount of code necessary to facilitate dryrun testing. If that were not the case, we would likely find ourselves in the awkward situation of requiring PRs to both PyTeal and this SDK in order to fix potential bugs or add new features -- we may still end up in that situation, but it would be less likely if the SDK features were minimal.

3. The relationship between this PR and https://github.com/algorand/py-algorand-sdk/pull/283 confuses me. You've expressed interest in merging this first, which seems backwards. #283 has a small, agreed-upon changeset that does not have many degrees of freedom to be altered since the same interface and operations are being implemented across every SDK for consistency. It seems more natural to first merge that, then attempt to use its features as much as possible in this PR.

4. Have you looked into working with "offline" dryrun results obtained from `goal clerk dryrun` at all? If possible, the framework here should be able to work with those as well.

tzaffi:
> I've not yet looked into [algorand/pyteal#249](https://github.com/algorand/pyteal/pull/249), but here are my comments/questions so far from looking at this PR in isolation.
> 
> High level questions:
> 
> 1. Is the `x` directory name temporary? If not, what is the justification? Logically it would make the most sense if this directory were grouped with the existing files `test_unit.py` and `test_integration.py` which also provide local (non-cucumber) tests.

`x` is expressing the idea that this is experimental (though it's feeling less experimental at this point as I've gone through a few iterations of the API). This can certainly be moved elsewhere. Let's decide where it should go.

> 2. It seems to me that `SequenceAssertion` is an opinionated and higher level way of verifying dryrun results. For that reason I argue it should not be included in this PR to the Python SDK, since the SDK should contain the minimal amount of code necessary to facilitate dryrun testing. If that were not the case, we would likely find ourselves in the awkward situation of requiring PRs to both PyTeal and this SDK in order to fix potential bugs or add new features -- we may still end up in that situation, but it would be less likely if the SDK features were minimal.

That makes sense. I found the `SequenceAssertion` quite useful in my own thinking and experimentation, but it is not strictly necessary. I propose that I move this code out of `teal_blackbox.py` and into the `blackbox_test.py` for direct usage (and reference example) but as a discouragement against importing. Worth further discussion.

> 3. The relationship between this PR and #283 confuses me. You've expressed interest in merging this first, which seems backwards. #283 has a small, agreed-upon changeset that does not have many degrees of freedom to be altered since the same interface and operations are being implemented across every SDK for consistency. It seems more natural to first merge that, then attempt to use its features as much as possible in this PR.
There was a race condition between #283 and the requirements of this PR so for practical reasons, I went ahead and wrote the code without relying on #283, and with full confidence that merging after the fact shouldn't be too difficult. But if it is important not to delay #283 any further, I'm glad to work with @barnjamin to get that merged ahead of this PR.

> 4. Have you looked into working with "offline" dryrun results obtained from `goal clerk dryrun` at all? If possible, the framework here should be able to work with those as well.
I have not been able to make much progress on this front.

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,610 @@
...
+
+    prop_assert(app_res, app_res.rejected(), False)
+    prop_assert(app_log_res, app_log_res.rejected(), False)
+    prop_assert(lsig_res, lsig_res.rejected(), False)
+    prop_assert(bad_lsig_res, bad_lsig_res.rejected(), True)
+
+    prop_assert(app_res, app_res.rejected(), False)
+    prop_assert(app_log_res, app_log_res.rejected(), False)
+    prop_assert(lsig_res, lsig_res.rejected(), False)
+    prop_assert(bad_lsig_res, bad_lsig_res.rejected(), True)
```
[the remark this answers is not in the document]
tzaffi: 👍

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,995 @@
...
+        return self.dig(DRProp.passed)
+
+    def rejected(self) -> bool:
+        """Assertable property for the program's dry run execution having FAILED
+        return type: bool
+        available: all modes
+        """
+        return self.dig(DRProp.rejected)
+
+    def error(self, pattern=None) -> bool:
```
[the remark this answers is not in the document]
tzaffi: 👍 (inheriting DryRunTestMixin's lingo, but no need to perpetuate it)

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,610 @@
...
+
+    prop_assert(app_res, app_res.error(), False)
+    prop_assert(app_log_res, app_log_res.error(), False)
+    prop_assert(lsig_res, lsig_res.error(), False)
+    prop_assert(bad_lsig_res, bad_lsig_res.error(), True)
+    assert (
+        bad_lsig_res.error(
+            pattern="logic 0 failed at line 7: log not allowed in current mode"
+        )
+        is True
```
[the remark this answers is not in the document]
tzaffi: 👍

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,995 @@
...
+        """Assertable property for a program having failed during dry run execution due to an error.
+        The optional `pattern` parameter allows specifying a particular string
+        expected to be a _substring_ of the error's message. In case the program errors, but
+        the pattern did not match the actual error, False is returned.
+            return type: bool
+            available: all modes
+        """
+        return self.dig(DRProp.error, pattern=pattern)
+
+    def noError(self) -> Union[bool, str]:
```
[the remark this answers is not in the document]
tzaffi: 👍

x/blackbox/blackbox_test.py
```diff
@@ -0,0 +1,610 @@
...
+    )
+
+
+APP_SCENARIOS = {
+    "app_exp": {
+        "inputs": [()],
+        # since only a single input, just assert a constant in each case
+        "assertions": {
+            DRProp.cost: 11,
+            # int assertions on log outputs need encoding to varuint-hex:
```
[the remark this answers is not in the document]
tzaffi: Looks like a comment for self-investigation that I forgot to remove

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,995 @@
...
+        cls, scenario: Dict[str, Union[list, dict]], mode: ExecutionMode
+    ) -> Tuple[List[tuple], Dict[DRProp, Any]]:
+        """
+        Validate that a Blackbox Test Scenario has been properly constructed, and return back
+        its components which consist of **inputs** and _optional_ **assertions**.
+
+        A scenario should adhere to the following schema:
+        ```
+        {
+            "inputs":       List[Tuple[Union[str, int]]],
```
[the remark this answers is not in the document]
tzaffi: 👍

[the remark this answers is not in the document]
tzaffi: 👍

x/blackbox/blackbox_test.py
tzaffi: Yes, it's repetitive. But I'm propagating the existing assertions in `DryRunTestMixin` up here and demonstrating their usage so anyone can pick and choose which ones they really need/want to use.

tzaffi:
image
Responding to @jasonpaulos - I added this test to help ensure that my refactoring of DryRunTestMixin was backwards compatible. However, it is not strictly necessary and I can remove it.

Makefile
tzaffi: I'm not sure what you mean here. This is not only a local test, but is a new integration test. Do you mean that it should be renamed, or that I should also add `test_unit.py` and `test_integration.py` to a section that looks something like:

```Makefile
python-specific:
    pytest x/blackbox unit_test.py integration_test.py
```
?

tzaffi: Ok I stewed a bit on this, I have a proposal. Restructure the test files as follows into new directories:
* `test/local`
  * `test/local/unit` 
    * `test_unit.py` (rename to `unit_test.py` ?)
  * `test/local/integration`
    * `test/local/testnet.py` (rename to `clients.py` ?)
    * `test_integration.py` (rename to `integration_test.py` ?)
      * NOTE this test is getting its client info from `examples/tokens`. Perhaps it should get it instead from `testnet.py`
    *  `blackbox_test.py`
    *  `dryrun_mixin_docs_test.py` (if we end up keeping it)
The now orphaned `README.md` could move to `algoskd/testing/README.md`

algosdk/testing/teal_blackbox.py
michaeldiamant: @tzaffi Thanks for bringing in `Enum` with `auto()` via https://github.com/algorand/py-algorand-sdk/commit/680e432bf586f59605f32e243690277bff715ee0.  It's working well with IDE facilities.

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,1010 @@
...
+        return f"0x{b64decode(self.b).hex()}" if self.is_b else str(self.i)
+
+    def as_python_type(self) -> Union[int, str, None]:
+        if self.is_b is None:
+            return None
+        return str(self) if self.is_b else self.i
+
+
+@dataclass
+class BlackBoxResults:
```
michaeldiamant, line 76: @tzaffi I've opened an optional PR for naming consistency via https://github.com/algorand/py-algorand-sdk/pull/304.   Can you have a look?

Let's resolve the thread once https://github.com/algorand/py-algorand-sdk/pull/304 is merged or closed.

x/blackbox/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+* Track the program execution by viewing its **App Trace**
+  * 2 was assigned to **scratch slot #0** at step 5
+  * the stack ended up with **4** on top
+  * the run **PASS**'ed
+* Read the message parameter that was provided and which explains in English what went wrong: `<<<<<<<<<<<expected 8 but got 4>>>>>>>>>>>>>`
+
+### EDRA: Exploratory Dry Run Analysis
+
+Let's expand our investigation from a single dry-run to to multiple runs or a **run sequence**. In other words, given a sequence of inputs, observe _assertable properties_ for the corresponding
```
michaeldiamant, line 200: @tzaffi Friendly pointer to minor typo.

```suggestion
Let's expand our investigation from a single dry-run to multiple runs or a **run sequence**. In other words, given a sequence of inputs, observe _assertable properties_ for the corresponding
```

Makefile
```diff
@@ -1,10 +1,24 @@
...
+cuke-integ: local-integration
+	behave --tags=$(INTEGS) test -f progress2
+
+unit: local-unit cuke-unit
+
+integration: local-integ cuke-integ
 
 docker-test:
 	./run_integration.sh
+
```
tzaffi, line 24:
```suggestion
```

Makefile
```diff
@@ -1,10 +1,25 @@
...
+include z.Makefile
```
tzaffi, line 25:
```suggestion
```

x/blackbox/README.md
tzaffi: fixing elsewhere

algosdk/testing/teal_blackbox.py
tzaffi: fixed 🙏

Makefile
```diff
@@ -1,10 +1,21 @@
+local-unit:
+	pytest test/unit
+	
+local-integ:
+	pytest test/integration/blackbox_test.py test/integration/dryrun_mixin_docs_test.py
+	# NOT WORKING: pytest test/integration/integration_test.py
```
tzaffi, line 6: Does `test_integration.py` actually pass for anyone?

tzaffi: @jasonpaulos @michaeldiamant ^^^

x/blackbox/dryrun_mixin_docs_test.py
tzaffi: Confirmed: I added it to ensure that refactoring `DryRunTestMixin` did not break existing functionality.

tzaffi:
I asked this on PR #283 but I'm asking it here too because I'm also importing `tabulate`:

> Question for @jasonpaulos . This PR (and #298 as well) adds a new package dependency `tabulate`. Should this be added to our `setup.py` here?

> https://github.com/algorand/py-algorand-sdk/blob/7c0b8a4bccfbf6a512c690b6f07707f9cc2d9b71/setup.py#L19

tzaffi:
I asked this on PR #283 but I'm asking it here too because I'm also importing `tabulate`:

Question for @jasonpaulos . This PR (and #298 as well) adds a new package dependency `tabulate`. Should this be added to our `setup.py` here?

https://github.com/algorand/py-algorand-sdk/blob/7c0b8a4bccfbf6a512c690b6f07707f9cc2d9b71/setup.py#L19

Makefile
michaeldiamant: @tzaffi It doesn't pass for me as of https://github.com/algorand/py-algorand-sdk/pull/298/commits/a47355bf7576e73b06cd8c4f3052f1aec31962b7.  

I spent only a few seconds reviewing.  My guess is the error predates the PR.  My unsubstantiated intuition is that the test is incompatible with running algod in a container.  I'm noting the local disk access attempts to resolve kmd.

Obviously, it's an undesirable state.  However, if we agree the issue predates the PR, then I think we ought to file an issue and circle back later.

```
~/dev/py-algorand-sdk teal-blackbox !2 ─────────────────────────────────────────────────────────────────────────────────────────────────── 12s  py-algorand-sdk 11:55:43 AM
❯ make local-integ
# pytest test/integration/blackbox_test.py test/integration/dryrun_mixin_docs_test.py
pytest test/integration/integration_test.py
============================================================================ test session starts ============================================================================
platform darwin -- Python 3.10.2, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/michael/dev/py-algorand-sdk
collected 0 items / 1 error

================================================================================== ERRORS ===================================================================================
___________________________________________________________ ERROR collecting test/integration/integration_test.py ___________________________________________________________
test/integration/integration_test.py:6: in <module>
    from examples import tokens
examples/tokens.py:27: in <module>
    for directory in listdir(data_dir_path):
E   FileNotFoundError: [Errno 2] No such file or directory: '/Users/michael/node/network/Node/'
```

jasonpaulos: Yes, as far as I can tell this test hasn't been ran ever on CI, so it's likely to be broken in a few ways. I don't think this PR should attempt to fix it, but perhaps putting a comment at the top of it might make sense

Makefile
```diff
@@ -1,10 +1,21 @@
...
 UNITS = "@unit.abijson or @unit.algod or @unit.applications or @unit.atomic_transaction_composer or @unit.dryrun or @unit.feetest or @unit.indexer or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring"
-unit:
+cuke-unit:
 	behave --tags=$(UNITS) test -f progress2
 
-INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send"
-integration:
-	behave --tags=$(INTEGRATIONS) test -f progress2
+INTEGS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send"
+cuke-integ:
```
jasonpaulos, line 13: nit: can we use `cucumber-unit` and `cucumber-integration` instead? It's more specific, and there's no need to be brief because the common use case will be to run `make unit` or `make integration` instead of invoking these individually.

Also, if `make unit` runs `local-unit` now, it's not strictly necessary to have a step for `make local-unit` in `.circleci/config.yml`, since the unit tests will get ran by the algorand-sdk-testing harness

Makefile
jasonpaulos: @tzaffi I see you decided against making the `test/cucumber` folder, why was that? I still recommend it for clarity because a lot of the cucumber setup is strange and it's best not to mix that with "normal" tests

Makefile
michaeldiamant: I'll open an issue on the topic that we can link inline.  I'm in-between a few items, but I'll supply an issue _soon_ and then we can safely resolve the thread.

michaeldiamant: @tzaffi I opened https://github.com/algorand/py-algorand-sdk/pull/307 to bring the thread to a close.  Let's resolve once #307 is merged.

test/integration/BLACKBOX_README.md
```diff
@@ -0,0 +1,398 @@
+# TEAL Blackbox Toolkit: Program Reporting and Testing via Dry Runs
+
+**NOTE: to get math formulas to render here using Chrome, add the [xhub extension](https://chrome.google.com/webstore/detail/xhub/anidddebgkllnnnnjfkmjcaallemhjee/related) and reload**
+
+## Blackbox Testing Howto
+
+### What is TEAL Blackbox Testing?
+
+TEAL Blackbox Testing lets you treat your TEAL program as a black box that for every received input produces an output and other observable effects. You can create reports that summarize those effects, and turn the _reports_ into _program invariant conjectures_ which you then check with _sequence assertions_.
```
algoanne, line 9: I personally find this paragraph a little daunting, and it makes me think "do I really need this complicated tool?" (especially the second sentence)

I think bullets on line 51-53 are saying a similar thing, but itemized in a way that feels friendlier/easier to digest - lets me easily see I could just use "make basic assertions and validate them using dry runs" if I want to, to start. That sounds useful to me already!

test/integration/BLACKBOX_README.md
```diff
@@ -0,0 +1,398 @@
...
+* stack's height
+* scratch variables
+* final log message (this is especially useful for [ABI-compliant programs](https://developer.algorand.org/docs/get-details/dapps/smart-contracts/ABI/))
+* status (**PASS**, **REJECT** or _erroring_)
+* error conditions that are and aren't encountered
+
+Even better, before making fine-grained assertions you'd like to get a sense of what the program is doing on a large set of inputs so you can discover program invariants. The TEAL Blackbox Toolkit's recommended approach for enabling these goals is to:
+
+* start by making basic assertions and validate them using dry runs (see "**Basic Assertions**" section below)
+* execute the program on a run-sequence of inputs and explore the results (see "**EDRA: Exploratory Dry Run Analysis**" section below)
```
algoanne, line 52: what does "run-sequence" mean? Is it the same as a "sequence" of inputs or does it have additional meaning?

test/integration/BLACKBOX_README.md
```diff
@@ -0,0 +1,398 @@
+# TEAL Blackbox Toolkit: Program Reporting and Testing via Dry Runs
```
algoanne, line 1: I see "TEAL Blackbox Toolkit" and "TEAL Blackbox Testing" : are these two different concepts or the same? (is there more to the toolkit than testing?)

test/integration/BLACKBOX_README.md
tzaffi: yeah, just a sequence of inputs.

test/integration/BLACKBOX_README.md
tzaffi: 👍

test/integration/BLACKBOX_README.md
```diff
@@ -0,0 +1,398 @@
...
+
+**STEP 4**. Write the TEAL program that you want to test. You can inline the test as described here or follow the approach of `x/blackbox/blackbox_test.py` and save under `x/blackbox/teal`. So following the inline
+appraoch we begin our TEAL Blackbox script with an <a name="teal">inline teal source variable</a>:
```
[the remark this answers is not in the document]
tzaffi: 👍

.circleci/config.yml
```diff
@@ -16,7 +16,6 @@ jobs:
       - checkout
       - run: pip install -r requirements.txt
       - run: black --check .
-      - run: python3 test_unit.py
```
tzaffi, line 19: this is now running in docker

Makefile
tzaffi: Am trying to revive the integration test, but am skipping 2 failures

algosdk/testing/teal_blackbox.py
```diff
@@ -0,0 +1,1010 @@
+from base64 import b64decode
+import csv
+from dataclasses import dataclass
+from enum import Enum, auto
+import io
+from inspect import signature
+from tabulate import tabulate
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
```
tzaffi, line 8:
```suggestion
from typing import Any, Callable, Dict, Sequence, List, Optional, Tuple, Union
```

tzaffi (commented):
`Sequence` is more suitable than `Iterable`

tzaffi:
this work is now in the [graviton](https://github.com/algorand/graviton) repo

Issue #284, opened 2022-02-02
Min fee not used in suggested parameters


plasmatech8:
### Subject of the issue

Note that I have discussed and described the issue [here](https://forum.algorand.org/t/cannot-calculate-suggested-fees-for-inner-transactions/5826/9?u=plasmatech8)

`suggested_params.min_fee` has no effect on the transaction object when passed into a constructor 
i.e.
```python
transaction.PaymentTxn(
    sp=suggusted_params,  
   # ...
)
```

this is because py-algorand-sdk [uses a constant](https://github.com/algorand/py-algorand-sdk/blob/2ef7d3312b6a2bd7bcb2dc430668060ae8ed80d8/algosdk/future/transaction.py#L367-L370) (`constants.min_txn_fee`) instead of the min-fee which is stored in the object (`sp.min_fee`).

```python
if not sp.flat_fee:
  self.fee = max(
      self.estimate_size() * self.fee, constants.min_txn_fee
  )
```

I do not know if this is intentional or not. 

However, it is a little bit more convenient when you want to do fee calculations by setting a min fee equal to a multiple of the number of transactions (namely for inner transactions). Also, will min-fee change in the future? If so then I assume people would get it via suggested params API.

### Your environment

* Python 3.7.6
* py-algorand-sdk==1.9.0

### Steps to reproduce

```python
suggusted_params = client.suggested_params()
suggusted_params.min_fee = 2000
txn = transaction.PaymentTxn(
    sender=sender_addresss,
    sp=suggusted_params,
    receiver=reciever_address,
    amt=amount
)
print(txn.fee)
```

### Expected behaviour

"2000"

### Actual behaviour

"1000"

winder:
At the moment there is rarely any network congestion, so the min fee should always be used.

We can revisit this in the future when that changes.

PR #299, opened 2022-03-06
Change min fee from constant to suggested param

plasmatech8:
I do not know what peoples thoughts are about this is since I did not receive a response from #284.

This PR changes the fee calculations for Transaction classes under future.transaction so that it uses `sp.min_fee` instead of `constants.min_txn_fee` in calculations when flat_fee=false.

Rationale:
* It is more intuitive to use min_fee in the SuggestedParams which the user is actively passing into the Transaction object, rather than a hidden/hard-coded value.
* Application calls with inner transactions require a higher min-transaction fee. It is more convenient to set the value `suggested_params.min_fee *= 2` rather than configuring flat fees.
* The min-fee might change in the future, and it is better to use the value returned from the Algorand REST API than a hard-coded constant.

I tried running tests using `make docker-test` but it is excruciatingly slow for some reason, so I will think about that another time.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -366,7 +366,7 @@ class PaymentTxn(Transaction):
         self.close_remainder_to = close_remainder_to
         if not sp.flat_fee:
             self.fee = max(
-                self.estimate_size() * self.fee, constants.min_txn_fee
+                self.estimate_size() * self.fee, sp.min_fee
             )
 
     def dictify(self):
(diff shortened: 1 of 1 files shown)


algochoi:
Thanks for your patience - I think the `SuggestedParameters` may or may not have the `min_fee` field, and it's causing our current tests to fail:

```
 File "/py-algorand-sdk/algosdk/future/transaction.py", line 369, in __init__
    self.estimate_size() * self.fee, sp.min_fee
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

jannotti:
It'll go in in #530

PR #304, opened 2022-03-22
Rename BlackBoxResults to BlackboxResults

michaeldiamant:
Optionally suggests a minor rename for consistency.  Since #298 applies the terminology _blackbox_, the PR suggests renaming from `BlackBoxResults` to `BlackboxResults`.

If the preference is to retain as is, then I welcome closing the PR.


tzaffi (approved):
🙏

PR #308, opened 2022-03-25
Adding method to verify signature on SignedTransaction object

barnjamin:
I can do this across SDKs if this is reasonable

Should fill the need of https://github.com/algorand-devrel/community/issues/10

API is 
```py
#...
pk, sk = get_account()
pay_txn = transaction.PaymentTxn(pk, sp, rcv, 10000)
stxn = pay_txn.sign(sk)
if not stxn.verify_signature():
    print("failzore")
#...
```

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -2145,6 +2145,35 @@ class SignedTransaction:
         """
         return self.transaction.get_txid()
 
+    def verify_signature(self):
+        """
+        Verify the signature on the transaction is valid
+
+        Returns:
+            bool: whether or not the signature is valid
+        """
+
+        if self.signature is None or len(self.signature) == 0:
+            # TODO: exception?
+            return False
+
+        public_key = self.transaction.sender
+        if self.authorizing_address is not None:
+            public_key = self.authorizing_address
+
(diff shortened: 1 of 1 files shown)


algosdk/future/transaction.py
```diff
@@ -2145,6 +2145,35 @@ def get_txid(self):
...
+
+        prefixed_message = constants.TXID_PREFIX + base64.b64decode(
+            encoding.msgpack_encode(self.transaction)
+        )
+        try:
+            verify_key.verify(
+                prefixed_message, base64.b64decode(self.signature)
+            )
+            return True
+        except BadSignatureError:
```
algochoi, line 2174: I think `nacl`'s `VerifyKey` API can now return a [`TypeError` or `ValueError`](https://github.com/pyca/pynacl/blob/b3f6c320569d858ba61d4bdf2ac788564528c1c9/src/nacl/signing.py#L98). The multisig/logicsig `verify` in the Python SDK doesn't catch these errors yet either so I think it would be nice to update them :D

algochoi (commented):
I was just wondering why this API has different parameters from the existing `verify` methods. I don't have a strong opinion on this, but I'm also curious on the naming (opposed to `raw_verify` or something suggested on the original ticket.

Issue #286, opened 2022-02-08
Deprecate non-future transactions module


barnjamin:
## Problem

Often when getting started devs will use the `algosdk.transactions` module but run into issues since the `algosdk.future.transactions` module is the one that should be used. 

The same type of issue happens when they try to use `msgpack_decode` instead of `future_msgpack_decode`

## Solution

Find a nice way to deprecate `algosdk.transactions` or move `algosdk.future.transactions` to `algosdk.transactions` so this doesn't happen

algochoi:
Agreed 👍  related to #252 

I think the only concern we had was backwards compatibility, but hopefully everyone has migrated over to the `future` package by now.

winder:
This is being addressed through other issues.

PR #279, opened 2022-01-20
Updates to pipeline

bricerisingalgorand:
Update circleci job to use docker instead of vms for unit tests. Also updated the unit tests to fail when they fail

Issue #252, opened 2021-11-18
Make future.transaction the transaction


algoanne:
And deprecate the old code, because it has many bugs in it that we have not addressed. 

Could begin by renaming transaction to old.transaction.

algochoi:
Found this related ticket (from Nov. 2020!): https://github.com/algorand/go-algorand-internal/issues/861

PR #309, opened 2022-03-25
Catch TypeError and ValueError in verify functions

algochoi:
`pynacl`'s `VerifyKey()` can now return a [`TypeError` or `ValueError`](https://github.com/pyca/pynacl/blob/b3f6c320569d858ba61d4bdf2ac788564528c1c9/src/nacl/signing.py#L98) ([related PR](https://github.com/pyca/pynacl/pull/600/files)). Our SDK should reflect this change by catching those errors as well, or users may encounter an unexpected exception. 

Picked some major release versions and set that as boundaries. Hopefully users are using `venv` or `conda` as well...
* [pynacl](https://github.com/pyca/pynacl/tags)
* [msgpack](https://github.com/msgpack/msgpack-python/releases)
* [pycryptodomex](https://github.com/Legrandin/pycryptodome/releases)

Also added the `black` formatter updates so CI can run normally.

I didn't change the v1 `transaction` methods, but the problem is there as well (I hope no one uses them because we haven't updated them for a very long time).

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -2393,7 +2393,7 @@ class Multisig:
                 try:
                     verify_key.verify(message, subsig.signature)
                     verified_count += 1
-                except BadSignatureError:
+                except (BadSignatureError, ValueError, TypeError):
                     return False
 
         if verified_count < self.threshold:
@@ -2562,7 +2562,7 @@ class LogicSig:
             try:
                 verify_key.verify(to_sign, base64.b64decode(self.sig))
                 return True
-            except (BadSignatureError, ValueError):
+            except (BadSignatureError, ValueError, TypeError):
                 return False
 
         return self.msig.verify(to_sign)
--- a/algosdk/abi/tuple_type.py
+++ b/algosdk/abi/tuple_type.py
@@ -17,7 +17,7 @@ class TupleType(ABIType):
     """
 
     def __init__(self, arg_types: List[Any]) -> None:
-        if len(arg_types) >= 2 ** 16:
+        if len(arg_types) >= 2**16:
             raise error.ABITypeError(
                 "tuple args cannot exceed a uint16: {}".format(len(arg_types))
             )
@@ -143,7 +143,7 @@ class TupleType(ABIType):
         Returns:
             bytes: encoded bytes of the tuple
         """
-        if len(self.child_types) >= (2 ** 16):
+        if len(self.child_types) >= (2**16):
             raise error.ABIEncodingError(
                 "length of tuple array should not exceed a uint16: {}".format(
                     len(self.child_types)
@@ -198,7 +198,7 @@ class TupleType(ABIType):
         for i in range(len(heads)):
             if is_dynamic_index[i]:
                 head_value = head_length + tail_curr_length
-                if head_value >= 2 ** 16:
+                if head_value >= 2**16:
                     raise error.ABIEncodingError(
                         "byte length {} should not exceed a uint16".format(
                             head_value
--- a/algosdk/abi/ufixed_type.py
+++ b/algosdk/abi/ufixed_type.py
@@ -69,7 +69,7 @@ class UfixedType(ABIType):
         """
         if (
             not isinstance(value, int)
-            or value >= (2 ** self.bit_size)
+            or value >= (2**self.bit_size)
             or value < 0
         ):
             raise error.ABIEncodingError(
--- a/algosdk/abi/uint_type.py
+++ b/algosdk/abi/uint_type.py
@@ -54,7 +54,7 @@ class UintType(ABIType):
         """
         if (
             not isinstance(value, int)
-            or value >= (2 ** self.bit_size)
+            or value >= (2**self.bit_size)
             or value < 0
         ):
             raise error.ABIEncodingError(
--- a/algosdk/util.py
+++ b/algosdk/util.py
@@ -69,7 +69,7 @@ def verify_bytes(message, signature, public_key):
     try:
         verify_key.verify(prefixed_message, base64.b64decode(signature))
         return True
-    except BadSignatureError:
+    except (BadSignatureError, ValueError, TypeError):
         return False
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 .
-black==21.9b0
+black==22.3.0
 glom==20.11.0
 pytest==6.2.5
 git+https://github.com/behave/behave
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,11 @@ setuptools.setup(
     project_urls={
         "Source": "https://github.com/algorand/py-algorand-sdk",
     },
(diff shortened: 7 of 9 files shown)


algosdk/future/transaction.py
```diff
@@ -2393,7 +2393,7 @@ def verify(self, message):
                 try:
                     verify_key.verify(message, subsig.signature)
                     verified_count += 1
-                except BadSignatureError:
+                except (BadSignatureError, ValueError, TypeError):
```
tzaffi, line 2396: I'm guessing that most users simply want to know that `verify()` has failed, deal with it, and won't really care why this has occurred (at least not in their runtime handling). For such users, swallowing these two exceptions makes life easier. 

But could there be something about these different errors that could allow users to have more nuanced recovery strategies? I kind of doubt it, but it's worth knowing before we just add the blanket conversion from error to False.

tzaffi: Thinking more about it, `ValueError` and `TypeError` sounds like they could be generated by bad user input/usage. If that is the case, then we really wouldn't want to swallow these errors.

algosdk/future/transaction.py
```diff
@@ -2393,7 +2393,7 @@ def verify(self, message):
                 try:
                     verify_key.verify(message, subsig.signature)
                     verified_count += 1
```
michaeldiamant, line 2395: @algochoi A couple of questions for my edification:
* For my edification:  How did you become aware of the change?
  * https://github.com/pyca/pynacl/pull/600/ was merged > 1 year ago.  I originally thought perhaps the change is recent.
* What is the policy for locking dependency versions?
  * https://github.com/algorand/py-algorand-sdk/blob/develop/setup.py suggests the version is _not_ locked. 
  *  To avoid future, unvetted API changes, I'm wondering if it's prudent to lock the version.  Though I'm generally unfamiliar with SDK conventions, so I'm raising the question.

algosdk/future/transaction.py
algochoi: Good point - I also thought about this and couldn't reach a conclusive decision. Our current unit tests do seem to test the `ValueError` here by testing the wrong number of bytes: https://github.com/algorand/py-algorand-sdk/blob/f0c9430ea6fda8e47ddcadd0579d36f5c66317e0/test_unit.py#L2694 

I'm not exactly sure how we want `verify` to behave in these cases so I wanted other people's inputs as well. If we choose to include this change, I can also add some test cases to check the two new error types.

algosdk/future/transaction.py
algochoi: * Brice noticed it in the repo a while ago and we tracked it down to the `pynacl` change - you can also see some of the discussion here, but some of it was discussed via slack: https://github.com/algorand/py-algorand-sdk/pull/279
* I'm not sure if such convention exists, but locking dependencies sound like a good decision to me. On the other hand, I'm not sure how we will be notified of more critical updates that happen to our dependencies (security patches? I guess if it's a really important patch, the word will get around anyway).

jasonpaulos: Since pip/python does not allow multiple different versions of a package to be installed (unlike JS), the best practice _appears_ to be that runtime dependencies should not be pinned to a specific version, since otherwise it would be nearly impossible to satisfy pinned dependencies when multiple packages are installed. That being said, I think it would be ok to enforce version minimum and/or maximum versions of packages, as long as those are reasonable.

However, for test/development dependencies (the items in our `requirements.txt`), we _should_ pin those, since they are only installed if you're developing on PyTeal, and it's best to have a consistent environment for that.

michaeldiamant: @jasonpaulos Thanks for the context RE: dependency management.  Some of my commentary stems from lack of familiarity with the Python environment.

Based on the above commentary, enforcing a major version lower + upper bound feels prudent.  

I feel more strongly about the upper bound because it's liable to break the build as new software is released.  Though I'll follow the group's convention.

algochoi: I bounded the package versions to some major releases - updated the PR description with links to these as well. It looks like popular packages do this as well:
* [requests](https://github.com/psf/requests/blob/main/setup.py)
* [web3.py](https://github.com/ethereum/web3.py/blob/master/setup.py)

I did enforce a lower bound on `pynacl` version `1.4.0`, which was introduced in 2020, because it introduces the changes that motivated this PR. `pycryptodomex` already had an existing mimum bound on the package of `3.6.0`.

michaeldiamant: @algochoi Looks reasonable to me - thanks for your patience through a couple of iterations.  Resolving.

algosdk/future/transaction.py
```diff
@@ -2393,7 +2393,7 @@ def verify(self, message):
                 try:
                     verify_key.verify(message, subsig.signature)
```
michaeldiamant, line 2394: @algochoi Opening a thread to clarify a point I missed earlier:

> I didn't change the v1 transaction methods, but the problem is there as well 

I'm unfamiliar with the _v1_ vs other version semantics.  As a sanity check:  Is there a reason _not_ to apply the PR's treatment to to _v1_ transaction methods?
* From looking around, I gathered you're referring to https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/transaction.py#L1208.
* I'm asking the question assuming there _is_ a reason not to make the update.  If easier to discuss live, let's do so.

algosdk/future/transaction.py
michaeldiamant: Summarizing a verbal with @algochoi, @tzaffi, and @jasonpaulos:
* We agreed the expedient path is to keep the PR's behavior and add unit tests validating behavior.
* We left with a loose agreement that revisiting the return type out-of-band is worthwhile.  For example, moving from boolean to a richer data type will simplify disambiguating failure cases.

As of https://github.com/algorand/py-algorand-sdk/pull/309/commits/c3b94a251444affa826a98a2a38eada8df86b1eb, I think the PR matches what we verbally agreed to.  Since I didn't open the thread, I'll leave as is without resolving.

algosdk/future/transaction.py
algochoi: Bingo, that's the right v1 transaction you're referring to.

We have several issues to deprecate the v1 module in the Python SDK (no longer supported) but haven't done it yet:
* https://github.com/algorand/py-algorand-sdk/issues/252
* https://github.com/algorand/py-algorand-sdk/issues/286

michaeldiamant: @algochoi Thanks for the pointers.  I took note of these issues on the deprecation + graduated levels of support front.

Since it's been the prevailing convention seems to be to avoid updates, then I'll follow the convention.  I confirmed the last change to `algosdk/transaction.py` is from 08/2021.  

Though I'm following the convention, I'm noting for completeness that it feels a little uncomfortable to ignore the API because it's _not_ deprecated.

Resolving.

michaeldiamant (approved):
@algochoi Thanks for the effort to clean up API inconsistencies here. ☕

PR #310, opened 2022-03-28
Bump minimum supported Python version to 3.7

algochoi:
This PR deletes the outdated `travis` configurations, and updates the `python_requires` version from `3.5` to `3.7`. 

Currently, the minimum Python version supported by `pynacl`, one of our dependencies, is `3.6`. It makes sense for our SDK to move to a higher version as well. 

Based on [pypistats](https://pypistats.org/packages/py-algorand-sdk), it looks like <3% of users are installing the SDK with a Python version  `<=3.6`, while around ~20% of users are installing it with `3.7`.


algosdk/abi/tuple_type.py
```diff
@@ -17,7 +17,7 @@ class TupleType(ABIType):
     """
 
     def __init__(self, arg_types: List[Any]) -> None:
```
michaeldiamant, line 19: @algochoi Sharing a few thoughts on 3.7 vs 3.10.  Let me know how you'd like to proceed and/or if you'd like to discuss verbally?

I'm open to a few routes that ultimately lead to 3.10:
* Change the current PR.
* Merge PR as is.  Release.  Wait ~1 week and merge + release follow on PR that bumps to 3.10.

I know the 3.10 move requires some developers to manage Python versions.  Optionally, I am up for adding a brief README or similar tutorial helping developers upgrade.  Example tooling for managing multiple Python versions includes http://asdf-vm.com/ and https://github.com/pyenv/pyenv.

My starting position is that I'd like to avoid a series of incremental upgrades because we're describing bumping to a new minor version.  So, I'd like to minimize the number of jumps made in the iteration.

As maintainers, I'd like to get to more recent Python versions because the language provides new facilities that can help us more easily provide error-free software (e.g. pattern matching).

Issue #300, opened 2022-03-11
Upgrade minimum Python version to 3.10


michaeldiamant:
## Problem

Via https://github.com/algorand/pyteal/issues/244, we plan to upgrade the minimum PyTeal Python version to 3.10.  Provided no py-algorand-sdk specific concerns exist, we'd like to make the same change here.

## Solution

We'll need to do the following:
* Provide advance community notice about the upcoming change.
* Update the build to reflect the change in minimum Python version.
* For PyTeal developer convenience, document how to configure a project-specific Python version.

## Dependencies

* Provide _sufficient_ notice for the community to hear about the pending change.
* Consider if community outreach ought to be synced with PyTeal change.

## Urgency

Depends on desire to sync changes with PyTeal minimum version bump.

algochoi:
Some stats on how people are downloading PyTeal here: https://pypistats.org/packages/pyteal

michaeldiamant:
After group discussion, we're opting to bump to Python 3.8 based on a group-generated rubric.

Here's related PRs:
* https://github.com/algorand/py-algorand-sdk/pull/322 - _Add minimum Python version policy to README_
* https://github.com/algorand/py-algorand-sdk/pull/323 - _Bump minimum Python version to 3.8_
* https://github.com/algorand/py-algorand-sdk/pull/325 - _Matrix test across Python versions for unit tests_

PR #313, opened 2022-03-30
Bump up minimum Python version to 3.10

algochoi:
This PR deletes the outdated travis configurations, and updates the python_requires version from 3.5 to 3.10. It also updates the Circle/Docker image Python versions to `3.10.x` as well. 

Currently, the minimum Python version supported by pynacl, one of our dependencies, is 3.6. It makes sense for our SDK to move to a higher version as well. Since PyTeal depends on this repo, we should also bump up the minimum Python version there too.

Closes https://github.com/algorand/py-algorand-sdk/issues/300


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ workflows:
 jobs:
   unit-test:
     docker:
-      - image: python:3.7.9
+      - image: python:3.10.4
```
jasonpaulos, line 14: Why not just use 3.10? I think the patch version need not be pinned down

algochoi: Good point - updated

docs/requirements.txt
```diff
@@ -1,3 +1,3 @@
-sphinx==4.0.1
+sphinx==4.2.0
```
algochoi, line 1: sphinx 4.2.0 is the first version that seems to work with Python 3.10, see discussion [here](https://github.com/sphinx-doc/sphinx/issues/9562)

michaeldiamant (approved):
@algochoi Thanks for the effort to bump the minimum Python version.  

Take my approval with a grain of salt due to my lack of py-algorand-sdk experience.  Here's what I checked:
* Visually sanity checked build steps to confirm tests executing as intended.
* Opened a draft PR to confirm pyteal works as a consumer:  https://github.com/algorand/pyteal/pull/275.
* Looked for missed Python version references via ` git grep "3\."`.

jasonpaulos (approved):
Looks good

michaeldiamant (changes requested):
@algochoi Sorry for the late change, but I dismissed my approval due to https://github.com/algorand/pyteal/pull/276.

Let's wait until https://github.com/algorand/pyteal/pull/276 is merged to confirm our understanding.

It feels like the repo ought to define a `.readthedocs.yaml`.  I don't (yet) have a readthedocs login, so I don't know what web configurations exist.  And it seems preferable to rely on a local, versioned config.

docs/requirements.txt
```diff
@@ -1,3 +1,3 @@
...
 sphinx-rtd-theme
```
michaeldiamant, line 2: @algochoi With the _suggestion_ below, I think the PR is ready for approval.

https://github.com/algorand/pyteal/pull/279 is the corresponding change set that fixed the PyTeal readthedocs build.  Thanks for reflecting `.readthedocs.yaml` into the PR.

```suggestion
sphinx-rtd-theme==1.0.0
```

algochoi: Thanks for the catch - will commit the suggestion

jasonpaulos (approved):
Readthedocs changes look good

michaeldiamant (approved):
@algochoi Thanks for the effort here to bring py-algorand-sdk up to Python 3.10!  ☕

PR #317, opened 2022-04-21
adding condition for allowing rcv to be none if close to is set

### LINKED ISSUE #316: Payment transaction decode with no Receiver
While decoding transactions from blocks, if a Payment Transaction includes no receiver but uses only close-to, the `future_msgpack_decode` fails with a complaint about no value set for rcv

We can either tweak this line to set the zero address instead of None
https://github.com/algorand/py-algorand-sdk/blob/ee24b9efd039bdb5473de3b9508183f55dc78dbb/algosdk/future/transaction.py#L395-L397

Or, maybe a better option is just to not require/error out when no rcv is set here
https://github.com/algorand/py-algorand-sdk/blob/ee24b9efd039bdb5473de3b9508183f55dc78dbb/algosdk/future/transaction.py#L358-L361

barnjamin:
closes #316 

This should still raise a useful error in the case that close to is _also_ zero since that is likely user error.

### DIFF
--- a/algosdk/constants.py
+++ b/algosdk/constants.py
@@ -89,6 +89,9 @@ LOGIC_SIG_MAX_COST = 20000
 LOGIC_SIG_MAX_SIZE = 1000
 """int: max size of a teal program and its arguments in bytes"""
 
+ZERO_ADDRESS = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ"
+"""str: algorand encoded address of 32 zero bytes"""
+
 # for backward compatibility:
 kmd_auth_header = KMD_AUTH_HEADER
 algod_auth_header = ALGOD_AUTH_HEADER
@@ -127,3 +130,4 @@ tx_group_limit = TX_GROUP_LIMIT
 max_asset_decimals = MAX_ASSET_DECIMALS
 logic_sig_max_cost = LOGIC_SIG_MAX_COST
 logic_sig_max_size = LOGIC_SIG_MAX_SIZE
+zero_address = ZERO_ADDRESS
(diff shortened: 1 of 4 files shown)


algochoi (approved):
Looks fine to me

nit: not sure if `ZERO_ADDRESS` should live in `transaction.py` or `constants.py`. I don't have a strong opinion so I'll just leave it as a thought.

barnjamin:
>nit: not sure if ZERO_ADDRESS should live in transaction.py or constants.py. I don't have a strong opinion so I'll just leave it as a thought.

Good point, moved it.

jasonpaulos (commented):
This seems like a good idea but it does open up a new edge case: what if amount is > 0? In that case the amount gets send to the receiver (which is the zero address now), and the account balance minus that amount gets send to the close to address.

Perhaps the condition should be `close_remainder_to is not None and not amt`?

jasonpaulos:
To clarify, is this change meant to fix decoding transactions properly from a block, or to let users instantiate transactions in a more useful way?

barnjamin:
>This seems like a good idea but it does open up a new edge case: what if amount is > 0? In that case the amount gets sent to the receiver (which is the zero address now), and the account balance minus that amount gets sent to the close to address.

Good point, I guess your suggestion of checking amt >0 should also be implemented.

> To clarify, is this change meant to fix decoding transactions properly from a block, or to let users instantiate transactions in a more useful way?

In this case the bug report I got was when decoding a transaction included in a block where apparently no `rcv` was set. Locally they can just patch it to set the zero address if it isnt already present

jasonpaulos:
Thanks for addressing the concerns I raised offline. This looks good, but it would be amazing if you could also add a test to `test_unit.py` that makes sure calling `future_msgpack_decode` on a txn where the receiver is the zero address works properly now

jasonpaulos (approved):
Looks great, thanks

PR #320, opened 2022-04-25
fixed typo in lsig trace

barnjamin:
🤦


michaeldiamant (approved):
@barnjamin Looks good - I made the following checks finding no other changes to suggest:
* Review all `DryrunTransactionResult.trace` invocations.
* Search repo for other _spaces_ references.

Aside:  Once we upgrade the repo past Python 3.5, we can install mypy.  I'm still a Python novice, but from what I understand, I think mypy catches the class of error.

PR #321, opened 2022-04-26
adding foreign-app-addr to dryrun creator

barnjamin:
We should include the app address of foreign apps


michaeldiamant (approved):
@barnjamin Based on my (limited) understanding, looks right.  I visually confirmed accounts are de-duplicated before usage.

PR #336, opened 2022-05-11
Generate updated client API code

algoidurovic:
Depends on: https://github.com/algorand/go-algorand/pull/3957


jasonpaulos (commented):
Looks good to me, but I expect the blackbox testing code that @tzaffi has added also needs to be updated because of this change. It seems to reference cost in many places.

tzaffi:
Perhaps `cost` is a synonym to `budget-debit`? In that case I would ask that `cost` be kept as an additional redundant field (which keeps the same behavior as currently), but that an issue be created in this repo to mark `cost` as deprecated. When this issue is created, I'll create an issues in the graviton and pyteal repo to stop using `cost` and which block the issue in this repo.

If I'm wrong in my assumption, then I'll need to understand a bit better the relationship between `cost` and the new fields.

tzaffi (approved):
As the repo has no existing tests that validate `cost` in dry-run, I think it's ok to accept as is. @barnjamin may want to have a look as well.

jasonpaulos (approved):

PR #337, opened 2022-05-19
starting to add box support

barnjamin:


barnjamin:
Will be completed in #338

PR #338, opened 2022-05-20
Box support for Application calls and Atomic Transaction Composer

algochoi:


algosdk/atomic_transaction_composer.py
```diff
@@ -1,6 +1,6 @@
-from abc import ABC, abstractmethod
```
algochoi, line 1: I used VSCode to sort some Python imports so there are some diffs at the top of these files.

run_integration.sh
```diff
@@ -7,7 +7,8 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+# TODO: Before merging, change branch back to master
```
algochoi, line 10: TODO: Revert this back to master before merging

algosdk/future/transaction.py
```diff
@@ -1650,23 +1651,24 @@ def teal_bytes(teal):
         ), "Program {} is not bytes".format(teal)
         return teal
 
+    @staticmethod
```
algochoi, line 1654: I'm not sure if this should be a static method on the `ApplicationCallTxn` class, but I kept it anyway (path of least resistance).

algosdk/box_reference.py
```diff
@@ -0,0 +1,64 @@
...
+        if self.app_index:
+            d["i"] = self.app_index
+        if self.name:
+            d["n"] = self.name
+        od = OrderedDict(sorted(d.items()))
+        return od
+
+    @staticmethod
+    def undictify(d):
+        args = {
```
algochoi, line 55: TODO: Need to untranslate here

algosdk/atomic_transaction_composer.py
```diff
@@ -1,6 +1,6 @@
...
 import base64
 import copy
+from abc import ABC, abstractmethod
```
tzaffi, line 3:
```suggestion
```
ghost import?

algochoi: I think the `TransactionSigner` currently uses the `abc`

https://github.com/algorand/py-algorand-sdk/blob/ae3fba9af8a83199b3faa649f7aa80fdcdb122fa/algosdk/atomic_transaction_composer.py#L561

algosdk/error.py
```diff
@@ -222,3 +222,8 @@ def __init__(self, msg):
 class AtomicTransactionComposerError(Exception):
     def __init__(self, msg):
         super().__init__(msg)
+
+
+class InvalidForeignAppIdError(Exception):
```
tzaffi, line 227: are you planning to also have `InvalidForeignAssetError` and similarly for **app accounts** and **boxes** ?

algosdk/future/transaction.py
tzaffi: Maybe this can just be a standalone function in `algosdk/encoding.py` ?

algosdk/future/transaction.py
```diff
@@ -1650,23 +1651,24 @@ def teal_bytes(teal):
...
+    def as_bytes(e):
+        """Confirm or coerce element to bytes."""
+        if isinstance(e, (bytes, bytearray)):
+            return e
+        if isinstance(e, str):
+            return e.encode()
+        if isinstance(e, int):
+            # Uses 8 bytes, big endian to match TEAL's btoi
+            return e.to_bytes(8, "big")  # raises for negative or too big
+        assert False, "{} is not bytes, str, or int".format(e)
```
tzaffi, line 1664: Is there a more informative error that can be raised as apposed to an `AssertionError` ?

algosdk/atomic_transaction_composer.py
tzaffi: ohhh.. the "-" part of the diff was right here staring at me

algosdk/box_reference.py
```diff
@@ -0,0 +1,92 @@
...
+        except ValueError:
+            # Check if the app referenced is itself after checking the
+            # foreign apps array (in case its own app id is in its own
+            # foreign apps array).
+            if ref_id == 0 or ref_id == this_app_id:
+                pass
+            else:
+                raise error.InvalidForeignAppIdError(
+                    f"Box ref with appId {ref_id} not in foreign-apps"
+                )
```
tzaffi, line 43:
```suggestion
            if ref_id and ref_id != this_app_id:
                raise error.InvalidForeignAppIdError(
                    f"Box ref with appId {ref_id} not in foreign-apps"
                )
```
nit: can remove an `else`

algosdk/box_reference.py
```diff
@@ -0,0 +1,92 @@
...
+        self.app_index = app_index
+        self.name = name
+
+    @staticmethod
+    def translate_box_reference(
+        ref: Tuple[int, bytes],
+        foreign_apps: List[int],
+        this_app_id: int,
+    ) -> "BoxReference":
+        # Try coercing reference id and name.
```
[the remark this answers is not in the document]
tzaffi: (I mean at the very end of processing in this method)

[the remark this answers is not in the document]
tzaffi: I really should have tagged the constructor `__init__()` with this comment... so really the question is: are we enforcing non-zero `app_index` and non-empty `name`

algosdk/box_reference.py
```diff
@@ -0,0 +1,92 @@
...
+
+        return [
+            BoxReference.translate_box_reference(
+                ref, foreign_apps, this_app_id
+            )
+            for ref in references
+        ]
+
+    def dictify(self):
+        d = dict()
```
tzaffi, line 73: if truthiness of `app_index` and `name` are enforced in `__init__()`, there will be no need to have this checks.

algosdk/box_reference.py
```diff
@@ -0,0 +1,92 @@
...
+        od = OrderedDict(sorted(d.items()))
+        return od
+
+    @staticmethod
+    def undictify(d):
+        args = {
+            "app_index": d["i"] if "i" in d else None,
+            "name": d["n"] if "n" in d else None,
+        }
+        return args
```
tzaffi, line 87:
```suggestion
        return BoxReference(**args)
```
Is this the intent?

tzaffi: I see now that usage of `undictify()` in the algosdk is inconsistent. I would argue that how `StateSchema` approaches this is an outlier. I would even recommend changing it's approach in this PR, or I can create a separate PR if you agree.

test/steps/v2_steps.py
```diff
@@ -36,17 +26,24 @@
...
+    register_type,  # pylint: disable=no-name-in-module
+    step,
+    then,
+    when,
+)
+from glom import glom
 
-from test.steps.steps import token as daemon_token
-from test.steps.steps import algod_port
+# TODO: This file is WAY TOO BIG. Break it up into logically related chunks.
```
tzaffi, line 46: I can take this on as a follow up if you have the time to review it!

algosdk/box_reference.py
algochoi: 👍

algosdk/future/transaction.py
algochoi: How about `TypeError` with that message?

test/steps/v2_steps.py
algochoi: 🙌 sounds good!

algosdk/box_reference.py
algochoi: I think this has to do with the msgpack encoding as we omit empty/zeroed fields when sending them over the wire. JJ mentioned that Box names and app ids can be 0's (which interestingly means that you can specify a box that will have just empty fields in the msgpack 😱 )

algosdk/error.py
algochoi: I wasn't planning to, but maybe it would be better to generalize this into something like: `InvalidForeignIndexError`?

algosdk/box_reference.py
algochoi: Sure thing, feel free to PR and I can take a look!

algosdk/future/transaction.py
algochoi: 👍

algosdk/box_reference.py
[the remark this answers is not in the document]
tzaffi: thx for explaining

algosdk/box_reference.py
tzaffi: I just opened #340 to address this issue. If you prefer not cluttering `feature/boxes` with that PR, I can change the base to `develop`.

LMK

test/steps/v2_steps.py
tzaffi: Here's the PR. I'll publicize it to other interested parties in the morning.

algochoi: Awesome work, thanks for doing this

tzaffi:
@algochoi - please consider #344

tzaffi (approved):
I'm approving with the contingency that this shouldn't be merged until https://github.com/algorand/algorand-sdk-testing/pull/181 has been merged, and that right before the merge the sdk branch referenced in `run_integration.sh` should point to `feature/box-storage`

algosdk/box_reference.py
```diff
@@ -0,0 +1,91 @@
...
+        self.name = name
+
+    @staticmethod
+    def translate_box_reference(
+        ref: Tuple[int, bytes],
+        foreign_apps: List[int],
+        this_app_id: int,
+    ) -> "BoxReference":
+        # Try coercing reference id and name.
+        ref_id, ref_name = int(ref[0]), encoding.encode_as_bytes(ref[1])
```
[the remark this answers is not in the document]
tzaffi: Similar question for `ref[1]`. If it's already `bytes` then encoding would be a no-op and therefore unnecessary.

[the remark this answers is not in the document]
tzaffi: This method is called `translate_box_reference()` and that's what it actually does. As such, it can handle a more general types than `Tuple[int, bytes]` for `ref`. So I think we should specify these more general types. In particular `ref[1]` should be exactly the type that `encode_as_bytes` accepts. Currently it accepts `Any`, but in reality it accepts `Union[bytes, bytearray, str, int]`.  `ref[0]` is a bit trickier as we're expecting an `int` but if supplied `"42"` we can "deal with it". So maybe a compromise is the following type hint?

```python
ref: Tuple[int, `Union[bytes, bytearray, str, int]
```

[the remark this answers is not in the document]
tzaffi: Similarly, I would argue for modifying `translate_box_references()`'s type hint for `references`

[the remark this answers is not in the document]
algochoi: Sounds good, I added a guard instead of casting to `int` just to reduce number of edge cases with int conversion and added appropriated type hints

algosdk/future/transaction.py
```diff
@@ -1624,6 +1624,7 @@ def __init__(
         self.foreign_apps = self.int_list(foreign_apps)
         self.foreign_assets = self.int_list(foreign_assets)
         self.extra_pages = extra_pages
+        self.boxes = BoxReference.translate_box_references(boxes, self.foreign_apps, self.index)  # type: ignore
```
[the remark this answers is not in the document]
tzaffi: Now that I think of it, how are you running `mypy` on the repo? I get an error when I try:

```sh
❯ mypy future/transaction.py
py-algorand-sdk is not a valid Python package name
```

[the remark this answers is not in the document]
tzaffi: Ok, I have one way of running it, using modules:

```sh
> mypy -p algosdk.future.transaction
```

But then I don't get on error on this line, even after I remove the `# type: ignore`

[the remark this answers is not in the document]
algochoi: Oh it showed up on my editor, and it looks like it is a "type issue" reported by [Pylance](https://github.com/microsoft/pylance-release)

PR #340, opened 2022-06-02
Make `StateSchema.undictify()` return an object

tzaffi:
Currently `StateSchema.undictify()` is the only such method in the repo (out of 28 definitions) which does _not_ return an AlgoSDK object.

This PR addresses this discrepancy.

### DIFF
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -1507,11 +1507,10 @@ class StateSchema:
 
     @staticmethod
     def undictify(d):
-        args = {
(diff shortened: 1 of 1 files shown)


algochoi (approved):
LGTM ! I think we can merge after tests pass and I'll make similar changes in the open PR

tzaffi:
Good point! I'll create a PR into `feature/box-storage` to modify the changelong with this information

tzaffi:
@algochoi - please consider #343

PR #345, opened 2022-06-06
Add method to get Method by name

barnjamin:
I expect a common use case when preparing an ATC group will be to grab a Method from some parsed json file. 

Rather than forcing the dev to maintain a separate list we should offer some utility method to get the Method by name.


jasonpaulos:
This feature makes sense, but I think it would be more useful as a method on the `Contract` & `Interface` classes, so you can just do `my_contract.get_method_by_name("...")`.

It looks like you've properly handled the issue where two methods have the same name, which was my primary concern.

All the SDKs would benefit from a method like this.

algosdk/abi/contract.py
```diff
@@ -63,6 +63,26 @@ def undictify(d: dict) -> "Contract":
...
+
+        if len(methods_filtered) > 1:
+            raise KeyError(
+                "Found {} methods with the same name {}".format(
+                    len(methods_filtered),
+                    ",".join(
+                        [method.get_signature() for method in methods_filtered]
+                    ),
+                )
+            )
```
ahangsu, line 79: so this does not allow for contract with two methods of same name, but type-signatures are different? I mean `add1(uint32)uint32` and `add1(uint16)uint16`?

Do you think it would be more helpful if we want to return a list of `methods` or method-signatures, rather than just one method?

ahangsu (commented):
just one comment/question about the return of `get_method_by_name`, the rest is fine

algosdk/abi/contract.py
ahangsu: ok, that makes sense

algosdk/abi/contract.py
```diff
@@ -63,6 +63,26 @@ def undictify(d: dict) -> "Contract":
             name=name, desc=desc, networks=networks, methods=method_list
         )
 
+    def get_method_by_name(self, name: str) -> Method:
```
tzaffi, line 66: Just linking to https://github.com/algorand/algorand-sdk-testing/pull/185/files# there is nothing to do here which has a resolved convo explaining why this interface is "good nuf"

algosdk/abi/interface.py
```diff
@@ -49,3 +49,23 @@ def undictify(d: dict) -> "Interface":
         method_list = [Method.undictify(method) for method in d["methods"]]
         desc = d["desc"] if "desc" in d else None
         return Interface(name=name, desc=desc, methods=method_list)
+
+    def get_method_by_name(self, name: str) -> Method:
+        methods_filtered = list(
```
tzaffi, line 54: similar bias here.... general question, and not for this PR. Seems like there's a lot of overlap between contract and interface. Having a common abstract supertype might make sense.

test/steps/application_v2_steps.py
```diff
@@ -992,3 +992,58 @@ def spin_results_satisfy(context, result_index, regex):
...
 
     assert re.search(regex, spin), f"{spin} did not match the regex {regex}"
+
+
+@when(
+    'I create another Method object from method signature "{extramethod:MaybeString}"'
+)
+def make_extra_method(context, extramethod):
+    if extramethod != "":
+        context.extramethod = abi.Method.from_signature(extramethod)
```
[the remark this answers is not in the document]
barnjamin: I think the suggestions in the sdk testing repo are good, and i've applied them and accounted for them here

algosdk/abi/interface.py
jasonpaulos: I think a change like this is a little outside of the scope of this PR, but it is something we should consider.

I definitely dislike how much code is repeated between the two methods though. Perhaps a short term solution is to define a `get_method_by_name(methods: list[Method], name: str) -> Method` helper function which implements the logic, then call into it from both `Interface.get_method_by_name` and `Contract.get_method_by_name`? I think all SDKs would benefit from this @barnjamin

barnjamin: If itll be abstracted in another PR anyway, I'm inclined to leave it as is for now unless it makes you want to puke

jasonpaulos: I've not thought through all the implications of making `Contract` inherit from `Interface`, so I'm not sure if/when we'd want to make that change. In an effort to not hold up these PRs because of that, I think they should be mergable on their own, which means trying to avoid duplicate code.

From a maintenance perspective there's way fewer opportunities for problems if the function is implemented in just one place.

barnjamin: Makes sense, just added it

tzaffi (commented):
Has testing been conducted using these steps against the scenarios in the accompanying [SDK testing PR](https://github.com/algorand/algorand-sdk-testing/pull/185)?

It's possible to merge the other one first, assuming [the suggestion](https://github.com/algorand/algorand-sdk-testing/pull/185#pullrequestreview-1001416152) to add a special tag to the scenarios such as `@abi.contract` is taken.

barnjamin:
>Has testing been conducted using these steps against the scenarios in the accompanying https://github.com/algorand/algorand-sdk-testing/pull/185?

Yes, the green check on `cdd0b8d` still had the sdk tests but I can add the special tag if that makes it easier to see

tzaffi:
> > Has testing been conducted using these steps against the scenarios in the accompanying [algorand/algorand-sdk-testing#185](https://github.com/algorand/algorand-sdk-testing/pull/185)?
> 
> Yes, the green check on `cdd0b8d` still had the sdk tests but I can add the special tag if that makes it easier to see

Got it. So since the SDK Testing PR as 3 approvals, once you're confident that the tag trick works, it would be great if you merged it in. It's annoying to have perfectly fine PR's lingering unmerged. The downside is that it's causing a proliferation of cucumber tags, but that can easily be cleaned up at a later date.

tzaffi (approved):
LGTM (🤞 on the tests passing)

PR #347, opened 2022-06-14
Implement new step asserting that AtomicTransactionComposer's attempt to add a method can fail with a particular error

tzaffi:
# Implementing the following step:

>  I add a method call with the signing account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments; any resulting exception has key "{none-or-exception-key}".

## Related Issues and PR's
* https://github.com/algorand/algorand-sdk-testing/pull/190
* https://github.com/algorand/algorand-sdk-testing/pull/189

### DIFF
--- a/run_integration.sh
+++ b/run_integration.sh
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch abi-too-many-args https://github.com/algorand/algorand-sdk-testing.git test-harness
 
 ## Copy feature files into the project resources
 mkdir -p test/features
--- a/run_integration.sh
+++ b/run_integration.sh
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch abi-too-many-args https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
 
 ## Copy feature files into the project resources
 mkdir -p test/features
--- a/test/steps/application_v2_steps.py
+++ b/test/steps/application_v2_steps.py
@@ -610,6 +610,7 @@ def abi_method_adder(
     local_ints=None,
     extra_pages=None,
     force_unique_transactions=False,
+    exception_regex="none",
 ):
     if account_type == "transient":
         sender = context.transient_pk
@@ -656,7 +657,6 @@ def abi_method_adder(
             + context.nonce.encode()
         )
 
-    context.most_recent_added_method_exception = None
     try:
         context.atomic_transaction_composer.add_method_call(
             app_id=app_id,
@@ -674,31 +674,26 @@ def abi_method_adder(
             note=note,
         )
     except AtomicTransactionComposerError as atce:
-        context.most_recent_added_method_exception = atce
+        assert (
+            exception_regex != "none"
+        ), f"cucumber step asserted that no exception resulted, but the following exception actually occurred: {atce}"
 
-
-@then(
-    'the most recently added method call has an exception which satisfies "{regex}".'
-)
-def most_recently_added_method_exception_satisfies(context, regex):
-    most_recent_exception = context.most_recent_added_method_exception
-    if regex == "none":
-        assert most_recent_exception is None
-        return
-    assert re.search(
-        regex, str(most_recent_exception)
-    ), f"{most_recent_exception} did not satisfy {regex}. FYI: method_args={context.method_args}; app_args={context.app_args}"
+        assert re.search(
+            exception_regex, str(atce)
+        ), f"{atce} did not satisfy the expected regular expression {exception_regex}"
 
 
 @step(
-    'I add a nonced method call with the {account_type} account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments.'
+    'I add a method call with the {account_type} account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments; any resulting exception satisfies the regex "{exception_regex}".'
 )
-def add_abi_method_call_nonced(context, account_type, operation):
+def add_abi_method_call_with_exception(
+    context, account_type, operation, exception_regex
+):
     abi_method_adder(
         context,
         account_type,
         operation,
-        force_unique_transactions=True,
+        exception_regex=exception_regex,
     )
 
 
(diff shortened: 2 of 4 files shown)


test/steps/application_v2_steps.py
```diff
@@ -621,28 +648,46 @@ def int_if_given(given):
...
+            clear_program=clear_program,
+            extra_pages=extra_pages,
+            note=note,
+        )
+    except AtomicTransactionComposerError as atce:
+        context.most_recent_added_method_exception = atce
+
+
+@then(
+    'the most recently added method call has an exception which satisfies "{regex}".'
```
jasonpaulos, line 681: As a design note, generally we want to add error checking in the step that causes the error, like the step `I sign and submit the transaction, saving the txid. If there is an error it is "{error_string:MaybeString}".`

The reason being if there is a separate step to check for an error, it's all too easy to forget to check this step, meaning you will be silently ignoring potential errors.

jasonpaulos: I think it makes sense to add more `I add a method call...` variants with a new `If there is an error it is...` suffix. That way existing `I add a method call...` usage does not need to change, there is just now a new version that should be used when you want to check for a specific error.

I believe all SDKs already use a common helper function to implement the 4 versions of this step, so adding potentially 4 more doesn't seem like a huge ask.

Is it weird to have 4+ variants of this step? Yes, but I think that's just the compromise we have to make when using cucumber.

run_integration.sh
```diff
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch abi-too-many-args https://github.com/algorand/algorand-sdk-testing.git test-harness
```
tzaffi, line 10: TODO: revert to point back to `master` before merging this PR.

test/steps/application_v2_steps.py
```diff
@@ -159,8 +179,13 @@ def lookup_application_include_all2(
         context.response = json.loads(str(e))
 
 
+@when("we make a LookupApplications call with applicationID {app_id}")
```
tzaffi, line 182: moved closer to similar function

test/steps/application_v2_steps.py
```diff
@@ -159,8 +179,13 @@ def lookup_application_include_all2(
...
+def lookup_application(context, app_id):
+    context.response = context.icl.applications(int(app_id))
+
+
 @when("I use {indexer} to lookup application with {application_id}")
-def lookup_application(context, indexer, application_id):
+def lookup_application2(context, indexer, application_id):
```
tzaffi, line 188: renamed to avoid method name collision

test/steps/application_v2_steps.py
```diff
@@ -512,35 +537,37 @@ def add_transaction_to_composer(context):
     )
 
 
-def process_abi_args(context, method, arg_tokens):
+def process_abi_args(context, method, arg_tokens, erase_empty_tokens=True):
+    if erase_empty_tokens:
```
tzaffi, line 541: Now that we no longer assume that the lengths of the method args and actual args are the same, a couple of side effects occurred. One side effect is that when `"".split()` is called on an empty string, you get a non-empty list. As in all existing cases, empty strings that are split meant to indicate no values at all, this change shouldn't break any tests.

test/steps/other_v2_steps.py
```diff
@@ -1406,27 +1406,6 @@ def algod_v2_client(context):
     context.app_acl = algod.AlgodClient(daemon_token, algod_address)
 
 
-@step(
-    'I sign and submit the transaction, saving the txid. If there is an error it is "{error_string:MaybeString}".'
```
tzaffi, line 1410: moved this to `application_v2_steps.py`

test/steps/application_v2_steps.py
jasonpaulos: I'm a little confused. Is the cause of this change that the step `I append the encoded arguments "<app-args>" to the method arguments array` now has to work when `<app-args>` is the empty string?

Or is there another cause?

tzaffi: Previously, all tests had exactly the same number of runtime args as defined in the method signature. The step logic iterated through the _method signature_ args and accessed the run-time arg with index corresponding to the signature arg. When there were no _method signature_ args, the loop was actually trivial (see the image below). However, in that trivial case, `args` was actually the list of length 1 containing an empty string `['']` (because of how python split works):
```python
>>> "".split(",")
['']
```
Now that we are no longer assuming that the number of args are the same, the 0 vs. 1 case no longer was correct so I needed to adjust the logic using this hack.

image

tzaffi: So yes, it has to work with empty string (which before was actually ignored in a sense).

jasonpaulos: Thanks for that explanation, I think I understand now. Alternatively, would it be an acceptable solution to modify the `I append the encoded arguments "<app-args>" to the method arguments array` step to check if `<app-args>` is an empty string and do nothing in that case?

The reason I suggest this is because we can semantically still pass empty string args with something like `a,,b`, and this would be ignored by the current code, but I think it should complain/error for this input.

tzaffi: yeah, you'd have to set `erase_empty_tokens` to false... still thinking about your first paragraph

tzaffi: I have a proposal. Rename this step to:

```
'I add a method call with the {account_type} account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments ignoring empty ones; any resulting exception has key "{exception_key}".'
```
the phrase "ignoring empty ones" indicates that we should not be allowing empty arguments for this step. And in fact, this step will eventually call `process_abi_args(... erase_empty_tokens=True)` while all the other steps will continue as before and therefor have `erase_empty_tokens=False`

Sounds good?

jasonpaulos: That's better, but it still looks like we're addressing the symptoms of the problem, not the root cause. I'd find it simpler if no filtering of arguments needed to be done after the fact, since this seems error-prone and hard to reason about.

tzaffi: I'm trying your original suggestion with `I append the encoded arguments`. Seems to work. We'll see if the tests pass.

run_integration.sh
tzaffi, line 10:
```suggestion
git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
```

jasonpaulos (approved):
Looks good

PR #348, opened 2022-06-15
Boxes: Add support for Boxes

algochoi:
Adds SDK support for Boxes introduced in https://github.com/algorand/indexer/pull/1168 and https://github.com/algorand/go-algorand/pull/4149.


run_integration.sh
```diff
@@ -7,7 +7,8 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+# TODO: Before merging, change branch back to master
+git clone --single-branch --branch feature/box-storage https://github.com/algorand/algorand-sdk-testing.git test-harness
```
algochoi, line 11: TODO: Change this branch back to master after merging in the Algorand SDK Testing changes

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+# BOX-STORAGE-VERSION TBD
+
+## Changed
+
+- `class StateSchema`'s method `undictify()` now returns a `StateSchema` object instead of a python `dict`
```
algochoi, line 7: It looks like this file might be autogenerated and might cause conflicts - just a note to take a look later

algosdk/v2client/algod.py
```diff
@@ -126,6 +126,23 @@ def application_info(self, application_id, **kwargs):
...
+
+        Args:
+            application_id (int): The ID of the application to look up.
+            box_name (bytes): The name or key of the box.
+        """
+        encoded_box = base64.b64encode(box_name).decode()
+        box_name_encoded = "b64:" + encoded_box
+        req = "/applications/" + str(application_id) + "/box"
+        params = {"name": box_name_encoded}
+        print(req, params)
```
algochoi, line 143: I'll delete this in the next commit

.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="feature/box-storage"
```
ahangsu, line 3: TODO: when algorand-sdk-testing PR merged:
```suggestion
SDK_TESTING_BRANCH="master"
```

algosdk/v2client/algod.py
```diff
@@ -126,6 +126,35 @@ def application_info(self, application_id, **kwargs):
...
+        return self.algod_request("GET", req, params=params, **kwargs)
+
+    def application_boxes(self, application_id: int, limit: int = 0, **kwargs):
+        """
+        Return a list of all the application's boxes.
+
+        Args:
+            application_id (int): The ID of the application to look up.
+            limit (int, optional): Max number of box names to return.
+                If max is not set, or max == 0, returns all box-names.
```
tzaffi, line 152:
```suggestion
                If max is not set, or max == 0, returns as many box-names as configured by the algod server being queried.
```

tests/steps/application_v2_steps.py
```diff
@@ -40,24 +43,49 @@ def operation_string_to_enum(operation):
         )
 
 
+# Takes in a tuple where first element is the encoding and second element is value.
+# If there is only one element, then it is assumed to be an int.
+def process_app_args(sub_arg):
```
tzaffi, line 48:
```suggestion
def process_app_arg(sub_arg):
```
Should this be renamed in the singular without an "s" because _if I understand correctly_ only one arg is being processed?

tests/steps/application_v2_steps.py
```diff
@@ -40,24 +43,49 @@ def operation_string_to_enum(operation):
...
-            app_args.append(int(sub_arg[0]))
-        elif sub_arg[0] == "str":
-            app_args.append(bytes(sub_arg[1], "ascii"))
-        elif sub_arg[0] == "b64":
-            app_args.append(base64.decodebytes(sub_arg[1].encode()))
-        elif sub_arg[0] == "int":
-            app_args.append(int(sub_arg[1]))
-        elif sub_arg[0] == "addr":
-            app_args.append(encoding.decode_address(sub_arg[1]))
+        app_args.append(process_app_args(sub_arg))
```
tzaffi, line 68: nice refactor!

tests/steps/application_v2_steps.py
```diff
@@ -40,24 +43,49 @@ def operation_string_to_enum(operation):
...
+    # Box strings alternate between the app ID and the encoded app arg.
+    for token in split_args:
+        try:
+            app_id = int(token)
+        except ValueError:
+            sub_arg = token.split(":")
+            sub_arg = process_app_args(sub_arg)
+            boxes.append((app_id, sub_arg))
+    # Sanity check that input correctly alternates between int and str.
+    assert len(boxes) == len(split_args) // 2
```
tzaffi, line 85: I'm looking for an example test case [here](https://github.com/algorand/algorand-sdk-testing/pull/192/files) where there would be more than 2 tokens, but not finding any (please do point out if I missed it). Is this meant to handle more generalized cases? Should we possibly have added such an example to our tests?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+
```
tzaffi, line 8:
```suggestion
- `AtomicTransactionComposer`'s `add_method_call()` and `future.transaction`'s `ApplicationCallTxn()` accept `boxes` box reference params
- Introducing `BoxReference` class received by `future.transaction`'s `ApplicationCallTxn()`
- Adding `application_box_by_name()` and `application_boxes()` to `AlgodClient`
- Adding `application_box_by_name()` and `application_boxes()` to `IndexerClient`
```

tests/steps/application_v2_steps.py
```diff
@@ -344,14 +419,19 @@ def remember_app_id(context):
     context.app_ids.append(app_id)
 
 
-def wait_for_algod_transaction_processing_to_complete():
+@then(
+    "I sleep for {millisecond_num} milliseconds for indexer to digest things down."
+)
+def wait_for_algod_transaction_processing_to_complete(
```
tzaffi, line 425: Is this being used as both a cucumber step function as well as called directly as a helper function by other steps? If so, I recommend refactoring so that the helper function remains as such (not a step), while the new step immediately calls the helper. Optionally, consider moving the helper near the top of the file.

tzaffi (commented):
Some minor suggestions, including more line items to the CHANGELOG.md

tests/steps/application_v2_steps.py
algochoi: I think this suggestion makes sense, and I will refactor this part with an updated comment.

algochoi: Addressed in https://github.com/algorand/py-algorand-sdk/pull/348/commits/35a1c43bc844a8fdbb7817e1252b2c405f40e102 - let me know how this looks

tests/steps/application_v2_steps.py
```diff
@@ -134,8 +134,8 @@ def wait_for_transaction_processing_to_complete_in_dev_mode(
 @then(
     "I sleep for {millisecond_num} milliseconds for indexer to digest things down."
 )
-def wait_for_indexer_in_dev_mode():
-    wait_for_transaction_processing_to_complete_in_dev_mode()
+def wait_for_indexer_in_dev_mode(millisecond_num):
```
ahangsu, line 137: nit: do you need to pass `context` in?

tzaffi (approved):
Approving as the only remaining discussion is around the CHANGELOG

ahangsu (approved):
my approval left for the part I didn't write, i.e. mostly outside `tests`

algosdk/v2client/algod.py
```diff
@@ -126,6 +126,35 @@ def application_info(self, application_id, **kwargs):
         req = "/applications/" + str(application_id)
         return self.algod_request("GET", req, **kwargs)
 
+    def application_box_by_name(
+        self, application_id: int, box_name: bytes, **kwargs
+    ):
+        """
+        Return the value of an application's box.
```
jasonpaulos, line 133: This endpoint (and the other 3) returns a dictionary containing base64 encoded values. This behavior is likely going to confuse consumers.

At the bare minimum, we should better document what the response looks like and perhaps even provide an example.

Though I'm debating whether we should actually process the response and decode the values into bytes, similar to what the JS SDK now does.

algosdk/box_reference.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.name:
+            d["n"] = self.name
+        od = OrderedDict(sorted(d.items()))
+        return od
+
+    @staticmethod
+    def undictify(d):
+        return BoxReference(
+            d["i"] if "i" in d else None,
+            d["n"] if "n" in d else None,
```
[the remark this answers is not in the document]
ahangsu: this give rise to a further question, do we want to use mypy to enforce the type checking in py-algorand-sdk? Since I ran mypy against sdk, and a few error occurred.

Zooming back to this particular case, I think the suggestion makes sense. Waiting on other folks for consensus.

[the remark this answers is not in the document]
ahangsu: Actually I can see the argument to mark type optional both ways, I just prefer to mark then to be hard required than optional.

algosdk/v2client/algod.py
jasonpaulos: After discussion with @winder and @michaeldiamant, we've decided to leave the behavior of these endpoints as-is for this SDK, while at the same time retaining the new behavior in the JS SDK.

Some of the factors that went into this decision were:
* The Python SDK lacks any code generation for response models. Returning a typed response object first involves creating a definition for that object, which is a considerable undertaking, especially if we need to maintain these definitions by hand.
* Type annotations are sparse in this SDK, and in fact it does not run `mypy` or a similar static analyzer. This diminishes the potential advantages of having typed response objects.
* Currently the Python SDK has a high level of consistency and simplicity, especially with response handling. A deviation for these endpoints would reduce those properties, and therefor reduce the maintainability of the code.

Generally speaking we are interesting in improving the ergonomics of this SDK, but at this time the effort needed is too high.

algosdk/box_reference.py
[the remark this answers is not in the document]
jasonpaulos: I opened #388 to resolve

algosdk/v2client/algod.py
jasonpaulos: I opened #388 to resolve this by mentioning box names/values will be base64 encoded in responses

jasonpaulos (approved):

PR #350, opened 2022-06-15
Refactor unit tests into separate files

algochoi:
This PR separates the different unit test classes into different files for improved readability. 

This PR also proposes to rename the `test` directory to `tests` directory to prevent any conflicts with the existing Python [`test`](https://docs.python.org/3/library/test.html) package. 

Currently merging into `feature/box-storage` in fear of merge conflicts.


.gitignore
```diff
@@ -114,7 +114,7 @@ venv.bak/
```
michaeldiamant, line 114: @algochoi Unrelated to the line, but opening a thread to minimize _Conversation_ chatter:

* Thanks for taking the exercise of splitting out tests from a single file.  
* Open to discussion - What are your thoughts?  
  * I think the chance for error will be higher by refactoring in the (long-lived) feature branch rather than `develop`. 
  * My 2c is to make the change against `develop` and once approved, then bring the change into the feature branch.  Using the feature branch requires us to repeat the refactoring for any test changed in `develop`.

algochoi: Merging to `develop` sounds fine to me. I'm hoping that the merge-back process will be smooth for other PRs touching the test files as well.

algochoi: Opening up https://github.com/algorand/py-algorand-sdk/pull/351 to PR against `develop`. There are two fewer test classes in that PR compared to this one.

PR #352, opened 2022-06-18
Control my test branches

tzaffi:
# Allowing easier control of which repos and branches are used for running tests

In theory, almost all the configs here are already possible to do. However they are spread in a couple of different files, and in the case of indexer, you need to modify the sdk-testing's docker file in order to run a different branch.

The PR proposes to unify the approach and expose all the git info in the `Makefile`.  In particular, a new target `build-for-local-test` is introduced that can be configured to pull any remote/branch of

* https://github.com/algorand/algorand-sdk-testing
* https://github.com/algorand/go-algorand
* https://github.com/algorand/indexer

**NOTE:** as opposed to `make docker-test`, `make build-for-local-test` does _not_ build a docker image off of the local directory, and does _not_ run any tests. Once the build completes, you can run integration tests as usual with `make integration`.

## Dependant Branches

SDK testing's [control-my-test-branches](https://github.com/algorand/algorand-sdk-testing/tree/control-my-test-branches) - the change there allows specifiing indexer's url and branch from outside the `docker-compose.yml`

## Future work

* Can we do this all using sandbox instead of building from scratch?
* What do we do about all the very old indexers that hardcoded to a hash that's from longer than a year ago?


run_integration.sh
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env bash
```
tzaffi, line 1: this was renamed `integration.sh` and unfortunately the diffs were lost

integration.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
tzaffi, line 1: formerly `run_integration.sh`. Some extra logic has been added to:
1. clone the sdk repo as configured by `$TESTING_BRANCH` + `$TESTING_URL`
2. only dockerize the python sdk when `$DOCKERIZE_PYSDK==true`
3. pass the `UP_ENV` file directly to `up.sh`
4. only run all the tests via docker when `$DOCKERIZE_PYSDK==true`

tests/steps/application_v2_steps.py
```diff
@@ -7,7 +7,11 @@
 
 from algosdk import abi, atomic_transaction_composer, encoding, mnemonic
 from algosdk.abi.contract import NetworkInfo
-from algosdk.error import ABITypeError, IndexerHTTPError
+from algosdk.error import (
```
tzaffi, line 10: this diff should go away after `refactor-unit-tests` merges in the latest on `develop`

tests/steps/other_v2_steps.py
```diff
@@ -603,7 +603,7 @@ def txns_by_addr(
 @when(
     'we make a Lookup Account Transactions call against account "{account:MaybeString}" with NotePrefix "{notePrefixB64:MaybeString}" TxType "{txType:MaybeString}" SigType "{sigType:MaybeString}" txid "{txid:MaybeString}" round {block} minRound {minRound} maxRound {maxRound} limit {limit} beforeTime "{beforeTime:MaybeString}" afterTime "{afterTime:MaybeString}" currencyGreaterThan {currencyGreaterThan} currencyLessThan {currencyLessThan} assetIndex {index}'
 )
-def txns_by_addr(
+def txns_by_addr2(
```
tzaffi, line 606: this diff should go away after `refactor-unit-tests` merges in the latest on `develop`

Makefile
```diff
@@ -7,5 +7,37 @@ integration:
 	behave --tags=$(INTEGRATIONS) tests -f progress2
 
 PYTHON_VERSION ?= 3.8
+docker-test: export PYTHON_VERSION := $(PYTHON_VERSION)
+docker-test: export DOCKERIZE_PYSDK ?= true
+docker-test: export TESTING_URL ?= https://github.com/algorand/algorand-sdk-testing.git
+docker-test: export TESTING_BRANCH ?= control-my-test-branches # revert to master before merge
```
tzaffi, line 13: reset before merging

Makefile
```diff
@@ -7,5 +7,37 @@ integration:
...
+	echo "contents of $(LOCAL_TESTENV)"
+	cat $(LOCAL_TESTENV) || echo "DNE"
+
+# configure the repos and branches that a local test will build off of:
+local-test: export TESTING_URL ?= https://github.com/algorand/algorand-sdk-testing.git
+local-test: export TESTING_BRANCH ?= control-my-test-branches # revert to master before merge
+local-test: export ALGOD_URL ?= https://github.com/algorand/go-algorand
+local-test: export ALGOD_BRANCH ?= feature/avm-boxes # revert to master before merge
+local-test: export INDEXER_URL ?= https://github.com/algorand/indexer
+local-test: export INDEXER_BRANCH ?= localledger/integration # revert to develop before merge
```
tzaffi, line 32: more reset before mergings here too

Makefile
```diff
@@ -7,5 +7,46 @@ integration:
...
+LOCAL_TESTENV ?= .local-env
+prep-local-testenv:
+	python tests/prep_local_testenv.py $(LOCAL_TESTENV)
+
+show-local-testenv:
+	echo "contents of $(LOCAL_TESTENV)"
+	cat $(LOCAL_TESTENV) || echo "DNE"
+
+# configure the repos and branches that a local test will build off of:
+build-for-local-test: export TESTING_URL ?= https://github.com/algorand/algorand-sdk-testing.git
```
tzaffi, line 36: @algoidurovic @algochoi - Do these changes make sense in addressing the branch control issues that you encountered last week?

algochoi: This is great - much easier to see when it's consolidated into one file. Will this override the up_env file in local tests?

tzaffi: yes - locally, this overrides Testing SDK's pulled in`.up_env`, while `make docker-test` still uses `.up_env`. 

This may be a new source of confusion. Some other approaches that could be taken:
* instead of having `prep-local-testenv` create a new file `.local-env` which supplants `.up_env`, have it overwrite `.up_env` (this seems a bit hacky and even more confusing)
* dispense with `prep-local-testenv` along with the python script and most of the `Makefile` exports by making permanent `.local-env`

The advantage of the current approach is that one can see all the repo url's and branches in the `Makefile`. But there's no rule that says that this is how things need to be.

tzaffi:
> These are some really nice improvements, it looks like they should really improve the testing cycle when working with Python.
> 
> There is [something similar in the Java SDK](https://github.com/algorand/java-algorand-sdk/blob/develop/run_integration_tests.sh). The main thing it has to offer is an option which just synchronizes the feature files.

RIght. So in the Java SDK, an alternate approach is taken: improve `run_integration.sh` to make it easy to use directly. Here, the Makefile becomes the main controller of the test configuration. I don't think we can ever achieve complete uniformity across the SDK's but it's preferrable to keep things as similar as possible.

Makefile
jdtzmn: I personally think the second bullet point is a cleaner solution than including exports in the Makefile. It also is more consistent with the way it is done in sandbox, which uses the `config.[network]` files. I'm curious what others think.

Makefile
```diff
@@ -7,5 +7,46 @@ integration:
...
+build-for-local-test: export INDEXER_BRANCH ?= localledger/integration # revert to develop before merge
+
+# other exports for `local-test`:
+build-for-local-test: export PYTHON_VERSION := $(PYTHON_VERSION)
+build-for-local-test: export LOCAL_TESTENV := $(abspath .)/$(LOCAL_TESTENV)
+build-for-local-test: export DOCKER_TESTING ?= false
+build-for-local-test: export TYPE_OVERRIDE ?= source
+
+# local-test bootstraps the necessary algod and indexer docker containers
+# for integration tests, but does not run the tests
```
jdtzmn, line 50: Consider amending this with "unless `DOCKER_TESTING` is true" to reduce potential confusion

jdtzmn (approved):
Looks great! It seems like this will greatly improve the testing experience! Just left one optional minor comment.

jdtzmn (approved):

tzaffi:
After some discussion with team members and @michaeldiamant we concluded that adding this layer of control inside the SDK, while convenient, is still an extra layer with its own pitfalls. All other things equal, more code is worse the less code.

PR #353, opened 2022-06-20
Dev Tools: Source map decoder

barnjamin:
The VLQ encoding is a lot to ask folks to implement themselves, lets provide _some_ easier way to get the numeric pc

Lifted from @algochoi https://github.com/algochoi/teal-sourcemap-decoder

@michaeldiamant I think this would need to be implemented across the SDKs assuming we can come up with a good way to expose it.

I'm imagining with the json src_map

```py
err_pc_num = 996 # really comes from error message returned by algod
sm = SourceMap(src_map)
print(f"Teal source line number: {sm.get_line_for_pc(err_pc_num)}")
```


algochoi:
Nice, I think this will work nicely when the APIs also support the disassemble endpoint [(related ticket)](https://github.com/algorand/algorand-sdk-testing/issues/170). We could add some cucumber tests in the testing repo as well to check the translation, but I'm not sure what would be the easiest way to test it (maybe output the VLQ->Int mapping in a unified JSON format?)

algosdk/logic.py
```diff
@@ -287,3 +289,82 @@ def get_application_address(appID: int) -> str:
     to_sign = constants.APPID_PREFIX + appID.to_bytes(8, "big")
     checksum = encoding.checksum(to_sign)
     return encoding.encode_address(checksum)
+
+
+class SourceMap:
```
michaeldiamant, line 294: @barnjamin Optional:  I can imagine adding more sourcemap-related facilities later on.  For example, assuming PyTeal-to-TEAL sourcemaps exist, we may wish to provide facilities for chaining sourcemaps.

I raise the note because I'm wondering:  What do you think about relocating the functionality to a new file?

I don't feel _too_ strongly though I can imagine `logic.py` growing to be large over time.

run_integration.sh
```diff
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch sourcemap https://github.com/algorand/algorand-sdk-testing.git test-harness
```
barnjamin, line 10: Will take out before merging, want to see it pass in CI

algosdk/source_map.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_line_for_pc(self, pc: int) -> int:
+        return self.pc_to_line[pc]
+
+    def get_pcs_for_line(self, line: int) -> List[int]:
+        return self.line_to_pc[line]
+
+
+def _decode_int_value(value: str) -> int:
+    decoded_value = base64vlq_decode(value)
+    return decoded_value[2] if decoded_value else None
```
algochoi, line 40: nit: adding a comment here would be helpful, e.g.

```
# Mappings may have up to 5 segments:
# Third segment represents the zero-based starting line in the original source represented.
```

tests/steps/other_v2_steps.py
```diff
@@ -1707,3 +1702,17 @@ def glom_app_eval_delta(context, i, path, field):
     assert field == str(
         actual_field
     ), f"path [{path}] expected value [{field}] but got [{actual_field}] instead"
+
+
+@given('a source map json file "{sourcemap_file}"')
+def step_impl(context, sourcemap_file):
```
algochoi, line 1708: nit: linter thing, rename `step_impl` here

algochoi (commented):
Looking good - are the integration tests running? The output looks a little empty 🤔

algosdk/logic.py
```diff
@@ -2,6 +2,8 @@
 import json
 import os
 
+from typing import Tuple, Dict, List, Any
```
algochoi, line 5: nit: I think this can be deleted

ahangsu (commented):
just took a quick pass, and I have a minor question: do we want to point the testing branch to `sourcemap`?

algosdk/logic.py
```diff
@@ -2,6 +2,8 @@
 import json
 import os
```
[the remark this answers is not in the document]
michaeldiamant: Resolving - Thanks for working towards consistency across SDKs.

algosdk/source_map.py
```diff
@@ -0,0 +1,97 @@
...
+
+_b64chars = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+_b64table = [None] * (max(_b64chars) + 1)
+for i, b in enumerate(_b64chars):
+    _b64table[b] = i
+
+_shiftsize, _flag, _mask = 5, 1 << 5, (1 << 5) - 1
+
+
+def base64vlq_decode(vlqval: str) -> Tuple[int]:
```
michaeldiamant, line 66: @barnjamin Minor:  Do you expect the VLQ decode + encode methods to be used _directly_ by SDK users?

My read is:  No - The methods are intended for internal API usage.  If that's the case, propose prefixing these methods with `_` to demarcate as private.
* `base64vlq_decode`
* `base64vlq_encode`

Aside:  If intended for internal usage, it might be possible to remove `base64vlq_encode`.  It's unused.

algosdk/source_map.py
```diff
@@ -0,0 +1,97 @@
...
+            for raw_val in self.mapping.split(delimiter)
+        ]
+
+        # Initialize with 0,0 for pc/line
+        self.pc_to_line: Dict[int, int] = {0: 0}
+        self.line_to_pc: Dict[int, List[int]] = {0: [0]}
+
+        last_line = 0
+        for index, line_num in enumerate(pc_list):
+            if line_num is not None:  # be careful for '0' checks!
```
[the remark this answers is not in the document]
barnjamin: No, the line_num may be None otherwise but I'm not sure exactly when. @algochoi ?

algosdk/source_map.py
barnjamin: I don't expect they'll be used outside the source map stuff, I'll make them private. 

I am a fan of deleting code but this is somewhat hairy to re-implement without a reference so in this case I'd be in favor of leaving it in unless you've got a strong opinion

algosdk/source_map.py
```diff
@@ -0,0 +1,97 @@
...
+    # Mappings may have up to 5 segments:
+    # Third segment represents the zero-based starting line in the original source represented.
+    decoded_value = base64vlq_decode(value)
+    return decoded_value[2] if decoded_value else None
+
+
+"""
+Source taken from: https://gist.github.com/mjpieters/86b0d152bb51d5f5979346d11005588b
+"""
+
```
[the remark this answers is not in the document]
ahangsu: ah, fair enough.

algosdk/source_map.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Decodes a VLQ-encoded source mapping between PC values and TEAL source code lines.
+    Spec available here: https://sourcemaps.info/spec.html
+
+    Args:
+        source_map (dict(str, Any)): source map JSON from algod
+        delimiter (str, optional): delimiter for mappings
+    """
+
+    def __init__(self, source_map: Dict[str, Any], delimiter: str = ";"):
```
michaeldiamant, line 14: @barnjamin For my understanding:  What motivated exposing `delimiter` for user configuration?

From my read of the spec, the delimiter is _always_ `;`.  I'm trying to gauge if a situation exists where the user supplies a source map v3 compatible mapping with different delimiter.

Aside:  _If_ the delimiter is removed here, then I think the same change ought to happen in js-algorand-sdk.

algosdk/source_map.py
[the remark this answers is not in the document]
algochoi: IIRC we are converting from PC->line here, so consider a mapping like: `;1;;2` (I used ints rather than VLQ for readability).

Then `pc_list` would contain `{0:None, 1: 1, 2: None, 3: 2}` (the values are decoded from `_decode_int_value`). I think I left a comment in the original decoding scheme to remind myself that doing

```
if line_num:
    blah
```

could be misleading since a `0` for line_num could be valid.

algosdk/source_map.py
```diff
@@ -0,0 +1,93 @@
...
+        self.delimter = delimiter
+
+        self.version: int = source_map["version"]
+        self.sources: List[str] = source_map["sources"]
+        self.mapping: str = source_map["mapping"]
```
michaeldiamant, line 19: @barnjamin Open to discussion:  While reviewing, I realized go-algorand incidentally returns _mapping_ instead of the spec defined _mappings_ key.

I think it's most prudent to address the underlying issue (which will cause a backwards incompatibility) because the current implementation deviates from the spec.

I apologize for the inconvenience - I reviewed the PR in-question and missed the deviation (https://github.com/algorand/go-algorand/pull/3726).

algochoi: Uhoh, good catch

barnjamin: I can update the SDKs and tests to `mappings` but I suppose we should wait until the changes are merged into go-algorand before releasing? I can also put something in here to check for either key for now?

michaeldiamant: @barnjamin Here's my thinking - What do you think?
* In go-algorand, the principled change is to:
  * Deprecate `mapping` via warning in OAS _description_, and continue populating it.  Circle back to removing it another day.
  * Add `mappings` field.  
*  In SDKs, check `mappings` first and if the key is absent, fallback to `mapping`.  When removing `mapping` server-side, we can remove the secondary look up in SDKs.

If the thinking makes sense, I can draft a story to track removal of the to-be-deprecated `mapping` field.

algosdk/source_map.py
michaeldiamant: @barnjamin I agree with your rationale.

I am surprised there's no _well-supported_ reference implementation in Python.  Feels preferable to depend on a library here, but I didn't find one.

algosdk/source_map.py
michaeldiamant: @barnjamin Resolving - Leaving a few notes for context.

* Thanks for making changes to support falling back between `mapping` and `mappings`.  
  * I have slight preference to prefer the non-deprecated field (`mappings`).  But, it's a minor concern and preferring `mapping` works for me.
* I opened a few stories to track deprecation:
  * Epic covering all deprecations - https://github.com/algorand/algorand-sdk-testing/issues/200
  * Story for `mapping` removal - https://github.com/algorand/algorand-sdk-testing/issues/202

algosdk/source_map.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            self.mappings: str = source_map["mappings"]
+
+        pc_list = [
+            _decode_int_value(raw_val) for raw_val in self.mappings.split(";")
+        ]
+
+        # Initialize with 0,0 for pc/line
+        self.pc_to_line: Dict[int, int] = {0: 0}
+        self.line_to_pc: Dict[int, List[int]] = {0: [0]}
```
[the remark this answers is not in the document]
michaeldiamant: @barnjamin Got it - I'd appreciate a double check on the spec from you + reviewers.  Provided no concerns raised, I favor removing the initialization to more closely adhere to the spec.

algosdk/source_map.py
```diff
@@ -0,0 +1,101 @@
...
+            if line_num is not None:  # be careful for '0' checks!
+                if line_num not in self.line_to_pc:
+                    self.line_to_pc[line_num] = []
+                self.line_to_pc[line_num].append(index)
+                last_line = line_num
+
+            self.pc_to_line[index] = last_line
+
+    def get_line_for_pc(self, pc: int) -> int:
+        return self.pc_to_line[pc]
```
michaeldiamant, line 49: @barnjamin What's the desired behavior when the key (PC) doesn't exist?

As written, `get_line_for_pc` and `get_pcs_for_line` throw `KeyError`.  I'm double checking if the behavior is by-design.

If not, propose switching to `dict.get` and optionally, adding `Optional` return type hint to both methods.

michaeldiamant: Aside:  I think the other SDKs warrant consideration too.  

For example, the as written Go implementation returns the Map's zero value when provided a non-existent key.  I imagine we want to return `Nil`.

algosdk/source_map.py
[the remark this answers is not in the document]
barnjamin: removed

algosdk/source_map.py
```diff
@@ -0,0 +1,102 @@
...
+            # For backwards compat until go-algorand has the change then we should remove
+            self.mappings: str = source_map["mapping"]
+        else:
+            self.mappings: str = source_map["mappings"]
+
+        pc_list = [
+            _decode_int_value(raw_val) for raw_val in self.mappings.split(";")
+        ]
+
+        # Initialize with 0,0 for pc/line
```
michaeldiamant, line 34: @barnjamin Minor:  Comment appears to be stale because the `0,0` mapping no longer exists.

```suggestion
```

algosdk/source_map.py
michaeldiamant: Additionally, and optionally I'd consider adding a non-Cucumber unit test to enforce the desired behavior.

barnjamin: I've no preference on how we handle this so I'll lean towards reviewers opinions

algochoi: I also think we should avoid returning an `Exception`. I would propose `dict.get(pc, None)` and return a `None` or (less preferred) return a deepcopy of the dictionary so users can handle their own errors.

ahangsu: I will up with returning `None`, but we may still want to expose `line_to_pc` or `pc_to_line` for someone wants to iterate through the structure.

algosdk/source_map.py
```diff
@@ -0,0 +1,101 @@
...
+        pc_list = [
+            _decode_int_value(raw_val) for raw_val in self.mappings.split(";")
+        ]
+
+        self.pc_to_line: Dict[int, int] = {}
+        self.line_to_pc: Dict[int, List[int]] = {}
+
+        last_line = 0
+        for index, line_delta in enumerate(pc_list):
+            if line_delta is not None:  # be careful for '0' checks!
```
ahangsu, line 39: nit: would be great if we could have a more elaborated comment on this part, discussing that `line_delta` would be `None` if `line_delta == 0`. if I am understanding it correctly

barnjamin: I think this is correct? @algochoi?

This comment made me re-evaluate the logic I had to add pc to the line_to_pc map, I think it was a bug to only append to the map IFF the line_delta was not None

algochoi: The current logic here looks correct to me 👍 

In the new encoding scheme, if the value is explicitly zero or `None`, it should equal the `last_line`.

algochoi: I think you can delete the comment there: `# be careful for '0' checks!` as it might cause further confusion?

barnjamin: its changed, the above snip is outdated 👍

algosdk/source_map.py
```diff
@@ -0,0 +1,99 @@
...
+        pc_list = [
+            _decode_int_value(raw_val) for raw_val in self.mappings.split(";")
+        ]
+
+        self.pc_to_line: Dict[int, int] = {}
+        self.line_to_pc: Dict[int, List[int]] = {}
+
+        last_line = 0
+        for index, line_delta in enumerate(pc_list):
+            # line_delta is None if the line number has not not changed
```
algochoi, line 35:
```suggestion
            # line_delta is None if the line number has not changed
```

ahangsu (approved):
looking good, once Andrew's suggestion is resolved, we should probably ship it?

michaeldiamant (approved):
@barnjamin Leaving approval with a friendly reminder to revert branch change in `run_integration.sh` prior to merging.

PR #356, opened 2022-06-29
Signer address

barnjamin:
I'd like to pass around only the signer but need easy access to the address for populating transactions as well

Would need to be implemented across SDKs


barnjamin:
@jasonpaulos as I think about this more, I think its fine to skip anything that accounts for a rekey explicitly, we can keep all else the same and if the dev is working with rekeyed accounts they should know not to set the sender as the `signer.address()`, wdyt?

PR #358, opened 2022-06-30
Github-Actions: Adding pr title and label checks

algojack:
* adding pr title and label checks
* adding github release notes generator config


algobarb (approved):
We should make sure developers are aware of this PR change on the SDK repos

PR #360, opened 2022-07-08
Testing: Modify cucumber steps to use dev mode network

algochoi:
This PR introduces some instrumentation to support dev mode testing (add zero payment transaction to advance blocks in certain scenarios) and delete v1 algod cucumber tests.

### v1 API Changes

There are two changes to the v1 steps:
* Added a transient dev mode account to safely send some transactions after a potential rekey transaction.
* Removed some algod calls in `the transaction should go through` step, especially ones that test recent block rounds (harder to test in dev mode as blocks need to be manually advanced and may not be recent).

### Experimental results

Initial testing looks like just deleting v1 algod steps will have a ~7 min speedup (21min -> 14min)
* Old: https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/687/workflows/aeaedd23-10de-40cf-9c76-3dd909e77df9/jobs/2435
* New https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/692/workflows/985979f8-54b3-45ac-b2b2-e4d24e989208/jobs/2499

Dev mode, disabling wait_for_transaction v1 step (~10 min): 
* https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/711/workflows/f1d3e702-cf2a-40dc-ac8a-49731c2b6f36/jobs/2643


run_integration.sh
```diff
@@ -7,7 +7,7 @@ pushd $rootdir
 
 # Reset test harness
 rm -rf test-harness
-git clone --single-branch --branch master https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch devmodenet https://github.com/algorand/algorand-sdk-testing.git test-harness
```
algochoi, line 10: Related SDK Testing PR: https://github.com/algorand/algorand-sdk-testing/pull/206

algochoi, line 10: Related SDK Testing PR: https://github.com/algorand/algorand-sdk-testing/pull/206

Makefile
```diff
@@ -2,7 +2,7 @@ UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.led
 unit:
 	behave --tags=$(UNITS) tests -f progress2
 
-INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey or @send.keyregtxn or @send or @compile.sourcemap"
+INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @send.keyregtxn or @send or @compile.sourcemap"
```
algochoi, line 5: The existing v1 rekeying tests alters the default wallet key - I decided to just remove it here, but a more complete solution would be to re-write them so that they alter a transient private key instead.

tests/steps/application_v2_steps.py
```diff
@@ -406,16 +408,15 @@ def remember_app_id(context):
 def wait_for_app_txn_confirm(context):
     sp = context.app_acl.suggested_params()
     last_round = sp.first
+    send_zero_transactions(context, 3)
```
tzaffi, line 411: Curious on how the number 3 was arrived at.

tests/steps/other_v2_steps.py
```diff
@@ -97,6 +98,31 @@ def read_program(context, path):
...
 
+# Send transactions to progress block numbers on dev mode.
+def send_zero_transactions(context, txns=1):
+    sp = context.app_acl.suggested_params()
+    for _ in range(txns):
+        payment = transaction.PaymentTxn(
+            context.accounts[0],
+            sp,
+            constants.ZERO_ADDRESS,
+            random.randint(100000, 900000),
```
tzaffi, line 109: Was this random payment arrived at for avoiding transaction duplication? If so, this can also be achieved by randomizing the note field. It might prove worthwhile to send the smallest amount possible.

tests/steps/other_v2_steps.py
```diff
@@ -97,6 +98,31 @@ def read_program(context, path):
...
+        )
+        signed_payment = context.wallet.sign_transaction(payment)
+        context.app_acl.send_transaction(signed_payment)
+        # Wait and confirm that the zero payment succeeded.
+        transaction.wait_for_confirmation(
```
tzaffi, line 114: Is the number of rounds that confirmations are waited for predictable in _dev mode_?

tests/steps/other_v2_steps.py
```diff
@@ -97,6 +98,31 @@ def read_program(context, path):
     return read_program_binary(path)
 
 
+# Send transactions to progress block numbers on dev mode.
+def send_zero_transactions(context, txns=1):
```
tzaffi, line 102: nit: I was a bit confused by the name of this function, thinking maybe it meant "send zero algos" or "send no transactions". So perhaps consider a more descriptive name such as "burn_algo_transactions(context, num_txns=1)". (also reccomend changing the parameter name)

tests/steps/steps.py
```diff
@@ -1,28 +1,42 @@
...
-from algosdk import util
-from algosdk import constants
-from algosdk import logic
-from algosdk.future import template
 import os
+import random
 from datetime import datetime
-import hashlib
 
+from algosdk import (
```
tzaffi, line 6: Definitely an improvement in making imports from `algosdk` more pythonic. However, I think the external imports from `behave` and `nacl.signing` ought to be put in a section above

tests/steps/steps.py
```diff
@@ -404,19 +421,21 @@ def send_msig_txn(context):
 
 @then("the transaction should go through")
 def check_txn(context):
-    last_round = context.acl.status()["lastRound"]
+    send_zero_transactions(context)
+    # last_round = context.acl.status()["lastRound"]
```
tzaffi, line 425: recommend removal of commented out code in this function

tzaffi (commented):
I haven't tested locally yet, but I plan to. I have some questions mostly regarding transaction amounts and the number of rounds needed to be waited for in dev mode.

Also some minor suggestions.

Makefile
tzaffi: Do you know if this problem is universal across SDK's?

tests/steps/other_v2_steps.py
algochoi: I played around with `randbytes` and the `note` field previously and just decided on randomizing the amount field: https://github.com/algorand/py-algorand-sdk/pull/360/commits/70c034b6405da73ac368ed36e7b77b8b2294652c#diff-9c51d98c2ed173b0192084b3c3aec41f30bfaa0ade57f240f63863602942d184L110

If you feel strongly about a particular way, I could go back to randomizing the note field instead.

tests/steps/other_v2_steps.py
algochoi: It's more or less always 1 (the block is finalized after the transaction is sent). 3 and 5 are mostly magic numbers right now - I could consolidate them into one for cleanliness.

tzaffi: I don't feel strongly. Knowing that an exact amount is deducted each time gives a way of calculating how many such transactions were called (but perhaps in dev mode, the round number gives similar info)

tests/steps/application_v2_steps.py
algochoi: In the line below, it checks the status after block `last_round + 2`, so I sent 3 transactions (advances 3 blocks in dev mode) for good measure!

tests/steps/steps.py
algochoi: Good point - I'm in favor or removing these lines because there are other steps that do the same thing in this SDK, e.g. 
`I can get the transaction by ID`

tests/steps/steps.py
algochoi: I used the auto-import on VS Code - I think your suggestion makes sense but might require more manual setup/overriding some defaults on the editor.

tests/steps/other_v2_steps.py
algochoi: Good point - done!

Makefile
algochoi: > Do you know if this problem is universal across SDK's?

I haven't looked into all of them, but I think it would be. I also think the order of execution matters - it poses the least amount of problems if `@rekey` is executed last.

I did push a workaround hack for the v1 functions in my latest commit - I created a transient dev account with some funds to send transactions before a potential rekey.

tests/steps/other_v2_steps.py
tzaffi: I don't feel strongly. Knowing that an exact amount is deducted each time gives a way of calculating how many such transactions were called (but perhaps in dev mode, the round number gives similar info)

tests/steps/other_v2_steps.py
tzaffi: I think this would be a good place to consolidate as you suggest. Also, hopefully we would discover at this juncture that 1 doesn't work. We could have `wait_for_confirmation()` default to 1 round and not even pass that parameter into it.

tzaffi: Same thought applies to the new `dev_mode_wait_for_confirmation()`

tests/steps/other_v2_steps.py
```diff
@@ -97,6 +98,31 @@ def read_program(context, path):
...
+        context.app_acl.send_transaction(signed_payment)
+        # Wait and confirm that the zero payment succeeded.
+        transaction.wait_for_confirmation(
+            context.app_acl, payment.get_txid(), 5
+        )
+
+
+# To prevent excess waiting, send a zero payment transaction before
+# the wait_for_confirmation function in dev mode.
+def dev_mode_wait_for_confirmation(context, txid, rounds):
```
tzaffi, line 121: default round to 1?

tests/steps/steps.py
```diff
@@ -1,28 +1,57 @@
...
 
 
+# Initialize dev mode accounts.
+def initialize_dev_mode_account(context):
+    context.dev_sk, context.dev_pk = account.generate_account()
+    payment = transaction.PaymentTxn(
+        sender=context.accounts[0],
+        sp=context.acl.suggested_params_as_object(),
+        receiver=context.dev_pk,
+        amt=10_000_000,
```
tzaffi, line 34: is 10 algo's enough? I'm not sure how much the account is supposed to spend during the lifetime of all tests, but it's not hard to imagine it could run out of algos if it's workload is sufficiently complex. Might be worth creating a constant `ACCOUNT_INITIAL_MICROALGOS: int = ... the amount...` at the top of the file and use that here.

algochoi: For the v1 steps, I thought it would be sufficient to hardcode to a smaller value because the workload isn't too complex and we probably wouldn't be extending the tests anyway - but to be extra safe, I set the receiver to itself so it wouldn't run out of algos. I do like the constant suggestion and implemented in the latest commit.

tests/steps/other_v2_steps.py
```diff
@@ -97,6 +98,32 @@ def read_program(context, path):
...
 
+# Send transactions to progress block numbers on dev mode.
+def burn_algo_transactions(context, num_txns=1):
+    sp = context.app_acl.suggested_params()
+    for _ in range(num_txns):
+        payment = transaction.PaymentTxn(
+            context.accounts[0],
+            sp,
+            constants.ZERO_ADDRESS,
+            random.randint(100000, 900000),
```
michaeldiamant, line 109: @algochoi Optionally, I propose configuring the upper bound proportional to the funding amount.  Intent is to avoid generating a value that's too large if/when the funding amount changes.

Here's the Java SDK analog:  https://github.com/algorand/java-algorand-sdk/pull/350/commits/3390a6d687c5e57c349064de5014afc4f34a3b2d.

Feel welcomed to resolve as is.

algochoi: Good point - Added proportional funding amounts in latest commit https://github.com/algorand/py-algorand-sdk/pull/360/commits/3f110572a023ddadaffd0988b02a48a16721c77c

tests/steps/steps.py
```diff
@@ -404,20 +468,18 @@ def send_msig_txn(context):
...
-    last_round = context.acl.status()["lastRound"]
+    wait_for_algod_transaction_processing_to_complete()
     assert "type" in context.acl.pending_transaction_info(
         context.txn.get_txid()
     )
-    context.acl.status_after_block(last_round + 2)
     assert "type" in context.acl.transaction_info(
         context.txn.sender, context.txn.get_txid()
     )
-    assert "type" in context.acl.transaction_by_id(context.txn.get_txid())
```
michaeldiamant, line 415: @algochoi Can you restore the deleted assertion?

I'm _mostly_ confident it works now that the waiting approach changed.  See corresponding change in https://github.com/algorand/java-algorand-sdk/pull/350/commits/fbcbf63efaa65f31d3cf011109b070b283f296db#diff-2c7314dd90c34b904f243500da609439f53ef9791207a076eec677cb0e8d2611L829-L830.

algochoi: Good point - done!

michaeldiamant (approved):
@algochoi Great work here - looks ready to me.  ☕ 

Approval presumes the following prior to merge:
* Revert algorand-sdk-testing branch to `master`.
* Remove `NETWORK_TEMPLATE` override.  Assumes upstream defaults to `DevModeNetwork`.

PR #365, opened 2022-07-28
Minimize use of transactions to advance rounds in #360

michaeldiamant:
Extends #360 to propose minimizing use of transactions to advance rounds.

### DIFF
--- a/tests/steps/account_v2_steps.py
+++ b/tests/steps/account_v2_steps.py
@@ -3,8 +3,7 @@ from typing import Union
 from algosdk import account, constants, encoding, logic
 from algosdk.future import transaction
 from behave import given, then, when
-from tests.steps.other_v2_steps import dev_mode_wait_for_confirmation
-
+import tests.steps.other_v2_steps # Imports MaybeString
 
 def fund_account_address(
     context, account_address: str, amount: Union[int, str]
@@ -18,7 +17,7 @@ def fund_account_address(
     )
     signed_payment = context.wallet.sign_transaction(payment)
     context.app_acl.send_transaction(signed_payment)
-    dev_mode_wait_for_confirmation(context, payment.get_txid(), 10)
+    transaction.wait_for_confirmation(context.app_acl, payment.get_txid(), 10)
 
 
 @when(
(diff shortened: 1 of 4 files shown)


tests/steps/account_v2_steps.py
```diff
@@ -3,7 +3,7 @@
 from algosdk import account, constants, encoding, logic
 from algosdk.future import transaction
 from behave import given, then, when
-from tests.steps.other_v2_steps import dev_mode_wait_for_confirmation
+import tests.steps.other_v2_steps  # Imports MaybeString
```
algochoi, line 6: Good catch

tests/steps/account_v2_steps.py
```diff
@@ -18,7 +18,7 @@ def fund_account_address(
     )
     signed_payment = context.wallet.sign_transaction(payment)
     context.app_acl.send_transaction(signed_payment)
-    dev_mode_wait_for_confirmation(context, payment.get_txid(), 10)
+    transaction.wait_for_confirmation(context.app_acl, payment.get_txid(), 10)
```
algochoi, line 21: I think waiting one round would be sufficient for dev mode:

```suggestion
    transaction.wait_for_confirmation(context.app_acl, payment.get_txid(), 1)
```

michaeldiamant: @algochoi I agree with the suggestion though I think it ought to be applied across-the-board (in other SDKs too).  https://github.com/algorand/py-algorand-sdk/pull/365/commits/a37f2407999dcc04cf9de83a51c377a9c2e1094c brings in the feedback.

Can you resolve if it looks good?

algochoi (approved):
Thanks for the changes! LGTM

tests/steps/application_v2_steps.py
```diff
@@ -404,21 +403,26 @@ def remember_app_id(context):
     context.app_ids.append(app_id)
 
 
+def wait_for_algod_transaction_processing_to_complete():
+    """
+    wait_wait_for_algod_transaction_processing_to_complete is a Dev mode helper method that's a rough analog to `context.app_acl.status_after_block(last_round + 2)`.
```
algochoi, line 408: I can change this after merge so we don't run the CI again:
```suggestion
    wait_for_algod_transaction_processing_to_complete is a Dev mode helper method that's a rough analog to `context.app_acl.status_after_block(last_round + 2)`.
```

PR #366, opened 2022-08-02
Box support for Indexer endpoints

algochoi:
This PR adds support for boxes endpoints for indexer and adds cucumber steps for testing the indexer path.


tests/steps/application_v2_steps.py
```diff
@@ -185,6 +185,23 @@ def search_application2(context, creator):
     context.response = context.icl.search_applications(creator=creator)
 
 
+@when(
+    'we make a LookupApplicationBoxByIDandName call with applicationID {app_id} with encoded box name "{box_name}"'
+)
+def lookup_application_box(context, app_id, box_name):
+    boxes = split_and_process_app_args(box_name)[0]
```
ahangsu, line 192: in both py/js-sdk, this `split_and_process_app_args` is applied to parse `boxName`, while in go/java-sdk, this is not the case, for `name` does simply put string into params in query.

Do we want to unify the sdk behaviors in this case?

algochoi: This is deliberate because the Java/Go SDKs have different inputs due to generated code. The JS/Py SDKs' APIs take box name as bytes instead of strings.

ahangsu (approved):

algosdk/v2client/indexer.py
```diff
@@ -926,6 +926,44 @@ def application_logs(
...
+    def application_boxes(
+        self, application_id: int, limit: int = 0, next_page=None, **kwargs
+    ):
+        """
+        Return a list of all the application's boxes.
+
+        Args:
+            application_id (int): The ID of the application to look up.
+            limit (int, optional): Max number of box names to return.
+                If max is not set, or max == 0, returns all box-names.
```
tzaffi, line 955:
```suggestion
                If max is not set, or max == 0, returns all box-names up to queried indexer's `defaultBoxesLimit`.
```

algosdk/v2client/indexer.py
```diff
@@ -926,6 +926,44 @@ def application_logs(
...
+            next_page (string, optional): used for pagination
```
tzaffi, line 956: `next_page` is a box name and expects the same encoding methodology as the `box_name` parameter of `application_box_by_name()`. As an example, [consider this](https://github.com/algorand/indexer/blob/590251145af5b0d08d37bed911caacc5b08681be/api/test_resources/boxes.json#L566-L605).

We should be clear about how this parameter should be passed back. I see two ways going forward:
1. callers provide `next_page` already `b64` encoded. This makes sense because this allows them to plug in _without any processing_ the value of `next-token` which they received from the previous call.
2. alternatively, they could provide the box name in the same format as in the call to `application_box_by_name()`. The downside is that the user whould first have to b64 _decode_ and then provide, and then this method would encoded it again. 

I favor appoach (1) because it's more efficient. But I could see arguments for (2).

In either case, the prefix `b64:` needs to be pre-prended to `next_page`, i.e. `f"b64:{next_page}"` , before gelling **GET**.

ahangsu: That is a good point. On the perspective of a caller, when we are querying for `next_page`, we would wish for passing in a string (or `Uint8Array`) that is close to readable format.

Another concern is that, if we want to prepend  `b64:` to encoded `next_page`, we may want to check if `next_page` is empty or not, or as you suggested in js-sdk, the resulting query would be funny with only `b64:`.

ahangsu: Going to take back what I thought about `b64:` empty string, if that is valid in indexer query, then we should be fine with it.

tzaffi: Yes, it's ok (if a bit strange) to give empty `b64:` Indexer will just query starting at this minimum of all strings and return the same results as if not provided `b64:`... ideally, it's better not to provide if it's not too big a hassle.

ahangsu: just circled back from langspec after a while, and I am in favor of option 1, which is caller provides `next_page` to be `b64` encoded.

My additional thought is that, but the method should check if `next_page` is `None` or `""` before prefixing with `b64:`, though I assume it is find for indexer to accept things like `b64:` with empty string.

Standing on the perspective of debugging, or making queries more 'cleaner', if one do not want to add `next_page`, then it should not appear in query tho... and I think the change across SDKs should be low effort.

ahangsu: From the decision we made in https://github.com/algorand/algorand-sdk-testing/pull/219, we keep the `nextToken` method unchanged.

run_integration.sh
```diff
@@ -8,7 +8,7 @@ pushd $rootdir
 # Reset test harness
 rm -rf test-harness
 # TODO: Before merging, change branch back to master
-git clone --single-branch --branch feature/box-storage https://github.com/algorand/algorand-sdk-testing.git test-harness
+git clone --single-branch --branch indexer-boxes https://github.com/algorand/algorand-sdk-testing.git test-harness
```
tzaffi, line 11: Same issue as with the other 2. This file shouldn't exist.

tzaffi (changes requested):
Similar issue. Apologies for this extra work due to refactoring the testing in all the sdk's.

run_integration.sh
ahangsu: correct, now we are changing inside of `.test-env`

.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="feature/box-storage"
+SDK_TESTING_BRANCH="indexer-boxes"
```
ahangsu, line 3: TODO: revert back once algorand-sdk-testing PR merge
```suggestion
SDK_TESTING_BRANCH="feature/box-storage"
```

tzaffi (approved):

PR #367, opened 2022-08-03
Enhancement: Trim the indexer images and use the sandbox instead of custom dockers

tzaffi:
# Utilize New Sandbox-based testing SDK and Obsolete Steps

* Refactoring `make docker-test` using pure-shell script `test-harness.sh` that uses the sandbox instead of the old docker scripts
  * NOTE: an attempt has been made to make this easy to re-use in other SDK's with slight modification
* Removing steps that belong to features being removed in [refactored SDK testing](https://github.com/algorand/algorand-sdk-testing/pull/220)

## Defining Issue
https://github.com/algorand/algorand-sdk-testing/issues/221

## TODO
- [x] following a conversation with @michaeldiamant - `.test-env` should have the bare essentials needed to control this SDK's test setup and run. It should not overwrite SDK Testing's `.env`. This is because the SDK Testing repo has the authority of what branches of `go-algorand`, `indexer`, etc. are relevant for a particular feature being tested.
- [x] create an issue in the sdk-testing repo to improve CI by standing up a test-harness with a simple call to `up.sh`as a github-actions based integrations test. **CREATED**: https://github.com/algorand/go-algorand-internal/issues/2235
- [x] Has `trim-indexer` branch of `algorand-sdk-testing` been merged?
  - [x] Repoint the dependency here to `algorand-sdk-testing/master`
  - [x] Now we can merge

## At a High Level
`make docker-test` triggers the following:
* **A.** `make harness` which calls `./test-harness.sh` (replacing `run_integration.sh`):
  1. clone the **SDK Testing Repo** into `./test-harness/*` (renamed from `/test`)
  2. copy the env vars in `.test-env` to `./test-harness/.env` (this is brand new)
  3. copy the cloned **features** into `./tests/features/*` (same as previous)
  4. run the cloned `./test-harness/up.sh` which now does the following:
    * **a.** use `.test-env` to interpolate a sandbox config template into a temporary file `./test-harness/_config.harness`
      * NOTE: just as previously, there is logic to distinguish between **TYPE** `channel` v. `source`
      * NOTE: there is also logic to handle the quirkiness of `sed` on mac os versus linux
   * **b.** clone the sandbox repo into `./test-harness/.sandbox/*`
   * **c.** copy `.test-env` (via its copy `/test-harness/.env`) into `./test-harness/.sandbox/.env` which serves as a Docker-compatible env file
   * **d.** move the populated config template into `./test-harness/.sandbox/config.harness`
   * **e.** stand up the sandbox with `./sandbox up -v harness`
* **B.** `make docker-pysdk-build`: builds this repo's `DockerFile` into the image `py-sdk-testing` 
* **C.** `make docker-pysdk-run`: executes the `py-sdk-testing` which runs all cucumber tests

## Summary of Changes
* `.test-env` is introduced to configure the test build and the cloned sandbox. During the build it 

## Step Removals
Please refer to the **step removal guide** in [the companion SDK Testing PR](https://github.com/algorand/algorand-sdk-testing/pull/220).

### Additional Steps being modified due to indexer v1 removal
* [`@rekey_v1` and `@send` tags] - `then("the transaction should go through")`
  * DO NOT `assert "type" in context.acl.transaction_by_id(context.txn.get_txid())` because this relies on indexer v1
* [`@applications.verified` tag] - `step("I wait for the transaction to be confirmed.")`
  * DO NOT `assert "type" in context.acl.transaction_by_id(context.app_txid)`

### Additional Steps being removed due to **Unused Steps Analysis**
* `@when("we make a SearchForApplications call with {application_id} and {round}")`
* `@when("we make a SearchForApplications call with {application_id} and {round}")`
* `@then("Every transaction works with asset-id {assetid}")`
* `@then('the parsed response should equal "{jsonfile}".')`
* `@then('we expect the path used to be "{path}"')`
* `@when("I create the payment transaction")`
* `@when('I read a transaction "{txn}" from file "{num}"')`
* `@when("I write the transaction to file")`
* `@then("the transaction should still be the same")`
* `@then("I do my part")`
* `@given('key registration transaction parameters {fee} {fv} {lv} "{gh}" "{votekey}" "{selkey}" {votefst} {votelst} {votekd} "{gen}" "{note}"')`
* `@when("I create the key registration transaction")`

## Statistics
Build time average case:
* latest 5.5 minutes (see the next section)
* former [7.5 minutes](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/761/workflows/9d916fec-64a7-4fd0-9bf5-3a04df265bb5/jobs/3017)

### Running log of build times
1. Py 3.10 @ 3,000 rounds: [5m 37s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/805/workflows/302732f3-ee1e-4fff-866d-4c814be3abdd/jobs/3377)
2. Py 3.10 @ 30,000 rounds **FIRST TIME**: [6m 18s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/806/workflows/c7431da7-7ae8-4125-b83a-a20ac88f1aa4/jobs/3380)
3. Py 3.10 @ 30,000 rounds **SECOND TIME**: [5m 41s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/809/workflows/e3c9494b-b5f5-44f9-b2eb-535de20b07ba/jobs/3406)
4. "" "" **THIRD TIME**: [6m 8s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/810/workflows/418a9f35-771b-4e31-8289-c30c907d29fb/jobs/3412)
5. Py 3.10 @ 30K rounds + w/o indexer v1 steps: [5m 34s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/813/workflows/91cda4e7-1530-43c4-be87-82304134f6ac/jobs/3434)
6. "" "" "": [5m 35s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/814/workflows/01e6180a-74c7-4af1-95eb-930529f8b39a/jobs/3439)
7. "" "" "" (but after sandbox bugfix): [5m 49s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/817/workflows/cdb1508c-13e4-4597-a825-13d0774c4a3f/jobs/3458)
8. "" "" "" [7m 42s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/826/workflows/15d98bcd-1e0c-437b-a886-67fd1cb8b12e/jobs/3520) `:-(` ...not sure if that's a blip or a new trend
9. "" "" "" [7m 37s](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/828/workflows/eca0fec2-dab2-45fb-8847-3e84a6833c76/jobs/3537) ... seems to be a new unfortunate trend

## Testing
No functional changes are introduced in this PR. Therefore, the currently running tests are sufficient but should be tested locally in addition to CI.


Makefile
```diff
@@ -4,8 +4,23 @@ unit:
...
+
+docker-pysdk-run:
+	docker ps -a | cat
+	docker run -it --network host py-sdk-testing:latest
+
+docker-test: harness docker-pysdk-build docker-pysdk-run
+
+
+
+
```
tzaffi, line 26:
```suggestion
```

test-harness.sh
```diff
@@ -0,0 +1,33 @@
...
+
+
+## Start test harness environment
+pushd "$SDK_TESTING_HARNESS"
+./scripts/up.sh
+popd
+echo "run_integration.sh: seconds it took to finish testing sdk's up.sh: " + $(($(date "+%s") - $START))
+echo "To run sandbox commands, cd into $SDK_TESTING_HARNESS/$SANDBOX"
+echo "time till end of test-harness bootstrapping: " + $(($(date "+%s") - $START))
+
```
tzaffi, line 33:
```suggestion
```

Makefile
tzaffi, line 26:
```suggestion
```

Makefile
```diff
@@ -4,8 +4,128 @@ unit:
...
+docker-restart:
+	./test-harness/scripts/restart.sh
+
+abi-args-integration:
+	behave -D BEHAVE_ON_ERROR --tags="@z" -k -f plain test/features/integration/abi.feature
+
+abi-args-unit:
+	behave -D BEHAVE_ON_ERROR --tags="@unit.atc_method_args" -k -f plain test/features/unit/atomic_transaction_composer.feature
+	# behave -D BEHAVE_ON_ERROR --tags="@z" -k -f plain test/features/unit/atomic_transaction_composer.feature
+
```
tzaffi, line 131:
```suggestion
```

.test-env
```diff
@@ -0,0 +1,42 @@
+# Configs for local python sdk dockerizing:
+SKIP_PYTHON_SDK_DOCKER=0
```
tzaffi, line 2:
```suggestion
```

.test-env
```diff
@@ -0,0 +1,41 @@
+# Configs for local python sdk dockerizing:
```
tzaffi, line 1:
```suggestion
# Configs for python sdk docker:
```

.test-env
```diff
@@ -0,0 +1,41 @@
...
+
+# Used to determine sandbox build type:
+TYPE=${TYPE:-source} # OR "channel"
+
+# Used when TYPE==channel
+ALGOD_CHANNEL="nightly"
+
+# Used when TYPE==source
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="feature/avm-box"
```
tzaffi, line 17: ^^^ TODO: don't merge while pointing to non-releasable branch ^^^

.test-env
```diff
@@ -0,0 +1,41 @@
...
+ALGOD_SHA=""
+
+# Used regardless of TYPE
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"
+NETWORK_NUM_ROUNDS=3000
+INDEXER_URL="https://github.com/algorand/indexer"
+NODE_ARCHIVAL="True"
+NODE_V1_INDEXER="True"
```
tzaffi, line 25: As long as we keep testing indexer v1, we'll need to keep these as `True`

.test-env
```diff
@@ -0,0 +1,41 @@
...
+INDEXER_BRANCH="tzaffi/box-search"
```
tzaffi, line 26: ^^^ TODO: don't merge with this non-releasable branch

.test-env
```diff
@@ -0,0 +1,41 @@
...
+INDEXER_SHA=""
+
+# Sandbox configuration:
+SANDBOX_BRANCH="configurable-ports"
```
tzaffi, line 30: ^^^ TODO: don't merge until this has been merged and re-point to `master`

test-harness.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
algochoi, line 1: This is just a general observation, but sandbox errors seem to be swallowed somewhere in the output. 

```
➜  py-algorand-sdk git:(trim-indexer) 

... (truncated)

Starting sandbox for: harness
see sandbox.log for detailed progress, or use -v.
* docker containers started!
* waiting for services to initialize.
* services ready!

algod version
make: *** [harness] Error 1

➜  py-algorand-sdk git:(trim-indexer) 
```

Sandbox logs does have the error though - would there be a way to pass a `-v` or something to echo this back into the terminal? If not, it's not a huge deal for the user to parse through the sandbox logs either.

test-harness.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "test-harness.sh: directory $SDK_TESTING_HARNESS does not exist - NOOP"
+fi
+
+git clone --single-branch --branch "$SDK_TESTING_BRANCH" "$SDK_TESTING_URL" "$SDK_TESTING_HARNESS"
+
+## OVERWRITE incoming .env with .test-env
+cp "$ENV_FILE" "$SDK_TESTING_HARNESS"/.env
+
+## Copy feature files into the project resources
+mkdir -p tests/features
```
algochoi, line 29: Should there be a line to remove any stale tests? If we are going for a "fresh" `make` and `test`, it's nicer to remove existing tests:

```
rm -rf tests/features
```

algochoi (commented):
Nice! The shorter build times are super encouraging and works fine on my local machine - left some thoughts after using it below 👇

.test-env
```diff
@@ -0,0 +1,41 @@
+# Configs for python sdk docker:
+PYTHON_VERSION=${PYTHON_VERSION:-3.10}
+
+# Configs for testing repo download:
+SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
+SDK_TESTING_BRANCH="trim-indexer"
```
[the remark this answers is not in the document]
tzaffi: ^^^ TODO: REVERT TO `master` BEFORE MERGING

test-harness.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "test-harness.sh: seconds it took to get to end of cloning + copying: " + $(($(date "+%s") - $START))
+
+
+## Start test harness environment
+pushd "$SDK_TESTING_HARNESS"
+./scripts/up.sh
+popd
+echo "test-harness.sh: seconds it took to finish testing sdk's up.sh: " + $(($(date "+%s") - $START))
+echo "To run sandbox commands, cd into $SDK_TESTING_HARNESS/$SANDBOX"
+echo "time till end of test-harness bootstrapping: " + $(($(date "+%s") - $START))
```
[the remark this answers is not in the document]
tzaffi: this is extraneous. Shall delete shortly.

test-harness.sh
```diff
@@ -0,0 +1,40 @@
...
+  popd
+  rm -rf "$SDK_TESTING_HARNESS"
+else
+  echo "test-harness.sh: directory $SDK_TESTING_HARNESS does not exist - NOOP"
+fi
+
+git clone --single-branch --branch "$SDK_TESTING_BRANCH" "$SDK_TESTING_URL" "$SDK_TESTING_HARNESS"
+
+## OVERWRITE incoming .env with .test-env
+cp "$ENV_FILE" "$SDK_TESTING_HARNESS"/.env
```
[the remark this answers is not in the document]
tzaffi: I've been grappling with this same issue. Here are some pros/cons for _keeping_ `.env` as is [in the SDK Testing Repo:](https://github.com/algorand/algorand-sdk-testing/blob/7af3ace6c9e66de45c20e5067fa4a63d8df8582a/.env)
* Pros
  1. Communicates and sets the standard defaults required to bootstrap the test harness. Any "future" SDK that wants to incorporate testing, has a reasonable starting point.
  2. `up.sh` in SDK testing doesn't need to know about other environment variables that need sourcing except for the standard `.env`
  3. By keeping all configurations in a single file `.test-env` we know exactly where configurations should be provided
* Cons
  1. Can achieve the same (as pro # 1 above) with the former `.up-env` (or copying ast `.test-env`) that doesn't overwrite SDK testing's `.env` and get similar effect by sourcing in sequence (however pro # 2 gives a counter-argument)
  2. We can make SDK Testing's `.env` empty and communicate to the downloader that they need to provide the variables

After writing these pros/cons down, I've mostly convinced myself that SDK Testing _should_ have an env. But I'm intrigued by the idea of renaming `.test-env` as `.env` to gain some consistency. The down side (or maybe up side) is that we then need to consider the impact on the python `Dockerfile` since `.env` is read by the docker build, unless explicitly ignored in a `.dockerignore`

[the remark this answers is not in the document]
tzaffi: But I could also be missing some pros/cons. I welcome additions.

test-harness.sh
tzaffi: I explored this a bit, but passing verbose seemed to break the build (however, there were some other unrelated conditions which were different simultaneously, so I may be overly pessimistic). I commit to exploring this some more. 

Also, @algochoi :  I'm wondering what your error was. It seems like that is something worth fixing.

test-harness.sh
tzaffi: Good catch!

test-harness.sh
algochoi: Thanks for looking into it - I'm happy with resolving as is. The error was a docker-related issue - I don't have the exact logs anymore but docker failed due to some container name so I killed all my existing containers

test-harness.sh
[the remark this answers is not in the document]
algochoi: > up.sh in SDK testing doesn't need to know about other environment variables that need sourcing except for the standard .env

This seems to be a good reason to me - maybe we can consider pruning or even deleting the `.env` file in the SDK Testing repo then.

test-harness.sh
tzaffi: Maybe the run didn't succeed to bring down the pre-existing images? That's something it's supposed to be able to handle.

test-harness.sh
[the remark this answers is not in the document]
tzaffi: The pros/cons are a bit confusing and too much double negation is going on. My argument for `up.sh` not needing to know about other env vars means that `up.sh` need only call `source .env`. 

However, not providing an actual `.env` in SDK Repo in that case, means that the script would be broken as a standalone. I think the scripts in the SDK Testing repo should be able to run by themselves, without another repo needing to tweak them for them to work properly. That's the ultimate argument for having a `.env` in that repo, and for having the `.env` contain the minimum necessary env vars.

test-harness.sh
algochoi: I can't seem to replicate the error, but I think it could have been an error on Docker's side. The Docker UI was malfunctioning and required some permissions to delete containers so I had to remove it on the terminal manually.

.test-env
```diff
@@ -0,0 +1,41 @@
...
+ALGOD_CHANNEL="nightly"
+
+# Used when TYPE==source
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="feature/avm-box"
+ALGOD_SHA=""
+
+# Used regardless of TYPE
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"
+NETWORK_NUM_ROUNDS=3000
```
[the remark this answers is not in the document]
tzaffi: I'm not sure. I thought I was following your previous work but could have made a mistake. Glad to follow your lead here or investigate further.

test-harness.sh
tzaffi: leaving open as i want to investigate -v

.test-env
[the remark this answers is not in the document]
algochoi: Ah ok, I'm not sure if there is a default value we should settle on, but it looks like [this PR](https://github.com/algorand/sandbox/pull/111) has some explanation:

> The new falcon keys take a long time to generate. With this change the templates are updated to specify a reduced number (30000) of keys to generate. A side effect of this change means that the private sandbox networks will stall sometime within 24-48 hours.

Also if we use the sandbox network templates, it seems okay to delete the network configs in the SDK repo. I left a comment here: https://github.com/algorand/algorand-sdk-testing/pull/220/files#r942540578

test-harness.sh
tzaffi: This is now working in verbose mode

.test-env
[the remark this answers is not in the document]
tzaffi: I'll bump it up to 30000 and see if there is any effect on build times.

Makefile
```diff
@@ -4,8 +4,19 @@ unit:
 
 INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey_v1 or @send.keyregtxn or @send or @compile.sourcemap"
 integration:
-	behave --tags=$(INTEGRATIONS) tests -f progress2
+	behave --tags=$(INTEGRATIONS) tests -f progress2 --no-capture
+
+# channel or source
+TYPE ?= channel
```
algochoi, line 10: I think this should be removed here and only in the `.env` so there is one central place to control these vars

tzaffi: After playing around some more here, I'm convinced. The less places to overwrite, the easier it is to get it right and use.

tzaffi: I sent the change up.

algochoi (commented):
Looks good + working with the boxes! Thanks for the work here

tests/steps/application_v2_steps.py
```diff
@@ -413,14 +369,15 @@ def wait_for_algod_transaction_processing_to_complete():
     time.sleep(0.5)
 
 
+# TODO: this needs to be modified/removed when v1 is no longer supported
```
algochoi, line 372: This should be a v2 step, so I think it should still be kept after v1 is _fully_ deprecated.

Makefile
```diff
@@ -1,11 +1,129 @@
...
+pause:
+	cd test-harness \
+	&& docker-compose pause 
+
+unpause:
+	cd test-harness \
+	&& docker-compose unpause 
+
+
+# sdk-harness-algod \
```
algochoi, line 54: Is everything below this necessary? If we decide to include it, it might be nice to have a comment to explain what they do

algochoi (commented):
LGTM - minor comments

Makefile
tzaffi: OOPS.... didn't mean to stage/push these changes. These are some shortcuts I use locally. Will revert shortly.

tests/steps/application_v2_steps.py
```diff
@@ -413,14 +369,16 @@ def wait_for_algod_transaction_processing_to_complete():
     time.sleep(0.5)
 
 
+# TODO: this needs to be modified to use v2 only
 @step("I wait for the transaction to be confirmed.")
 def wait_for_app_txn_confirm(context):
     wait_for_algod_transaction_processing_to_complete()
     if hasattr(context, "acl"):
+        # TODO: get rid of this branch of logic when v1 fully deprecated
```
algochoi, line 377: 👍

algochoi (approved):
Looks good! Thanks for the changes

.test-env
```diff
@@ -0,0 +1,40 @@
...
+LOCAL_SANDBOX_DIR=".sandbox"
+
+# replacement values for Sandbox's docker-compose:
+ALGOD_CONTAINER=sdk-harness-algod
+KMD_PORT=60001
+ALGOD_PORT=60000
+INDEXER_CONTAINER=sdk-harness-indexer
+INDEXER_PORT=59999
+POSTGRES_CONTAINER=sdk-harness-postgres
+POSTGRES_PORT=65432
```
michaeldiamant, line 40: @tzaffi Leaving a _suggestion_ as a visual reminder following up on our verbal.  Prior to merging, my request is to remove algorand-sdk-testing overrides.

For the benefit of other reviewers:  While open to discussion, my view continues to be:
* `.env` in algorand-sdk-testing governs configuration across SDKs.  Consequently, I expect it'll be rare to merge `.test-env` to `develop` with algorand-sdk-testing overrides.
* I see local testing and/or one-off SDK PRs where `.test-env` overrides are a welcomed convenience. 
* On balance, I think the PR adds a useful + flexible facility that requires intentional usage to minimize misconfigurations.

```suggestion
```

test-harness.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ -d "$SDK_TESTING_HARNESS" ]; then
+  pushd "$SDK_TESTING_HARNESS"
+  ./scripts/down.sh
+  popd
+  rm -rf "$SDK_TESTING_HARNESS"
+else
+  echo "test-harness.sh: directory $SDK_TESTING_HARNESS does not exist - NOOP"
+fi
+
+git clone --single-branch --branch "$SDK_TESTING_BRANCH" "$SDK_TESTING_URL" "$SDK_TESTING_HARNESS"
```
michaeldiamant, line 25: @tzaffi Optionally, you may prefer to limit the cloned repo history as discussed in https://github.com/algorand/algorand-sdk-testing/pull/220#discussion_r946267733.  If you prefer no change, feel welcomed to resolve as is.

```suggestion
git clone --depth 1 --single-branch --branch "$SDK_TESTING_BRANCH" "$SDK_TESTING_URL" "$SDK_TESTING_HARNESS"
```

Makefile
```diff
@@ -1,11 +1,20 @@
-UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.sourcemap"
+UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.responses.unlimited_assets or @unit.sourcemap or @unit.tealsign or @unit.transactions or @unit.transactions.payment "
```
michaeldiamant, line 1: @tzaffi Is the trailing space deliberate or incidental?  Does not appear to affect execution.

```suggestion
UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.responses.unlimited_assets or @unit.sourcemap or @unit.tealsign or @unit.transactions or @unit.transactions.payment"
```

.test-env
```diff
@@ -0,0 +1,40 @@
+# Configs for python sdk docker:
+PYTHON_VERSION=${PYTHON_VERSION:-3.8}
```
michaeldiamant, line 2: @tzaffi Optional:  I'd prefer to _avoid_ exposing the Python version for the Docker build here.  If you prefer as is, feel welcomed to resolve.

Rationale:
* Since the repo defines a specific minimum Python version, I feel local customization shouldn't be encouraged.  From my perspective, exposing the environment variable gives the appearance that configuration is freely supported.
* To me, the name `PYTHON_VERSION` implies a broader scope than the reality.  In practice, it _only_ governs the Docker build and _not_ `setup.py`.  
  * If retained, I'd feel more comfortable renaming the variable to more closely reflect its scope.   

```suggestion
```

test-harness.sh
```diff
@@ -0,0 +1,46 @@
...
+
+## OVERWRITE incoming .env with .test-env
+cp "$ENV_FILE" "$SDK_TESTING_HARNESS"/.env
```
michaeldiamant, line 28: @tzaffi Sorry - During our earlier verbal, I misunderstood the relationship between `.test-env` and algorand-sdk-testing `.env`.  

I thought we had override if-absent capabilities.  I see several options:
* Implement a merge giving preference to algorand-sdk-testing.  Allows continued use of a single environment file.
  * At least 2 options exists:  https://stackoverflow.com/a/14017638 and https://stackoverflow.com/a/57691790. 
  * I tested the former and I'm convinced it produces a correct diff.  If used, I'd cite the link source because it'd otherwise take me some time to decipher usage.
* Partition `SDK_TESTING_*` into a separate environment file.  Avoids the merge at the expense of 2nd config file.

Based on the PR's spirit, it seems the merge option is preferred.

michaeldiamant: Additionally, assuming merge happens, then following line must be moved from earlier in the script to after the highlighted line.

```
echo "test-harness.sh: AFTER sourcing from $ENV_FILE. Build TYPE=$TYPE"
```

In its current location, `$TYPE` won't be populated.

.test-env
tzaffi: Will probably take the suggestion.

.test-env
tzaffi: Makes sense. Will verify.

test-harness.sh
tzaffi: more experimentation

michaeldiamant: A note I missed previously:  At least with the `awk` approach, assuming no further customization, in-place changes are unsupported.  So, I think you'll need a 3rd temporary file like:

```
awk -F= '!a[$1]++' "$ENV_FILE" "$SDK_TESTING_HARNESS"/.env.tmp
mv "$SDK_TESTING_HARNESS"/.env "$SDK_TESTING_HARNESS"/.env.tmp
```

.test-env
tzaffi: I've introduced a new variable `OVERWRITE_TESTING_ENVIRONMENT=0` which allows for overwriting, along with a warning to the user.

test-harness.sh
tzaffi: I hope it's okay with you. I've opted for a "KISS" approach. As I explained in the previous comment, we don't overwrite testing's .env's unless `OVERWRITE_TESTING_ENVIRONMENT=1`, but in that case, the onus is left to the developer to explicitly provide out all of the expected env vars. LMK if you are of the opinion that a more sophisticated approach as suggested above is still warranted.

.test-env
```diff
@@ -0,0 +1,8 @@
+# Configs for testing repo download:
+SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
+SDK_TESTING_BRANCH="trim-indexer"
+SDK_TESTING_HARNESS="test-harness"
+
+# WARNING: Be careful when turning on the next variable.
+# In that case you'll need to provide all variables expected by `algorand-sdk-testing`'s `.env`
+OVERWRITE_TESTING_ENVIRONMENT=0
```
tzaffi, line 8:
```suggestion
OVERWRITE_TESTING_ENVIRONMENT=0

```

test-harness.sh
```diff
@@ -0,0 +1,51 @@
...
+  rm -rf "$SDK_TESTING_HARNESS"
+else
+  echo "$THIS: directory $SDK_TESTING_HARNESS does not exist - NOOP"
+fi
+
+git clone --depth 1 --single-branch --branch "$SDK_TESTING_BRANCH" "$SDK_TESTING_URL" "$SDK_TESTING_HARNESS"
+
+
+if [[ $OVERWRITE_TESTING_ENVIRONMENT == 1 ]]; then
+  echo "$THIS: OVERWRITE downloaded $SDK_TESTING_HARNESS/.env with $ENV_FILE:"
```
michaeldiamant, line 31: @tzaffi Optional _suggestion_ - When debugging a local run, the term _downloaded_ briefly confused me.  If you prefer as is, feel welcomed to resolve without change.

```suggestion
  echo "$THIS: OVERWRITE replaced $SDK_TESTING_HARNESS/.env with $ENV_FILE:"
```

test-harness.sh
michaeldiamant: @tzaffi Resolving - The overwrite flag works for me.  Thanks for working through the feedback to find a path forward.

Aside:  Here's a briefly considered thought for a future iteration.
* If we add merge support, it'd simplify promoting the Cucumber test tags into algorand-sdk-testing.  I think the idea reduces the chance of an SDK failing to declare support for a set of tests.
* In the case where an SDK lags behind, the SDK _only_ overrides _only_ a single environment variable to declare its supported tests.
* I think there's still friction, but in my imagination, it'd be a net improvement over the status quo.

michaeldiamant (approved):
@tzaffi Excited to bring in these changes - appreciate the effort!  ☕

.test-env
```diff
@@ -0,0 +1,8 @@
+# Configs for testing repo download:
+SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
+SDK_TESTING_BRANCH="trim-indexer"
+SDK_TESTING_HARNESS="test-harness"
```
tzaffi, line 4: ^^^ DO NOT MERGE UNTIL `trim-indexer` above is merged. And then point to `master` instead.

Makefile
```diff
@@ -1,11 +1,21 @@
-UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.sourcemap"
+UNIT_TAGS :=  "$(subst :, or ,$(shell awk '{print $2}' tests/unit.tags | paste -s -d: -))"
+INTEGRATIONS_TAGS := "$(subst :, or ,$(shell awk '{print $2}' tests/integration.tags | paste -s -d: -))"
```
tzaffi, line 2:
```suggestion
INTEGRATION_TAGS := "$(subst :, or ,$(shell awk '{print $2}' tests/integration.tags | paste -s -d: -))"
```

Makefile
```diff
@@ -1,11 +1,21 @@
...
+
 unit:
 	behave --tags=$(UNITS) tests -f progress2
 
-INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey_v1 or @send.keyregtxn or @send or @compile.sourcemap"
 integration:
-	behave --tags=$(INTEGRATIONS) tests -f progress2
+	behave --tags=$(INTEGRATIONS) tests -f progress2 --no-capture
```
tzaffi, line 8:
```suggestion
	behave --tags=$(INTEGRATION_TAGS) tests -f progress2 --no-capture
```

Makefile
```diff
@@ -1,11 +1,21 @@
-UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.sourcemap"
+UNIT_TAGS :=  "$(subst :, or ,$(shell awk '{print $2}' tests/unit.tags | paste -s -d: -))"
+INTEGRATIONS_TAGS := "$(subst :, or ,$(shell awk '{print $2}' tests/integration.tags | paste -s -d: -))"
+
 unit:
 	behave --tags=$(UNITS) tests -f progress2
```
tzaffi, line 5:
```suggestion
	behave --tags=$(UNIT_TAGS) tests -f progress2
```

tzaffi (commented):
oopsies

Makefile
```diff
@@ -1,11 +1,24 @@
...
 unit:
-	behave --tags=$(UNITS) tests -f progress2
+	behave --tags=$(UNIT_TAGS) tests -f progress2
 
-INTEGRATIONS = "@abi or @algod or @applications or @applications.verified or @assets or @auction or @c2c or @compile or @dryrun or @dryrun.testing or @indexer or @indexer.231 or @indexer.applications or @kmd or @rekey_v1 or @send.keyregtxn or @send or @compile.sourcemap"
 integration:
-	behave --tags=$(INTEGRATIONS) tests -f progress2
+	behave --tags=$(INTEGRATION_TAGS) tests -f progress2 --no-capture
+
+display-all-python-steps:
```
tzaffi, line 10: generates text used by the **Unused Steps Analysis** script

Makefile
```diff
@@ -1,11 +1,24 @@
-UNITS = "@unit.abijson or @unit.abijson.byname or @unit.algod or @unit.algod.ledger_refactoring or @unit.applications or @unit.atc_method_args or @unit.atomic_transaction_composer or @unit.dryrun or @unit.dryrun.trace.application or @unit.feetest or @unit.indexer or @unit.indexer.ledger_refactoring or @unit.indexer.logs or @unit.offline or @unit.rekey or @unit.transactions.keyreg or @unit.responses or @unit.responses.231 or @unit.tealsign or @unit.transactions or @unit.transactions.payment or @unit.responses.unlimited_assets or @unit.sourcemap"
+UNIT_TAGS :=  "$(subst :, or ,$(shell awk '{print $2}' tests/unit.tags | paste -s -d: -))"
+INTEGRATION_TAGS := "$(subst :, or ,$(shell awk '{print $2}' tests/integration.tags | paste -s -d: -))"
```
tzaffi, line 2: cucumber tags now reside in:
* `tests/unit.tags`
* `tests/integration.tags`

.test-env
```diff
@@ -0,0 +1,14 @@
+# Configs for testing repo download:
+SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
+SDK_TESTING_BRANCH="trim-indexer"
```
michaeldiamant, line 3: Leaving a friendly visual reminder to reset the branch prior to merge.

```suggestion
SDK_TESTING_BRANCH="master"
```

michaeldiamant (approved):
Approval assumes https://github.com/algorand/py-algorand-sdk/pull/367#discussion_r951583410 is addressed prior to merge.

PR #368, opened 2022-08-10
bug-fix: add check to desc so we dont output null if undefined

barnjamin:
This seems to choke the js-sdk which is expecting `string | undefined`

When there is a `None` in the dict, the json dumped includes `null`

```
>>> x = {'z':None}
>>> import json
>>> json.dumps(x)
'{"z": null}'
```

Other dictify methods have checks for this (Method/Args)

### DIFF
--- a/algosdk/abi/contract.py
+++ b/algosdk/abi/contract.py
@@ -47,8 +47,9 @@ class Contract:
         d = {}
         d["name"] = self.name
         d["methods"] = [m.dictify() for m in self.methods]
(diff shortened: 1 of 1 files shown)


algochoi (approved):
Thanks for catching this - it looks like we (me) added this check for the [method](https://github.com/algorand/py-algorand-sdk/blob/cb93e8969a52e85fc92b7c58a43ca39776b1d313/algosdk/abi/method.py#L116) and interface but not for the contract.

PR #370, opened 2022-08-19
Enhancement: Add State Proof support

shiqizng:
Add support for new state proof endpoints and transaction deserialization.


tests/unit_tests/test_other.py
```diff
@@ -248,6 +248,170 @@ def test_keyreg_txn_nonpart(self):
             encoding.msgpack_encode(encoding.future_msgpack_decode(keyregtxn)),
         )
 
+    def test_stateproof_txn(self):
```
winder, line 251: Is this one of the msgpack tests that the python cucumber test doesn't support?

shiqizng: yes, I extracted the txn object from v2_indexer_lookupBlock_transaction.json from encoded it using msgpacktool.

algosdk/v2client/algod.py
```diff
@@ -397,16 +397,58 @@ def genesis(self, **kwargs):
...
-        return self.algod_request("GET", req, **kwargs)
+        return self.algod_request(
+            "GET",
+            req,
+            params=params,
+            response_format=response_format,
+            **kwargs
+        )
+
+    def lightheader(self, round_num, response_format="json", **kwargs):
```
algoidan, line 425: maybe we should rename it to `lightheader_proof`, `lightblockheader_proof` or something similar

algosdk/v2client/algod.py
```diff
@@ -397,16 +397,58 @@ def genesis(self, **kwargs):
...
+    def transaction_proof(
+        self, round_num, txid, hashtype="", response_format="json", **kwargs
+    ):
         """
         Get the proof for a given transaction in a round.
 
         Args:
             round_num (int): The round in which the transaction appears.
             txid (str): The transaction ID for which to generate a proof.
+            hashtype (str): The state proof hash type, sha256 or sha512_256.
```
algoidan, line 409:
```suggestion
            hashtype (str): The type of hash function used to create the proof, must be one of: 
* sha512_256 
* sha256
```

algosdk/v2client/algod.py
```diff
@@ -397,16 +397,58 @@ def genesis(self, **kwargs):
...
+            response_format=response_format,
+            **kwargs
+        )
+
+    def lightheader(self, round_num, response_format="json", **kwargs):
+        """
+        Gets a proof for a given light block header inside a state proof commitment.
+
+        Args:
+            round_num (int): The round in which the transaction appears.
```
algoidan, line 430: this is not really a transaction, the light block header is a subset of the original block header.

algosdk/future/transaction.py
```diff
@@ -2906,6 +2909,103 @@ def __eq__(self, other):
...
+        self.sprf = state_proof
+        self.sprfmsg = state_proof_message
+
+    def dictify(self):
+        d = dict()
+        if self.sprf_type:
+            d["sptype"] = self.sprf_type
+        if self.sprfmsg:
+            d["spmsg"] = self.sprfmsg
+        d["sp"] = self.sprf
```
algoidan, line 2978: Is there a reason why we don't check if self.sprf

algosdk/future/transaction.py
```diff
@@ -2906,6 +2909,103 @@ def __eq__(self, other):
...
+    """
+    Represents a state proof transaction
+
+    Arguments:
+        sender (str): address of the sender
+        state_proof (dict())
+        state_proof_message (dict(), optional)
+        state_proof_type (str, optional): state proof type
+        sp (SuggestedParams): suggested params from algod
+        receiver (str): address of the receiver
```
algoidan, line 2922: Basically, a state proof transaction should not have any fee, note, rekeying field, or lease. I don't think we should assert that, but maybe just change the comments?

algosdk/v2client/algod.py
```diff
@@ -397,16 +397,58 @@ def genesis(self, **kwargs):
...
+        return self.algod_request(
+            "GET", req, response_format=response_format, **kwargs
+        )
+
+    def stateproofs(self, round_num, response_format="json", **kwargs):
+        """
+        Get a state proof that covers a given round.
+
+        Args:
+            round_num (int): The round in which the transaction appears.
```
algoidan, line 444: This is not completely accurate. The API would search for a state proof that covers the given round. The state proof txn would not appear there.

algosdk/future/transaction.py
shiqizng: I was thinking that since it's a stateproof txn type, state_roof can't be None.

algosdk/v2client/algod.py
shiqizng: updated the description.

AlgoStephenAkiki (approved):

algosdk/future/transaction.py
algoidan: This is how the stateproof txn is defined on go-algorand
https://github.com/algorand/go-algorand/blob/7de989aa62a01ce9971b48cb1c2ca6cc2f18f3bf/data/transactions/stateproof.go#L31

Looks like the type, the message, and the proof can't be None (not pointers). I think they should be referenced in the same way (i.e should remove the if for all of them I guess)

algosdk/future/transaction.py
```diff
@@ -260,6 +260,11 @@ def undictify(d):
         elif txn_type == constants.appcall_txn:
             args.update(ApplicationCallTxn._undictify(d))
             txn = ApplicationCallTxn(**args)
+        elif txn_type == constants.stateproof_txn:
+            # a state proof txn does not have these fields
+            args.pop("note"), args.pop("rekey_to"), args.pop("lease")
```
algoidan, line 265: should we remove 'group' as well?

algosdk/future/transaction.py
```diff
@@ -2906,6 +2911,83 @@ def __eq__(self, other):
...
+            args["state_proof_message"] = d["spmsg"]
+
+        return args
+
+    def __eq__(self, other):
+        if not isinstance(other, StateProofTxn):
+            return False
+        return (
+            super(StateProofTxn, self).__eq__(other)
+            and self.sprf_type == other.state_proof_type
```
algoidan, line 2983: I guess I'm getting it wrong but is other the same object as self? i.e shouldn't that be and self.sprf_type == other.sprf_type

shiqizng: you're right. I'm fixing it.

algosdk/future/transaction.py
shiqizng: the struct has `omitempty,omitemptyarray`, I thought some of these fields could be empty.

algoidan (approved):

PR #380, opened 2022-09-06
Enhancement: Set default value of wait_rounds in the function definition

efe:


algochoi (dismissed):
Thanks for the change!

jasonpaulos:
Let's hold on merging this. This change removes the behavior that passing in 0 is supposed to mean wait as long as possible (1000 rounds is an upper bound on the maximum wait time for a pending txn).

We need to investigate whether changing this would be detrimental or not.

algochoi:
> Let's hold on merging this. This change removes the behavior that passing in 0 is supposed to mean wait as long as possible (1000 rounds is an upper bound on the maximum wait time for a pending txn).
> 
> We need to investigate whether changing this would be detrimental or not.

Good point - this is potentially a breaking change

efe:
It waits `1000` rounds if `0` is passed as `wait_rounds`. I think that this is a clear unintended behaviour.

IMO, This is not an enhancement or a breaking change. It is a bug-fix.

jasonpaulos:
The fact that the function treats 0 as 1000 is intentional. More background here: https://github.com/algorand/py-algorand-sdk/pull/263/files#r764285721

I agree that this default value is not optimal, but changing it would introduce a breaking change, which we should avoid until the next major version of this SDK.

Perhaps at this point the best thing we can do is to improve the function's documentation.

algosdk/future/transaction.py
```diff
@@ -3181,9 +3184,6 @@ def wait_for_confirmation(
     last_round = algod_client.status()["last-round"]
     current_round = last_round + 1
 
-    if wait_rounds == 0:
-        wait_rounds = 1000
-
```
michaeldiamant, line 3186: @jasonpaulos Do you consider the following proposal a worthwhile improvement that retains backwards compatibility?

* Restore the highlighted, deleted lines so that callers supplying `wait_rounds = 0` retain the existing behavior + leave rest of PR as is.
* Optionally, add a brief inline comment explaining the behavior.

Otherwise, I agree the PR proposes a backwards incompatible change.

jasonpaulos: So are you suggesting to _only_ change the default value of the `wait_rounds` parameter from 0 to 1000? That seems fine to me.

Independently of that change, I agree we should document _why_ the special case for 0 must exist, and explain this in the docstring as well.

michaeldiamant: > you suggesting to only change the default value of the wait_rounds parameter from 0 to 1000? 

@jasonpaulos Correct.

@efe Provided you're on-board too, can you make the change + add a docstring remark explaining the significance of the 0 value?

bbroder-algo:
can re-evaluate on v3

PR #371, opened 2022-08-19
Enhancement: Deprecating use of langspec 

ahangsu:

PR #381, opened 2022-09-06
Deprecation: Add deprecation warnings on v1 algod API and old transaction format

algochoi:
This PR adds deprecation warnings to v1 algod APIs and the old transaction format. In addition, there are warnings raised when the deprecated methods are called. 

The PR also adds a deprecation comment to `template` - previously, a similar comment was put in for `future.template`.

The `DeprecationWarning` that is added to this PR and the [langspec PR](https://github.com/algorand/py-algorand-sdk/pull/371/files) does trigger warnings on pytest (`pytest tests/unit_tests`). The corresponding tests should also be deleted when these features are removed from the SDK.

Related issue: https://github.com/algorand/algorand-sdk-testing/issues/218

### DIFF
--- a/algosdk/encoding.py
+++ b/algosdk/encoding.py
@@ -1,8 +1,11 @@
 import base64
-import msgpack
+import warnings
 from collections import OrderedDict
+
+import msgpack
 from Cryptodome.Hash import SHA512
-from . import transaction, error, auction, constants, future
+
+from . import auction, constants, error, future, transaction
 
 
 def msgpack_encode(obj):
@@ -94,6 +97,9 @@ def future_msgpack_decode(enc):
 
 def msgpack_decode(enc):
     """
+    NOTE: This method is deprecated:
+    Please use `future_msgpack_decode` instead.
+
     Decode a msgpack encoded object from a string.
 
     Args:
@@ -103,6 +109,11 @@ def msgpack_decode(enc):
         Transaction, SignedTransaction, Multisig, Bid, or SignedBid:\
             decoded object
     """
+    warnings.warn(
+        "`msgpack_decode` is being deprecated. "
+        "Please use `future_msgpack_decode` instead.",
+        DeprecationWarning,
+    )
     decoded = enc
     if not isinstance(enc, dict):
         decoded = msgpack.unpackb(base64.b64decode(enc), raw=False)
--- a/algosdk/future/transaction.py
+++ b/algosdk/future/transaction.py
@@ -1,18 +1,16 @@
-from typing import List, Union
 import base64
 import binascii
-from enum import IntEnum
-import msgpack
+import warnings
 from collections import OrderedDict
-from .. import account
-from .. import constants
-from .. import encoding
-from .. import error
-from .. import logic
-from .. import transaction
-from ..v2client import algod, models
-from nacl.signing import SigningKey, VerifyKey
+from enum import IntEnum
+from typing import List, Union
+
+import msgpack
 from nacl.exceptions import BadSignatureError
+from nacl.signing import SigningKey, VerifyKey
+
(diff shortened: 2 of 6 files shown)


algosdk/transaction.py
```diff
@@ -52,11 +58,19 @@ def __init__(
...
         """
+        NOTE: This method is deprecated:
+        Please use the v2 equivalent in `future.transaction` instead.
+
         Get the transaction's ID.
 
         Returns:
             str: transaction ID
         """
+        warnings.warn(
```
algochoi, line 69: Adding warnings to each method might be pedantic - in `algod.py`, only the constructor has a `DeprecationWarning` triggered.

algosdk/template.py
```diff
@@ -6,6 +6,9 @@
 
 
 class Template:
+    """
+    NOTE: This class is deprecated
```
algochoi, line 10: note: `future.template` also has a deprecation warning in the comments.

algosdk/future/transaction.py
```diff
@@ -273,6 +271,13 @@ def undictify(d):
     def __eq__(self, other):
         if not isinstance(other, (Transaction, transaction.Transaction)):
             return False
+        if isinstance(other, transaction.Transaction):
```
algochoi, line 274: Added a warning for people who might try to compare the old `Transaction` object to a `future.Transaction` object

algosdk/transaction.py
```diff
@@ -1632,7 +2107,21 @@ def retrieve_from_file(path):
...
     def __init__(self, txns):
+        """
+        NOTE: This method is deprecated:
+        Please use the v2 equivalent in `future.transaction` instead.
+        """
+        warnings.warn(
+            "`TxGroup` is a part of v1 `transaction` that is being deprecated. "
+            "Please use the v2 equivalent in `future.transaction` instead.",
+            DeprecationWarning,
+        )
```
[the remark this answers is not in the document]
ahangsu: nvm, I got it wrong.

ahangsu (commented):
generally looking good to me

algosdk/transaction.py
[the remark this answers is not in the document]
algochoi: I think it's an interesting point - `v1` and `v2` markings might not be necessary so I decided to remove them in the latest commit

algosdk/transaction.py
jasonpaulos: I agree warnings in each method is likely too much. If there's a warning in the constructor, that should be good enough in my opinion

jasonpaulos (commented):
Mostly good, but I agree with your earlier comment that we may want to scale back the warnings in transaction methods

algosdk/encoding.py
```diff
@@ -103,6 +109,11 @@ def msgpack_decode(enc):
         Transaction, SignedTransaction, Multisig, Bid, or SignedBid:\
             decoded object
     """
+    warnings.warn(
+        "`msgpack_decode` is being deprecated. "
+        "Please use `future_msgpack_decode` instead.",
+        DeprecationWarning,
+    )
```
ahangsu, line 116: I don't have much objection here, just a minor question: since encode-decode is coupled, do we want to mark encode as deprecated? Or it is because of `Transaction` had a structural change (i.e., some new fields)?

ahangsu (approved):
lgtm

algosdk/encoding.py
algochoi: Serializing into `msgpack` is okay - the old `Transaction` and new `Transaction` can both be serialized into msgpack as long as the fields are ordered and can be represented by a key-value map/dict. Decoding msgpack into the old `Transaction` does not seem preferable, so marked it with the deprecation warning here.

jasonpaulos (approved):

PR #390, opened 2022-10-06
REST API: Add algod block hash endpoint, add indexer block header-only param.

winder:
Add tests to new REST API code.


tzaffi (approved):
LGTM

PR #396, opened 2022-11-08
Bug-fix: Implement `TransactionRejectedError`

jdtzmn:
This PR implements the `TransactionRejectedError` which was referenced within the logic of the `wait_for_confirmation` method despite being undefined.

This is a small PR to quickly fix the issue at hand, however future work is necessary to ensure that this class of errors is caught moving forward.


michaeldiamant (approved):
@jdtzmn Thanks for the quick fix here - let's use mypy enablement as a way to sanity check correctness.

PR #403, opened 2022-12-01
examples: Migrate v1 algod usage to v2 algod

algochoi:
This PR changes the `examples` to use the `v2client` instead of v1 algod. 

As a sanity check, ran the examples on my local machine.

### DIFF
--- a/examples/log_sig_example.py
+++ b/examples/log_sig_example.py
@@ -1,13 +1,14 @@
 # Example: creating a LogicSig transaction signed by a program that never approves the transfer.
 
 import tokens
-from algosdk import algod, account
+from algosdk import account
+from algosdk.v2client import algod
 from algosdk.future import transaction
 
 program = b"\x01\x20\x01\x00\x22"  # int 0
 lsig = transaction.LogicSigAccount(program)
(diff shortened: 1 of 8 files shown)


examples/log_sig_example.py
```diff
@@ -1,13 +1,14 @@
 # Example: creating a LogicSig transaction signed by a program that never approves the transfer.
 
 import tokens
-from algosdk import algod, account
+from algosdk import account
+from algosdk.v2client import algod
```
michaeldiamant, line 5: @algochoi Minor:  Unless opposed, can you update the PR to be sandbox-compatible?

I think you raised a good point in our offline chat.  I suspect others are more likely to test via sandbox as a default.  

It feels to me like a minimally involved improvement to _examples_.  With the understanding that a more in-depth treatment is out-of-scope.

algochoi: That sounds like a good improvement - I tried addressing this in [aafcf35](https://github.com/algorand/py-algorand-sdk/pull/403/commits/aafcf352914e2b210ec3aeedfe44f85da4007ca2)

PR #406, opened 2022-12-03
Fix populate_foreign_array offset logic

### LINKED ISSUE #405: populate_foreign_array behaves unexpectedly during application creation tx
### Subject of the issue

In atomic_transaction_composer.py when adding methods to AtomicTransactionComposer the function populate_foreign_array  is called to add the appropriate app args to the tx.
`
offset = 0 if not zero_value else 1
`

`If not <var>` will return True if the var is None or 0. Offset should only be used for foreign apps or foreign accounts, which should always return a number whether it is the app id or 0. None will only return for assets which will be None.

### Your environment
`algod` 3.12.2.stable [rel/stable] (commit #181490e3)
`py-algorand-sdk` 1.20.1
`pyteal`, https://github.com/algorand/pyteal/commit/de90e93e07a28f162a3b5fe2174889c05b73d509

### Steps to reproduce

1. Use AtomicTransactionComposer and add_method_call to create an ABI application using foreign apps as args.

### Expected behaviour
The intended behavior was to offset the app args by one for foreign accounts and applications due to zero value being set to either the sender or the created app ID.
`
offset = 0 if not zero_value else 1
`

### Actual behaviour
While using AtomicTransactionComposer during a creation tx, populate_foreign_array is called and sets the app_id to 0 which will also trigger True for the offset logic. If not will return True for both None and 0 values.
`
if not zero_value
`

This causes the app args ints referencing foreign apps will always be 1 less than what they should be.

JGomezST:
Resolves #405

### DIFF
--- a/algosdk/atomic_transaction_composer.py
+++ b/algosdk/atomic_transaction_composer.py
@@ -59,7 +59,7 @@ def populate_foreign_array(
     if zero_value and value_to_add == zero_value:
         return 0
 
-    offset = 0 if not zero_value else 1
+    offset = 0 if zero_value is None else 1
 
     if value_to_add in foreign_array:
         return foreign_array.index(value_to_add) + offset


algosdk/atomic_transaction_composer.py
```diff
@@ -59,7 +59,7 @@ def populate_foreign_array(
     if zero_value and value_to_add == zero_value:
```
algochoi, line 59:
```suggestion
    if zero_value is not None and value_to_add == zero_value:
```

algochoi (commented):
Thanks for the fix! I agree this looks like a bug when the supplied app_id is 0 for an earlier created transaction.

algosdk/atomic_transaction_composer.py
jgomezst: Yeah this looks good, didn't notice that. Its usually a good idea to specifically check for None and 0 values.

algochoi (approved):
Thanks! The changes look good to merge after merging the SDK test case: https://github.com/algorand/algorand-sdk-testing/pull/253

Issue #384, opened 2022-09-12
Breaking changes in v2.0.0


michaeldiamant:
Itemizes functionality that will be removed in the upcoming major release.

Task list for algorand-sdk-testing stories:
- [x] https://github.com/algorand/algorand-sdk-testing/issues/201 - https://github.com/algorand/py-algorand-sdk/pull/412
- [x] https://github.com/algorand/algorand-sdk-testing/issues/202 - N/A no removal needed 
- [x] https://github.com/algorand/algorand-sdk-testing/issues/223 - https://github.com/algorand/py-algorand-sdk/pull/411
- [x] https://github.com/algorand/algorand-sdk-testing/issues/254 - https://github.com/algorand/py-algorand-sdk/pull/410
- [x] https://github.com/algorand/algorand-sdk-testing/issues/255 - https://github.com/algorand/py-algorand-sdk/pull/418
- [x] https://github.com/algorand/generator/issues/46 - N/A no generated source code

algochoi:
Let's consider changing `wait_rounds` to match the other SDKs - when the value is zero, the other SDKs wait zero rounds instead of 1000 (e.g. [JS](https://github.com/algorand/js-algorand-sdk/blob/develop/src/wait.ts)) 

Related PR: https://github.com/algorand/py-algorand-sdk/pull/380

PR #410, opened 2022-12-06
Remove logicsig templates from SDKs for v2.0.0 breaking changes

ahangsu:
From https://github.com/algorand/go-algorand-internal/issues/1415, we have already removed algorand-sdk-testing integration part for template.

We just need to remove the `template.py` file.

Part of epic https://github.com/algorand/py-algorand-sdk/issues/384.


ahangsu:
good call out, I just take all of them out

algochoi (commented):
Apologies for not noticing earlier:
* I think there are some errors that are only being used in templates, so we can safely delete them: `WrongContractError`, `TemplateInputError`, `TemplateError`
* There are references to the templates in the `docs` directory. I believe these are manually edited, so we can probably delete them as well?

ahangsu:
I just removed all the unused errors and docs, maybe this applies to other deprecation PRs, we will see...

docs/algosdk/future/index.rst
```diff
@@ -3,6 +3,5 @@ future
 .. toctree::
    :maxdepth: 10
 
-   template
```
algochoi, line 6: I think there is also a reference in `algosdk/index.rst`: https://github.com/algorand/py-algorand-sdk/blob/develop/docs/algosdk/index.rst

ahangsu: 👍

algochoi (approved):
LGTM!

PR #415, opened 2022-12-06
v2: Breaking changes from v1 to v2.0.0

ahangsu:
This PR contains all (breaking) changes to be made for v1 to v2.0.0 SDK, epic here: #384

PR #416, opened 2022-12-07
Merge develop into feature/v2.0.0 break

michaeldiamant:
Merges #397 into #415.

Notes:
* There's a large number of diffs due to the introduction of mypy.  I'm trusting a passing build covers the bases.
* Since it's a merge, PR should _not_ be squash merged.


ahangsu (approved):
👍

PR #420, opened 2022-12-13
New Feature: Adding methods to use the simulate endpoint

barnjamin:
Adding Simulate endpoint support to algod client

Also added simulate as a method on the AtomicTransactionComposer to allow an api like:
image


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="simulate-endpoint-tests"
```
barnjamin, line 3: remember to change this back

tests/steps/other_v2_steps.py
```diff
@@ -1414,3 +1418,33 @@ def transaction_proof(context, round, txid, hashtype):
...
+def simulate_transaction(context):
+    context.simulate_response = context.app_acl.simulate_transactions(
+        [context.stx], response_format="json"
+    )
+
+
+@then("the simulation should succeed")
+def simulate_transaction_succeed(context):
+    if hasattr(context, "simulate_response"):
+        assert context.simulate_response["would-succeed"] is True
```
bbroder-algo, line 1433: should succeed or would succeed

barnjamin: I'm not sure I follow?

It's `would-succeed` in the api 
https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas2.json#L4044

algosdk/atomic_transaction_composer.py
```diff
@@ -479,6 +479,55 @@ def submit(self, client: algod.AlgodClient) -> List[str]:
...
+        if self.status <= AtomicTransactionComposerStatus.SUBMITTED:
+            self.gather_signatures()
+        else:
+            raise error.AtomicTransactionComposerError(
+                "AtomicTransactionComposerStatus must be submitted or "
+                "lower to simulate a group"
+            )
+
+        simulation_result: Dict[str, Any] = client.simulate_transactions(
+            self.signed_txns, response_format="json"
```
algochoi, line 495: I *think* `json` is the default for every algod call in this SDK so we don't need to define this: https://github.com/algorand/py-algorand-sdk/blob/882052efa331ceafa5d0afd695bb8c8a426efc6d/algosdk/v2client/algod.py#L39

tests/steps/other_v2_steps.py
```diff
@@ -1414,3 +1418,33 @@ def transaction_proof(context, round, txid, hashtype):
 @when("we make a Lookup Block Hash call against round {round}")
 def get_block_hash(context, round):
     context.response = context.acl.get_block_hash(round)
+
+
+@when("I simulate the transaction")
+def simulate_transaction(context):
+    context.simulate_response = context.app_acl.simulate_transactions(
+        [context.stx], response_format="json"
```
algochoi, line 1426: nit: I don't think we need to specify `json` for this SDK since it is the default

tests/steps/other_v2_steps.py
algochoi: sorry I also changed the step names upstream: https://github.com/algorand/algorand-sdk-testing/pull/265/files#diff-15dc2e739dbef5460705db37e5398ba985224f5ba449daa0a32dfdaaf1dfad49R30

current format: 
```    
    the simulation should succeed without any failure message
    And the simulation should report a failure at path "" with message ""
    ```

File: tests/steps/other_v2_steps.py [hunk 5]
```diff
@@ -1414,3 +1418,33 @@ def transaction_proof(context, round, txid, hashtype):
...
+    context.simulate_atc_response = (
+        context.atomic_transaction_composer.simulate(context.app_acl)
+    )
+
+
+@then('the simulation should fail at path "{path}" with message "{message}"')
+def simulate_atc_failure(context, path, message):
+    resp: SimulateAtomicTransactionResponse = context.simulate_atc_response
+    assert resp.would_succeed is False
+    assert resp.failed_at[0] == int(path)
```
algochoi, line 1449: I changed the upstream step so path can now be a comma separated string "2,0". This basically means a failure in the third transaction, first inner

algochoi (commented):
Sorry for some of the sudden upstream SDK step-related changes - I commented the places where you might need to update it

algosdk/atomic_transaction_composer.py
```diff
@@ -479,6 +479,55 @@ def submit(self, client: algod.AlgodClient) -> List[str]:
         self.status = AtomicTransactionComposerStatus.SUBMITTED
         return self.tx_ids
 
+    def simulate(
+        self, client: algod.AlgodClient
+    ) -> "SimulateAtomicTransactionResponse":
```
tzaffi, line 484: nit: If you define `class SimulateAtomicTransactionResponse` above `AtomicTransactionComposer`, you should be able to refer to it without these quotes.

algosdk/atomic_transaction_composer.py
```diff
@@ -479,6 +479,55 @@ def submit(self, client: algod.AlgodClient) -> List[str]:
...
+
+        if self.status <= AtomicTransactionComposerStatus.SUBMITTED:
+            self.gather_signatures()
```
tzaffi, line 487: I thought you could simulate without signing

algosdk/v2client/algod.py
```diff
@@ -483,6 +483,47 @@ def get_block_hash(self, round_num, **kwargs):
         req = "/blocks/{}/hash".format(round_num)
         return self.algod_request("GET", req, **kwargs)
 
+    def simulate_transactions(self, txns, **kwargs):
```
tzaffi, line 486: nit: type-hints would be nice for the `simulate_*` . Even though the rest of the file isn't conforming here, it should be okay to have new methods with annotations and mypy should be able to handle it fine (though you may need to sprinkle some `# type: ignore`'s in the method bodies.

algosdk/atomic_transaction_composer.py
barnjamin: I think you can, but should I make this just build group? If they're using ATC they have to pass _some_ signer anyway

algosdk/atomic_transaction_composer.py
barnjamin: I tried to make it match the other methods here, should I do this and also move the stuff returned from `execute` as well?

algosdk/v2client/algod.py
barnjamin: Can we make an issue to fully type hint this? feels weird to do just the one method here

tzaffi: I support that idea! Shall I create a child PR, or you want to take this?

algosdk/atomic_transaction_composer.py
tzaffi: I'm not sure. I'm trying to imagine what a dry-run like experience should look like with the new endpoint. We want to enable a path that involves minimal friction. Currently ATC's `add_transaction()` expects a `TransactionWithSigner`. Perhaps solving the problem is as simple as introducing a new `TransactionSigner` sub-class, something like `SimulateSkipSigner` which instead of actually signing leaves a marker that is later picked up by `gather_signatures` and the other downstream ATC methods which let the transaction through in the case of `/simulate`.

I'm leery about being too ambitious in this PR so I don't want to push too hard on this front for this review.

So probably it's best to build out the MVP for supporting simulate, but now is also a good time to create a design doc for making simulate as easy to use as dry run.

algosdk/v2client/algod.py
barnjamin: have at it! thank you!

algosdk/atomic_transaction_composer.py
barnjamin: A Dummy or PlaceholderSigner would probably make a good addition to all the SDKs, I've got one here for when I don't yet have any signers available for example https://github.com/algorand-devrel/hello-beaker/blob/master/src/App.tsx#L11-L17

algosdk/atomic_transaction_composer.py
barnjamin: @tzaffi thumbs up mean do it?

It'll need to be all of the classes below AtomicTransactionComposer though so will be a large diff

algosdk/atomic_transaction_composer.py
bbroder-algo: I'm in favor of a `EmptySigner` subclass now that submitting blank signature transactions is acceptable.

algosdk/atomic_transaction_composer.py
tzaffi: I'm not bothered by a large diff here, but most people probly are. We can kick this can as a follow up that simply re-arranges the order as it's own PR.

algosdk/v2client/algod.py
tzaffi: Oh well. This dragged in too many un-related changes. So I ended up making a PR into `develop`:

https://github.com/algorand/py-algorand-sdk/pull/442

I can also add in the ATC re-ordering so we don't have to stringify types (see another comment above)

algosdk/atomic_transaction_composer.py
barnjamin: https://github.com/algorand/py-algorand-sdk/issues/443 Filed an issue to track this one

tests/steps/other_v2_steps.py
```diff
@@ -1414,3 +1418,44 @@ def transaction_proof(context, round, txid, hashtype):
...
+    )
+
+
+@then(
+    'the simulation should report a failure at group "{group}", path "{path}" with message "{message}"'
+)
+def simulate_atc_failure(context, group, path, message):
+    resp: SimulateAtomicTransactionResponse = context.simulate_atc_response
+    group_idx: int = int(group)
+    fail_path = ",".join(
```
algochoi, line 1451: nit: 
```suggestion
    fail_path: str = ",".join(
```

algosdk/atomic_transaction_composer.py
```diff
@@ -517,20 +582,33 @@ def execute(
         self.status = AtomicTransactionComposerStatus.COMMITTED
 
         confirmed_round = resp["confirmed-round"]
-        method_results = []
 
-        for i, tx_id in enumerate(self.tx_ids):
+        tx_results = [
+            client.pending_transaction_info(tx_id) for tx_id in self.tx_ids
```
algochoi, line 587: caveat/nit: I don't think we need to get the txn info for every transaction, just the ones with method calls (`self.method_dict`). If the call overhead isn't too bad, we could keep as written in the current PR.

algochoi (commented):
Looks good 👍  minor comments and ready to approve when SDK testing PR is merged and the branch is reverted back to master.

.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="simulate-unsigned-txn"
```
algochoi, line 3:
```suggestion
SDK_TESTING_BRANCH="simulate-endpoint-tests"
```

algosdk/atomic_transaction_composer.py
```diff
@@ -601,6 +676,19 @@ def sign_transactions(
...
 
 
+class EmptySigner(TransactionSigner):
+    def __init__(self) -> None:
+        super().__init__()
+
+    def sign_transactions(
+        self, txn_group: List[transaction.Transaction], indexes: List[int]
+    ) -> List[GenericSignedTransaction]:
+        stxns = []
```
bbroder-algo, line 686:
```python
- stxns = []
+ stxns: List[GenericSignedTransaction] = []

bbroder-algo: the linter is complaining `error: Incompatible return value type`

.test-env
bbroder-algo: i looked at the trap ray

bbroder-algo: https://github.com/algorand/py-algorand-sdk/pull/445

PR #423, opened 2022-12-14
Keep `LogicSig` class, more detailed description in `LogicSig` comments

ahangsu:
Part of #384, mark deprecated `LogicSig` as hidden for implementations based on it.

Merge into #415

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@
 
 ### Breaking Changes
 
-* Removed `mnemonic.to_public_key` in favor of `account.address_from_private_key`.
-* Removed `algosdk.future` package.  Moved package contents to `algosdk`.
-* Removed `encoding.future_msgpack_decode` method in favor of `encoding.msgpack_decode` method.
+* Remove v1 algod API (`algosdk/algod.py`) due to API end-of-life (2022-12-01).  Instead, use v2 algod API (`algosdk/v2client/algod.py`).
+* Remove `algosdk.future` package.  Move package contents to `algosdk`.
+* Remove `encoding.future_msgpack_decode` method in favor of `encoding.msgpack_decode` method.
+* Remove `cost` field in `DryrunTxnResult` in favor of 2 fields:  `budget-added` and `budget-consumed`.  `cost` can be derived by `budget-consumed - budget-added`.
+* Remove `mnemonic.to_public_key` in favor of `account.address_from_private_key`.
+* Remove logicsig templates, `algosdk/data/langspec.json` and all methods in `logic` depending on it.
 * Mark `LogicSig` class as hidden in favor of internal use, renaming it to `_LogicSig`.
 
 # v1.20.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 * Remove `cost` field in `DryrunTxnResult` in favor of 2 fields:  `budget-added` and `budget-consumed`.  `cost` can be derived by `budget-consumed - budget-added`.
 * Remove `mnemonic.to_public_key` in favor of `account.address_from_private_key`.
 * Remove logicsig templates, `algosdk/data/langspec.json` and all methods in `logic` depending on it.
-* Mark `LogicSig` class as hidden in favor of internal use, renaming it to `_LogicSig`.
 
 # v1.20.2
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Removed `mnemonic.to_public_key` in favor of `account.address_from_private_key`.
 * Removed `algosdk.future` package.  Moved package contents to `algosdk`.
 * Removed `encoding.future_msgpack_decode` method in favor of `encoding.msgpack_decode` method.
+* Mark `LogicSig` class as hidden in favor of internal us, renaming it to `_LogicSig`.
```
michaeldiamant, line 12: @ahangsu I _think_ there's a minor typo here.  I left a _suggestion_ with what I believe to be the intended phrasing.

```suggestion
* Mark `LogicSig` class as hidden in favor of internal use, renaming it to `_LogicSig`.
```

algosdk/transaction.py
```diff
@@ -2503,7 +2503,7 @@ def __eq__(self, other):
         )
 
 
-class LogicSig:
+class _LogicSig:
```
michaeldiamant, line 2506: @ahangsu + reviewers - The following remark might be off-base.  I feel like I don't have a good handle (yet) on `LogicSig` vs `LogicSigAccount` use cases.

`_LogicSig` is exposed in several public methods/fields (e.g. `LogicSigAccount.lsig`, `LogicSigTransaction.__init__`).  Consequently:
* I don't feel confident marking the class private is sufficient.  Though I'm open to being told I'm off-base.
* I _think_ we'd need to remove all public references to `_LogicSig`.  Since I don't feel I understand the usage patterns, I'm unsure how palatable the change feels.

jasonpaulos: I agree that `LogicSig` should not be hidden/removed from the SDK. I think deprecating the whole class was too coarse grain. We need `LogicSig` to represent the logic signature structure for a transaction, i.e. `LogicSigTransaction.lsig`.

However, we don't need need to make it possible to directly construct `LogicSig`s now that we have `LogicSigAccount`. Perhaps functions like `LogicSig.sign` and `LogicSig.append_to_multisig` can be deprecated, hidden, or removed.

michaeldiamant: I still feel fuzzy and may benefit from a verbal - Here's my understanding:
* `LogicSig` is still safe to use for _some_ use cases (e.g. escrow).
* `LogicSig` is _not_ fully compatible with all delegated logicsig use cases (e.g. rekeying like in https://github.com/algorand/go-algorand-sdk/pull/230).  Here, `LogicSigAccount` fills the gap and additionally supports all `LogicSig` use cases.

If the above statements are accurate, then I think:
* I still don't know enough to recommend alternatives.  I imagine there's alternate object modelings that better represent desired behaviors + prevent undesired behaviors (like using `LogicSig` in an incompatible way).  
* I'd like someone to confirm the worst case behavior here.  As I understand it, an unintended combination of `LogicSig` + signing address results in a runtime exception.  If so, it feels _OK_.
* Assuming others see the topic as _large_, I propose:
  * Remove deprecation on `LogicSig` and keep it for v2.0.0. 
  * Extend `LogicSig` docstring to qualify when to use `LogicSig` vs `LogicSigAccount`.  Possibly with a similar note in `LogicSig`.
    * If we take the proposal, I'd appreciate help in writing the commentary.  I still don't think I fully grok the distinctions.

algochoi: I think the suggestion sounds reasonable:
* We want to discourage public usage of `LogicSig`, but the deprecation comment might be a bit misleading (since it is still internally used by other public-facing methods). Adding commentary sounds prudent.
* I'm also okay with keeping everything as-is (state before this PR) as a path of least resistance. The behavior was a bit opaque to me, but it doesn't seem incorrect.

jasonpaulos: That sounds reasonable to me as well

algosdk/transaction.py
```diff
@@ -2507,7 +2507,9 @@ class LogicSig:
     """
     Represents a logic signature
 
-    NOTE: This type is deprecated. Use LogicSigAccount instead.
+    NOTE: LogicSig is still safe for escrow and some other use cases.
+    But it is not fully compatible with delegated logicsig use cases, e.g., rekeying.
+    Under that circumstance, it is recommended to use LogicSigAccount in that case.
```
michaeldiamant, line 2512: @ahangsu Optional - I don't feel strongly, but I think the docstring should read more objectively.   

I realize we're describing functionality where we're a bit unfamiliar with the details.  If you don't feel comfortable with the _suggestion_, then feel welcomed to keep as is.

```suggestion
    NOTE: LogicSig is safe for non-rekeying usage.  If using rekeyed accounts, use LogicSigAccount instead.
```

michaeldiamant (approved):
@ahangsu Thanks for your effort here - I realize the exercise wasn't straightforward.

Issue #409, opened 2022-12-06
CI: Check for sdist metadata warnings


michaeldiamant:
## Problem

During https://github.com/algorand/py-algorand-sdk/pull/397#discussion_r1030005036, we identified setup.py metadata warnings.  These warnings suggest the repo's source distribution fails to meet Python's source package distribution requirements.

Although these warnings have _not_ prevented pypi uploads, it feels like good hygiene to check for + address warnings.  The story requests:
* Add `python setup.py check --strict` to the CI process.
* Address warnings and errors.

Example invocation:
```
~/dev/py-algorand-sdk develop *1 ?4 ──────────────────────────────────────────────────────────────────────────────────────────────────────────  3.8.13 08:52:40 AM
❯ python setup.py check --strict
/Users/michael/.asdf/installs/python/3.8.13/lib/python3.8/site-packages/setuptools/dist.py:463: UserWarning: Normalizing 'v1.20.2' to '1.20.2'
  warnings.warn(tmpl.format(**locals()))
running check
warning: check: missing required meta-data: url

error: Please correct your package.

~/dev/py-algorand-sdk develop *1 ?4 ──────────────────────────────────────────────────────────────────────────────────────────────────────────  3.8.13 08:52:47 AM
❯ echo $?
1
```

Related links:
* https://docs.python.org/3/distutils/examples.html#checking-a-package
* https://docs.python.org/3/distutils/apiref.html#module-distutils.command.check

PR #427, opened 2022-12-23
Infrastructure: Add setup.py check to circle ci

algochoi:
This PR adds a check to see if `setup.py` is well-formed. It also removes the `v` prefix from the version, as it does not seem necessary [(related change)](https://github.com/algorand/py-algorand-sdk/commit/66660112154f85a17a5ccfd9ec36d28a6236dbbb#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R12) and generates a user warning: 
```
dist.py:487: UserWarning: Normalizing 'v1.20.2' to '1.20.2'
  warnings.warn(tmpl.format(**locals()))
```

Closes https://github.com/algorand/py-algorand-sdk/issues/409

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
       - run: pip install -r requirements.txt
       - run: black --check .
       - run: mypy algosdk
-      - run: ! python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet warning
+      - run: { ! python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet warning }
       - run: pytest tests/unit_tests
   integration-test:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,8 @@ jobs:
       - run: black --check .
       - run: mypy algosdk
       - run: |
-          ! python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet warning
+          python setup.py check -s
+          ! python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet --ignore-case "warning"
       - run: pytest tests/unit_tests
   integration-test:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
       - run: pip install -r requirements.txt
       - run: black --check .
       - run: mypy algosdk
-      - run: python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet "warning"; test $? -eq 1
+      - run: python setup.py check -s 2>&1 | tee /dev/tty | {! grep --quiet "warning"}
       - run: pytest tests/unit_tests
   integration-test:
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
       - run: pip install -r requirements.txt
       - run: black --check .
       - run: mypy algosdk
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       - run: pip install -r requirements.txt
       - run: black --check .
       - run: mypy algosdk
+      - run: python setup.py check --strict
```
michaeldiamant, line 29: @algochoi Surprisingly, the check does _not_ catch warnings from normalized versions.  Otherwise, we'd expect the following build to fail:  https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/1174/workflows/147f202e-e47f-4998-b5e8-4437e5d8a0cc/jobs/6005.

In my brief search, I didn't find an explanation for _why_ the check completes _with_ the warning present.

Have I missed an obvious solution here?  If not, are you opposed to extending the check to confirm exit code = 0 _and_ no warnings in output?

algochoi: Yes, your observation is correct. I'm not opposed to adding that check, maybe something like the following to output and check the results of the command:
```
python setup.py check -s 2>&1 | tee /dev/tty | grep --quiet "warning"; test $? -eq 1
```

michaeldiamant: @algochoi As of https://github.com/algorand/py-algorand-sdk/pull/427/commits/2fa369559fb601533c5d9b6e5f5ad57ec6bafddc, I think the check catches the normalizing error, but does _not_ catch errors emitted by `setup.py`.  If you agree with the observation, can you review the following proposal?

Here's a proposal based on local testing:

```
python setup.py check -s 2>&1 | tee /dev/tty | (! grep -qEi 'error|warning')
```

I tested against 2 commits:
* `e90337dcf0dfc9d9b63c5a786c22228058379a38` - `master` as of writing containing _only_ normalization warning (`setup.py check` exit code = 0).
* `a4114577c41ba0abcd5f4fad50419c6abc0ba6e9` - Arbitrary commit containing URL metadata error + normalization warning.

algochoi: The exit code looks like it is returned correctly (both the current check and your proposal), but it's not output to the console. I proposed the following in [6fd99ca](https://github.com/algorand/py-algorand-sdk/pull/427/commits/6fd99caa78dbd19e8cb1a446bf5a500123971aa3):

```
- run: |
          python setup.py check -s
          ! python setup.py check -s 2>&1 | grep --quiet --ignore-case "warning"
```

The first line should fail for any errors anyway and output all errors and warnings. The second line is there to fail if there are any warnings.

.circleci/config.yml
```diff
@@ -26,6 +26,9 @@ jobs:
       - run: pip install -r requirements.txt
       - run: black --check .
       - run: mypy algosdk
+      - run: |
+          python setup.py check -s
+          ! python setup.py check -s 2>&1 | grep --quiet --ignore-case "warning"
```
michaeldiamant, line 31: @algochoi When I locally configured the warning to appear, I see the check passing (return code = 0).

Here's a _suggestion_ that works in my local testing.  I agree with retaining the earlier `setup.py check` invocation present for the user to see output.
 
```suggestion
          python setup.py check -s 2>&1 | (! grep -qEi 'error|warning')
```

algochoi: Thanks for the check, I will commit the suggestion - both variants seem to work on my machine.

michaeldiamant (approved):
@algochoi Thanks for coming up with a solution here so that we avoid regressions!

Issue #374, opened 2022-08-23
Move/rename "tests" module


MaxKh:
## Problem

The py-algorand-sdk's "tests" module may conflict with a local one. I believe that "tests" is a quite popular name for a projects' UTs.

## Solution

Please rename "tests", move into "algosdk" or ship "tests" as "extras" which allows to exclude this module from installation.

winder:
@tzaffi are you familiar with this issue?

tzaffi:
I was involved in reviewing #351 which renamed the `test` directory to `tests`. As a starting point to help investigate a solution, @MaxKh - could you provide a reproducible method to observe the conflict? For example, a github repo with instructions for building but which gets stuck because of this naming issue.

tzaffi:
Also, @jasonpaulos might have some more familiarity with the history of why we chose `test` for the SDK's testing directory starting out.

jasonpaulos:
`test`/`tests` is a pretty common convention for testing directories, and I think it's in our best interests to keep the directory structure like this. If implemented properly, our choice of internal directory names should not cause any problems for consumers of this library.

Perhaps the issue is that our `tests` folder is unintentionally being picked by `setuptools` as a package we wish to export? We do ask `setuptools` to automatically find all packages: https://github.com/algorand/py-algorand-sdk/blob/e39f97ce6c302448814480b16d528cbdcabbbd93/setup.py#L24

If this is the case, there are [alternatives](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html) we could consider, but we first need a reproducible example to be sure.

algochoi:
We no longer ship `tests` as a part of the `py-algorand-sdk` installation, as of #428 . Let us know if this resolves the issue!

PR #428, opened 2022-12-27
Infrastructure: Only package `algosdk` in `setup.py`

### LINKED ISSUE #408: Releases:  Source distribution bundles extraneous artifacts
Like https://github.com/algorand/pyteal/issues/610, py-algorand-sdk bundles extra artifacts into its source distribution (e.g. tests).  The story's request is to limit the source distribution to `algosdk`.

algochoi:
This PR fixes `setup.py` so only `algosdk` and its subdirectories are installed. 

Locally tested:
* Delete `*.egg-info` and `build` directories
* Run `pip install .`
* Check that only `algosdk` is included in the install in `site-packages`. `tests` and `examples` are no longer shipped with the installation, which may address #374 as well.

Closes #408

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,12 @@ setuptools.setup(
         "pycryptodomex>=3.6.0,<4",
         "msgpack>=1.0.0,<2",
     ],
-    packages=setuptools.find_packages(),
+    packages=setuptools.find_packages(
+        include=(
+            "algosdk",
+            "algosdk.*",
+        )
+    ),
     python_requires=">=3.8",
     package_data={"": ["*.pyi", "py.typed"]},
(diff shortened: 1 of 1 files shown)


tzaffi (approved):
I tested py preparing `requirements.txt` files pointing to the following alternatives:
* `py-algorand-sdk@git+https://github.com/algorand/py-algorand-sdk@develop`
* `py-algorand-sdk@git+https://github.com/algorand/py-algorand-sdk@package-algosdk`
I confirmed that `pip install -r` on the second _did not_ include `tests` under `site-packages` but the firs _DID_. BTW, neither includes the `examples` directory.

Issue #431, opened 2023-01-11
AtomicTransactionComposer supplies the incorrect index in method arguments on Create


barnjamin:
### Subject of the issue

If the ATC is used to create an application and a foreign app id is supplied, the method arg will be 0 despite the AVM expecting index 0 of the applications array to be the current app.

I've not tested but this line looks suspect
https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/atomic_transaction_composer.py#L70

barnjamin:
This is fixed in 2.0

PR #432, opened 2023-01-11
Bug-Fix: App id check set

barnjamin:
addresses #431 but I'm not crazy about it

Really the caller should just pass 0 instead of passing None but this is a sort of fail-safe


algosdk/atomic_transaction_composer.py
```diff
@@ -239,6 +239,10 @@ def add_method_call(
             raise error.AtomicTransactionComposerError(
                 "AtomicTransactionComposer cannot exceed MAX_GROUP_SIZE transactions"
             )
+
+        if app_id is None:
```
jasonpaulos, line 243: This argument types for this function don't allow `app_id` to be None. Since that's the case I'm not sure that #431 is an issue.

Do you `app_id` _should_ be allowed to be None?

barnjamin: I don't but I can successfully pass None without it whining anywhere

bbroder-algo:
I think we need to look closer at the signature design of populate_foreign_array

algosdk/atomic_transaction_composer.py
jasonpaulos: Perhaps we should make it throw an error if None is passed?

barnjamin: That could also work, as long as the caller doesn't get a sneaky 0 in app args

jasonpaulos: To confirm, you are seeing the right behavior for foreign app args when 0 is passed in for the app ID during create?

If so, based on that and the type annotations, I'm pretty sure the intended path is for users to explicitly pass 0 in for the app ID and we can consider passing None to be an error.

PR #435, opened 2023-01-15
Enhancement: Type Friendly Exports

tzaffi:
# Introduce PyTeal's fix for VsCode type exports issue

This approach was introduced into PyTeal by @barnjamin. I'm just bringing it into this repo.

### Former VsCode Pylance error when importing `py-algorand-sdk`:

image

## Testing
* There is no specific test in C.I. to prove that it works. But it "works on my machine"
* Also introducing python 3.11 into the C.I. workflow


.circleci/config.yml
```diff
@@ -24,12 +24,9 @@ jobs:
...
       - run: pip install -r requirements.txt
-      - run: black --check .
-      - run: mypy algosdk
-      - run: |
-          python setup.py check -s
-          python setup.py check -s 2>&1 | (! grep -qEi 'error|warning')
-      - run: pytest tests/unit_tests
+      - run: make lint
+      - run: make check-setup
+      - run: make pytest-unit
```
tzaffi, line 29: bringing these commands into the `Makefile`

ahangsu (approved):

Makefile
```diff
@@ -1,6 +1,23 @@
 UNIT_TAGS :=  "$(subst :, or ,$(shell awk '{print $2}' tests/unit.tags | paste -s -d: -))"
 INTEGRATION_TAGS := "$(subst :, or ,$(shell awk '{print $2}' tests/integration.tags | paste -s -d: -))"
 
+generate-init:
+	python -m scripts.generate_init
+
+check-generate-init:
+	python -m scripts.generate_init --check
+
+lint: check-generate-init
```
algochoi, line 10: I think it would be nice if these commands were specified in the README as well!

algosdk/__init__.pyi
```diff
@@ -0,0 +1,68 @@
+## File generated from scripts/generate_init.py.
```
algochoi, line 1: Is there an advantage in checking in the `.pyi` file into the repo?

algochoi (commented):
Explicitly defining `__all__` seems like a good practice. Strangely, I don't notice any PyLance errors on my VSCode.

Makefile
algochoi: In the development section, maybe we can just combine the `black` step and `mypy` step with the equivalent `make` command

algosdk/__init__.pyi
tzaffi: From my own usage experience, to benefit from these, the `pip installing` project only benefits under this condition. But maybe there's a way to add this into the build step and not check in?

tzaffi: Following up from standup, @jasonpaulos made the point that it'll be a little easier in our release process if we _don't_ add the new step of running `generate_init.py` to create an artifact `__init__.pyi` which will only exist in the release. I'm convinced of this point but am open to further discussion.

algochoi: I think it makes sense to include it to simplify the release process - resolving 👍

Makefile
```diff
@@ -1,6 +1,23 @@
...
+	black --check .
+	mypy algosdk
+
+check-setup:
```
algochoi, line 14: nit: Just to be consistent with PyTeal, we could rename this `sdist-check` and include this step under `lint`

https://github.com/algorand/pyteal/blob/master/Makefile#L50

tzaffi:
@algochoi - please have a look at my latest commit 6e70ed02c68e3fcd111cd335a0c11e80052883c2

I believe that most of your points have been addressed.

barnjamin (approved):

algochoi (approved):
LGTM

PR #438, opened 2023-01-21
fix: remove unused positional argument "contract_type" from OverspecifiedRoundError and UnderspecifiedRoundError

ori-shem-tov:
Hi, Ori from the Foundation here.

I'm getting the following error when calling `AlgodClient.block_info(None)`
```
__init__() missing 1 required positional argument: 'contract_type'
```

Looked in the code and realized that both `OverspecifiedRoundError` and `UnderspecifiedRoundError` expect the unused positional argument `contract_type`. This PR fixes that.

### DIFF
--- a/algosdk/error.py
+++ b/algosdk/error.py
@@ -141,7 +141,7 @@ class EmptyAddressError(Exception):
 
 
 class OverspecifiedRoundError(Exception):
-    def __init__(self, contract_type):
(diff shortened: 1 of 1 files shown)


algochoi (approved):
Thank you for the changes - this does look like a fix to me and is flagged by PyLance as well.

PR #442, opened 2023-03-02
Enhancement: Upgrade black, mypy, and add type annotations to algod.py

tzaffi:
`mypy`-ifying `algod.py


algosdk/atomic_transaction_composer.py
```diff
@@ -577,7 +577,8 @@ def execute(
                 tx_info=cast(Any, tx_info),
                 method=self.method_dict[i],
             )
-            method_results.append(abi_result)
+
+        method_results.append(abi_result)
```
tzaffi, line 581:
```suggestion
            method_results.append(abi_result)
```

tests/integration.tags
```diff
@@ -6,6 +6,7 @@
 @auction
 @c2c
 @compile
+@compile.disassemble
```
tzaffi, line 9:
```suggestion
```

examples/log_sig_example.py
```diff
@@ -26,4 +26,6 @@
 
 # send them over network
 # Logicsig will reject the transaction
-acl.send_transaction(lstx)
+
+# TODO: as of v2.0.0, this will fail.
```
tzaffi, line 30: Discovered to be broken according to `mypy`

algosdk/v2client/indexer.py
```diff
@@ -222,7 +222,7 @@ def block_info(
             header_only (bool, optional):
         """
         if block is None and round_num is None:
-            raise error.UnderspecifiedRoundError
+            raise error.UnderspecifiedRoundError()
```
tzaffi, line 225: Makes sense?

algosdk/v2client/algod.py
```diff
@@ -220,24 +262,29 @@ def block_info(
         """
         query = {"format": response_format}
         if block is None and round_num is None:
-            raise error.UnderspecifiedRoundError
+            raise error.UnderspecifiedRoundError()
```
tzaffi, line 265: any objections to this change (and other similar) ?

algosdk/v2client/algod.py
```diff
@@ -266,12 +315,14 @@ def send_transaction(self, txn, **kwargs):
         """
         assert not isinstance(
             txn, transaction.Transaction
-        ), "Attempt to send UNSIGNED transaction {}".format(txn)
+        ), "Attempt to send UNSUPPORTED type of transaction {}".format(txn)
```
tzaffi, line 318: @ahangsu - does this make sense to you?

ahangsu: sure, I think it's a reasonable move.

tzaffi:
> `mypy`-ifying `algod.py

Shall I also re-order the classes in `atomic_transaction_composer.py` ?

@algochoi @barnjamin

algochoi:
> Shall I also re-order the classes in atomic_transaction_composer.py ?

I'm fine either way - certainly won't object to it.

algosdk/v2client/algod.py
```diff
@@ -515,9 +581,10 @@ def _specify_round_string(block, round_num):
         round_num (int): user specified variable
     """
 
-    if block is not None and round_num is not None:
-        raise error.OverspecifiedRoundError
-    elif block is not None:
+    if not (block is None or round_num is None):
```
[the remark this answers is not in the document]
bbroder-algo: equivalent and way harder to reason with

algosdk/v2client/algod.py
```diff
@@ -515,9 +581,10 @@ def _specify_round_string(block, round_num):
...
-    elif block is not None:
-        return str(block)
-    elif round_num is not None:
+    if not (block is None or round_num is None):
+        raise error.OverspecifiedRoundError()
+
+    if round_num is not None:
         return str(round_num)
+
+    return str(block)
```
bbroder-algo, line 590: the prior version returned None if block was None and round_number was None, now you are returning str (None).

algosdk/v2client/algod.py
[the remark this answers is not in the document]
tzaffi: I think that's in the eye of the beholder. But when in doubt, such objection rules the day (don't poke the bear) and so I will revert this change.

algosdk/v2client/algod.py
tzaffi: Here I'm going to push back a bit more.  I argue that the current logic is misleading. It's actually impossible to return `None`. Why?

return `None` meant that:

```python
not (block is not None and round_num is not None)
and not (block is not None)
and not (round_num is not None)
```
That's equivalant to:
```python
not (block is not None and round_num is not None)
and block is None
and round_num is None
```
which is a contradiction.

Therefore, we should not confuse the reader into believing that `None` is possible and that's why I want to make the change

tzaffi:
> > `mypy`-ifying `algod.py
> 
> Shall I also re-order the classes in `atomic_transaction_composer.py` ?
> 
> @algochoi @barnjamin

re-order commit: https://github.com/algorand/py-algorand-sdk/pull/442/commits/22cc731c59decc63d55b64810fa4d01cfa2058e6

barnjamin (approved):

algosdk/v2client/algod.py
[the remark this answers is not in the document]
algochoi: I agree it's less readable, but if `mypy` complains about it, I'd be okay with it if there was a comment explaining the DeMorgan-style logic and optionally making it more explicit by having an if-elif-else structure here.

examples/log_sig_example.py
algochoi: Does mypy complain that the output is not consumed? Sending the logicsig seems integral to this example

algosdk/v2client/algod.py
```diff
@@ -247,13 +294,15 @@ def status_after_block(self, block_num=None, round_num=None, **kwargs):
...
+            raise error.UnderspecifiedRoundError()
         req = "/status/wait-for-block-after/" + _specify_round_string(
             block_num, round_num
         )
         return self.algod_request("GET", req, **kwargs)
 
-    def send_transaction(self, txn, **kwargs):
+    def send_transaction(
+        self, txn: "transaction.Transaction", **kwargs: Any
+    ) -> Any:
```
algochoi, line 305:
```suggestion
    ) -> str:
```

algosdk/atomic_transaction_composer.py
```diff
@@ -14,6 +14,7 @@
 )
 
 from algosdk import abi, error, transaction
+from algosdk.transaction import GenericSignedTransaction
```
algochoi, line 17: nice👍

algosdk/v2client/algod.py
```diff
@@ -331,17 +384,21 @@ def pending_transaction_info(
...
         """Return algod versions."""
         req = "/versions"
         return self.algod_request("GET", req, **kwargs)
 
-    def send_transactions(self, txns, **kwargs):
+    def send_transactions(
+        self,
+        txns: "Iterable[transaction.GenericSignedTransaction]",
+        **kwargs: Any,
+    ) -> Any:
```
algochoi, line 401: maybe return str here too

algochoi (commented):
Looking good, minor questions/comments 🥇

algosdk/v2client/algod.py
bbroder-algo: ah ok.  I figured that might be true but I didn't turn the crank on it.

examples/log_sig_example.py
algochoi: Sure thing, I'm not even sure if we enforce `mypy` to be run on the `examples` yet, so we could uncomment back out and investigate later.

algochoi: It looks like we only `mypy` algosdk: https://github.com/algorand/py-algorand-sdk/blob/develop/Makefile#L14

tzaffi:
I must have been conflating `mypy` with `pylance` in VS Code

algosdk/v2client/algod.py
```diff
@@ -220,24 +262,29 @@ def block_info(
         """
         query = {"format": response_format}
         if block is None and round_num is None:
```
bbroder-algo, line 264: Why can't this check \be in _specify_round_string?

algosdk/v2client/algod.py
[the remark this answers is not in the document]
tzaffi: reverted the DeMorgan here: https://github.com/algorand/py-algorand-sdk/pull/442/commits/e988d3a2d5bd21ae6cd6284380dc3de396034524

algosdk/v2client/algod.py
tzaffi: good catch!

tzaffi: 3e13b44dbe0b576a861363c0386e038ff7b18137

bbroder-algo (approved):

examples/log_sig_example.py
```diff
@@ -26,4 +26,6 @@
...
+# acl.send_transaction(lstx)
```
algochoi, line 31:
```suggestion
acl.send_transaction(lstx)
```

tzaffi: 20327a3579c1bd689d28d7cc6e7791c0364c4197

algochoi (approved):

algosdk/v2client/algod.py
```diff
@@ -93,7 +123,19 @@ def algod_request(
...
-    def account_info(self, address, exclude=None, **kwargs):
+    @classmethod
+    def _assert_json_response(
+        cls, params: Mapping[str, Any], endpoint: str = ""
+    ) -> None:
+        if "response_format" in params:
+            if params["response_format"] != "json":
+                raise error.AlgodRequestError(
+                    f"Only json response is supported{ (' for ' + endpoint) if endpoint else ''}."
+                )
```
tzaffi, line 134:
```suggestion
        if "response_format" in params and params["response_format"] != "json":
            raise error.AlgodRequestError(
                f"Only json response is supported{ (' for ' + endpoint) if endpoint else ''}."
            )
```

algosdk/v2client/algod.py
```diff
@@ -93,7 +123,18 @@ def algod_request(
         else:
             return resp.read()
 
-    def account_info(self, address, exclude=None, **kwargs):
+    @classmethod
+    def _assert_json_response(
+        cls, params: Mapping[str, Any], endpoint: str = ""
+    ) -> None:
+        if "response_format" in params and params["response_format"] != "json":
```
tzaffi, line 130:
```suggestion
        if params.get("response_format", "json") != "json":
```

algosdk/atomic_transaction_composer.py
```diff
@@ -76,11 +56,194 @@ def populate_foreign_array(
...
+        pass
+
+    @abstractmethod
+    def sign_transactions(
+        self, txn_group: List[transaction.Transaction], indexes: List[int]
+    ) -> List[GenericSignedTransaction]:
+        pass
+
+
+class AccountTransactionSigner(TransactionSigner):
```
tzaffi, line 95: moved theseup

algosdk/atomic_transaction_composer.py
```diff
@@ -76,11 +56,234 @@ def populate_foreign_array(
     return offset + len(foreign_array) - 1
 
 
-GenericSignedTransaction = Union[
```
tzaffi, line 79: now this is imported from `transaction.py`

algosdk/atomic_transaction_composer.py
```diff
@@ -174,16 +379,14 @@ def add_method_call(
...
-        ] = None,
+        signer: TransactionSigner,
+        method_args: Optional[List[Union[Any, TransactionWithSigner]]] = None,
         on_complete: transaction.OnComplete = transaction.OnComplete.NoOpOC,
         local_schema: Optional[transaction.StateSchema] = None,
         global_schema: Optional[transaction.StateSchema] = None,
         approval_program: Optional[bytes] = None,
         clear_program: Optional[bytes] = None,
-        extra_pages: Optional[int] = None,
+        extra_pages: int = 0,
```
tzaffi, line 389: downstream, the optionality of the`extra_pages` params is signalted with `0` so I'm unifying the API here.

algosdk/v2client/algod.py
```diff
@@ -1,14 +1,37 @@
...
+    Union,
+    cast,
+)
 import urllib.error
 from urllib import parse
 from urllib.request import Request, urlopen
 
 from algosdk import constants, encoding, error, transaction, util
 
+AlgodResponseType = Union[Dict[str, Any], bytes]
```
tzaffi, line 22: `AlgodResponseType` is a result of the fact that the method `algod_request()` returns a dict when its `response_format == 'json'` (this is the default) and otherwise `bytes`

bbroder-algo (approved):

PR #441, opened 2023-02-20
Documentation: Adding examples to be pulled in to docs

barnjamin:
why do i always have these release tags in my prs 🤔

PR #444, opened 2023-03-05
Draft: Secure key implementation.

BarryJRowe:
There is an unfortunate practice of including mnemonics/private keys in example code for pyalgosdk. This results in developers including such keys in their code when deploying smart contracts, and as a result, every week there are mainnet dev keys being leaked on github by developers who accidentally include such a key somewhere is some code on a public repository.

Within about 5 hours of such a leak, those keys are found by hackers with a github search composed of "(some mnemonic word) algosdk", with a sort based on most recently indexed (after about 5 hours, committed code is indexed and searchable). Besides the lost algo and assets from these leaks, the rekeying that the hackers do means that contracts that are still upgradable (and assets with a manager address) must be reissued because of how easy it is to upgrade contracts/assets. Having a dev key leak on Ethereum is bad, but leaking a dev key on Algorand is far, far worse.

This PR is a potential solution by including a simple way to include encryption of keys on disk with a password. When creating a key, you can paste in your mnemonic and include a password to encrypt the key to disk in a json file. When loading the key, you can type in the password to decrypt the json file and return the private key. This is particularly useful for scripts that need to run periodically since the decryption of the key only needs to happen once when the script is started.

Example usage:

```
#writing a key
enc_key = SecureKey("my_key.json")
>> Input mnemonic
>> Input encryption password
>> Reinput password
>> Input key description 
enc_key.write_key()

#Loading a key from prompt
loaded_pk = SecureKey("my_key.json").load_key()
>> Input password

#Loading a key from password (eg: for scripts that must run on startup)
loaded_pk = SecureKey("my_key.json").load_key(password)

```

This is a draft commit. Unittests will be added once requirements are solidified.

### DIFF
--- a/algosdk/secure_key.py
+++ b/algosdk/secure_key.py
@@ -1,0 +1,120 @@
+from password_strength import PasswordPolicy, PasswordStats
+from algosdk import mnemonic, logic, account
+from Crypto.Cipher import AES
+
+import random
+import json
+import base64
+import getpass
+import hashlib
+
+
+class InvalidPasswordError(Exception):
+    def __init__(self):
+        self.message = "Invalid password!"
+
+
+class SecureKey:
+    def __init__(self, filename):
+        self.filename = filename
+
+    def write_key(self, pk=None, passwword=None, description=None):
+        print("Please copy and paste your private key mnemonic below")
+        while mnem is None:
+            mnem = getpass.getpass("Mnemonic: ")
+            mnem = mnem.replace(",", " ").replace("  ", " ")
+            try:
+                pk = mnemonic.to_private_key(mnem)
+                print("Key address: " + account.address_from_private_key(pk))
+            except:
+                if len(mnem.split(" ") != 25):
+                    print("Mnemonic must be 25 words long. Please retry.")
+                else:
+                    print("Invalid mnemonic. Please retry.")
+                mnem = None
+
+        address = account.address_from_private_key(pk)
+        nonce1 = str(random.random())
+        nonce2 = str(random.random())
+
+        while password is None:
+            password1 = getpass.getpass("Choose a password: ")
+            password2 = getpass.getpass("Confirm password: ")
+            strength = PasswordStats(password1).strength()
+            if password1 != password2:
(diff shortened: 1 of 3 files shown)


bbroder-algo:
Thanks Barry.  We'll take a look.

algochoi:
Hey there!

The SDKs currently serve as a client library for algod/kmd/indexer and aim to be consistent across the four languages that we support. 

From a cursory look, it looks like this PR is a standalone script that might be more suitable as its own repo or somewhere like https://github.com/barnjamin/sdk-extras. 

> There is an unfortunate practice of including mnemonics/private keys in example code for pyalgosdk. 

Agreed, the example is only meant for toy programs on testnets, and otherwise, key management tools like [kmd](https://github.com/algorand/go-algorand/blob/master/daemon/kmd/README.md) should be used.

BarryJRowe:
> Hey there!
> 
> The SDKs currently serve as a client library for algod/kmd/indexer and aim to be consistent across the four languages that we support.
> 
> From a cursory look, it looks like this PR is a standalone script that might be more suitable as its own repo or somewhere like https://github.com/barnjamin/sdk-extras.
> 
> > There is an unfortunate practice of including mnemonics/private keys in example code for pyalgosdk.
> 
> Agreed, the example is only meant for toy programs on private networks/sandboxes, and otherwise, key management tools like [kmd](https://github.com/algorand/go-algorand/blob/master/daemon/kmd/README.md) should be used.

This sounds more like a reason to replicate a class like this to the four sdks and have it used in the example documentation. As I said, a quick search shows that this is happening every week with dev keys, if not more with testnet keys being recycled for use on mainnet. I also know that not everyone uses kmd, especially if they're using an api to connect to the network.

barnjamin:
hey @BarryJRowe this is a good point re examples using bad practice of pasting in mnemonic.

We've been reworking the examples for all the sdks (ex https://github.com/algorand/py-algorand-sdk/pull/441) that removes the practice altogether save for the example showing how to recover a mnemonic. Instead the examples just pull from sandbox KMD outside the snippet and use those accounts throughout

PR #451, opened 2023-03-22
ATC: Refactor Pending Transaction Information in ATC into try block

algochoi:
This PR moves back the pending_transaction_information call in the ATC into a try block. This PR also refactors the `parse_response` helper method so individual ABI result are parsed at a time (instead of the entire group), and can report the `decode_error` at the transaction level. This also reduces the total number of pending tx info calls we potentially make, as we only need to call them for ABI Method calls in the transaction group.


algosdk/atomic_transaction_composer.py
```diff
@@ -785,80 +803,78 @@ def execute(
...
-            except Exception as e:
-                decode_error = e
-
-            method_results.append(
-                ABIResult(
+    def parse_result(
+        self, method: abi.Method, txid: str, txn: Dict[str, Any]
+    ) -> ABIResult:
+        tx_id = txid
+        raw_value = b""
```
bbroder-algo, line 839: `bytes()`?

algosdk/atomic_transaction_composer.py
```diff
@@ -785,80 +803,78 @@ def execute(
...
-        ]
-
-        method_results = self.parse_response(tx_results)
+        method_results: List[ABIResult] = []
+
+        for method_index, method in self.method_dict.items():
+            tx_id = self.tx_ids[method_index]
+            result: ABIResult = ABIResult(
+                tx_id=tx_id,
+                raw_value=b"",
```
bbroder-algo, line 812: `bytes()`?

algosdk/atomic_transaction_composer.py
```diff
@@ -718,7 +718,25 @@ def simulate(
...
-        method_results = self.parse_response(txn_results)
+        method_results: List[ABIResult] = []
+        for method_index, method in self.method_dict.items():
+            tx_id = self.tx_ids[method_index]
+            result: ABIResult = ABIResult(
+                tx_id=tx_id,
+                raw_value=b"",
+                return_value=None,
+                decode_error=None,
+                tx_info={},
```
bbroder-algo, line 729: `tx_info = txn_results[method_index]`?

algochoi: good point! addressed in [5e2d2b7](https://github.com/algorand/py-algorand-sdk/pull/451/commits/5e2d2b70930d8fe2fc623becc941d743fa5fc1d0)

algosdk/atomic_transaction_composer.py
algochoi: they are functionally the same, but I changed all instances of empty byte literals to `bytes()` just in case it helps for readability in [5e2d2b7](https://github.com/algorand/py-algorand-sdk/pull/451/commits/5e2d2b70930d8fe2fc623becc941d743fa5fc1d0)

algosdk/atomic_transaction_composer.py
```diff
@@ -785,80 +803,78 @@ def execute(
...
-    def parse_response(self, txns: List[Dict[str, Any]]) -> List[ABIResult]:
-        method_results = []
-        for i, tx_info in enumerate(txns):
-            tx_id = self.tx_ids[i]
-            raw_value: Optional[bytes] = None
-            return_value = None
-            decode_error = None
-
-            if i not in self.method_dict:
-                continue
```
jasonpaulos, line 811: > This also reduces the total number of pending tx info calls we potentially make, as we only need to call them for ABI Method calls in the transaction group.

Doesn't this branch in the previous code prevent us from getting pending txn info from non-method calls?

jasonpaulos (approved):
Looks good, thanks for making this change.

I have an inconsequential comment about one of your claims for this PR, but that's by no means a blocker.

algosdk/atomic_transaction_composer.py
algochoi: This is for parsing the txn info - in line 789, I think we previously tried getting all tx info results

algochoi: To be a bit clearer, this was from the most recent commit. 

You're right that originally this branch prevented us from getting the pending txn info call in the try block: https://github.com/algorand/py-algorand-sdk/blob/ce14799e3e6a134a14061730a9385cb1ab844b4d/algosdk/atomic_transaction_composer.py#L533

PR #457, opened 2023-03-24
Fix: improve SignedTransaction type signature for dryrun and send_transaction

barnjamin:
Adds type parameter of GenericSignedTransaction to a couple places that had explicitly listed SignedTransaction types

### DIFF
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -307,13 +307,13 @@ class AlgodClient:
         return self.algod_request("GET", req, **kwargs)
 
     def send_transaction(
-        self, txn: "transaction.Transaction", **kwargs: Any
+        self, txn: "transaction.GenericSignedTransaction", **kwargs: Any
     ) -> str:
         """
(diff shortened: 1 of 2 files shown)


algochoi (commented):
thanks for catching this - there is also a `GenericSignedTransaction` type defined at the end of this file that defines the union of signed transactions

algochoi (approved):

algosdk/v2client/algod.py
```diff
@@ -307,13 +307,13 @@ def status_after_block(
         return self.algod_request("GET", req, **kwargs)
 
     def send_transaction(
-        self, txn: "transaction.Transaction", **kwargs: Any
+        self, txn: "transaction.GenericSignedTransaction", **kwargs: Any
```
algochoi, line 310: nit: do we still need the quotes?

barnjamin: It failed CI the first time with a message about partially initialized module transaction, didnt dig deep but quoting fixed it

Issue #458, opened 2023-03-25
algosdk.wallet.Wallet.sign_multisig_transaction() fails for rekeyed multisig accounts


oysterpack:
### algosdk.wallet.Wallet.sign_multisig_transaction() fails for rekeyed multisig accounts

When a multisig account is rekeyed to another multisig account, the KMD wallet fails to sign the `MultisigTransaction` with the following error message: "given multisig preimage hashes to neither Sender nor AuthAddr"

### Your environment
```
> algokit doctor
timestamp: 2023-03-25T20:14:50+00:00
AlgoKit: 0.5.0
AlgoKit Python: 3.11.2 (main, Feb 25 2023, 06:33:57) [GCC 11.3.0] (location: /home/alfiozappala/.local/pipx/venvs/algokit)
OS: Linux-6.2.0-76060200-generic-x86_64-with-glibc2.35
docker: 20.10.21
docker compose: 2.16.0
git: 2.34.1
python: 3.11.2 (location: /home/alfiozappala/.cache/pypoetry/virtualenvs/oysterpack-ZmNOoksg-py3.11/bin/python)
python3: 3.11.2 (location: /home/alfiozappala/.cache/pypoetry/virtualenvs/oysterpack-ZmNOoksg-py3.11/bin/python3)
pipx: 1.1.0
poetry: 1.4.0
node: 18.14.0
npm: 9.3.1
```
py-algorand-sdk   2.1.2

### Steps to reproduce

Below is a unit test that replicates the issue against a local running sandbox environment spun up via AlgoKit.
Signing the MultisigTransaction via KMD fails. The work around is to export the private keys and sign the MultisigTransaction directly using the private keys.

```python
import unittest
from typing import Final, cast, Any

from algosdk import kmd
from algosdk.account import generate_account
from algosdk.error import AlgodHTTPError, KMDHTTPError
from algosdk.transaction import (
    Multisig,
    PaymentTxn,
    wait_for_confirmation,
    SuggestedParams,
    MultisigTransaction,
)
from algosdk.v2client.algod import AlgodClient
from algosdk.wallet import Wallet
from beaker import sandbox
from beaker.consts import algo
from beaker.sandbox import SandboxAccount
from beaker.sandbox.kmd import get_sandbox_default_wallet


class AccountDoesNotExist(Exception):
    pass


Address = str
MicroAlgos = int


def get_auth_address(address: Address, algod_client: AlgodClient) -> Address:
    """
[...]

PR #460, opened 2023-04-03
adding authaddr for multisig sign when the msig has been rekeyed

barnjamin:
Addresses https://github.com/algorand/py-algorand-sdk/issues/458 

Makes the test included in that issue pass but some should be included here


bbroder-algo:
I like this but there's no tests

bbroder-algo (approved):
lgtm

PR #463, opened 2023-04-06
BugFix: ATC error message improvement

### LINKED ISSUE #462: Difficult to debug error message for ATC
### Subject of the issue

While looking into https://github.com/algorand-devrel/beaker/issues/231

I found that the error message here:
https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/atomic_transaction_composer.py#L498-L507

Was insufficient to easily figure out the issue was _really_ that the transaction type didn't match.

### Solution

Also return the transaction type expected/got

barnjamin:
closes #462


tests/steps/steps.py
```diff
@@ -32,7 +32,7 @@ def parse_string(text):
 algod_port = 60000
 kmd_port = 60001
 
-DEV_ACCOUNT_INITIAL_MICROALGOS: int = 10_000_000
+DEV_ACCOUNT_INITIAL_MICROALGOS: int = 100_000_000
```
jasonpaulos, line 35: Why is this here too? Seems completely unrelated

jasonpaulos (commented):
Looks good but why change `steps.py`?

tests/steps/steps.py
barnjamin: Tests have been v flakey with issues related to overspend errors. This change was an experiment to see if it fixed the overspend errors.

The tests passed on the first attempt once I included it, but I cant be sure this is _the_ fix

algochoi: I outlined the rationale for this one liner here as well: https://github.com/algorand/py-algorand-sdk/pull/459

It resolves current flaky tests because rekeying will change how transient accounts are funded in a non-deterministic way

algochoi: We already do this in the JS SDK but swept it under the rug in another PR: https://github.com/algorand/js-algorand-sdk/blob/develop/tests/cucumber/steps/steps.js#L122

algochoi (approved):
This looks like a good UX improvement, and I also think the one liner change in `steps.py` is nice to immediately fix any flaky tests

jasonpaulos (approved):

tests/steps/steps.py
bbroder-algo: The rekey changes the ordering of the kmd wallets in py-algorand-sdk (I believe)
i think the change should be to provide non-deterministic access to the main network accounts instead

PR #467, opened 2023-04-18
Update Multisig .sign() Method

mccabe-david:
This allows for easier signing when some of the subsig accounts have been rekeyed

### DIFF
--- a/algosdk/transaction.py
+++ b/algosdk/transaction.py
@@ -2224,12 +2224,13 @@ class MultisigTransaction:
         else:
             self.auth_addr = None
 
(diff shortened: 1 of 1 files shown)


jasonpaulos:
@mccabe-david that's right. The important thing to understand is that the individual parts of multisigs are not accounts, they are public keys. The private key corresponding to that public key must always be used to sign that multisig, even if the individual account matching that public key has been rekeyed.

Though if you wanted to, you could "change" a multisig account's keys by rekeying it to a different multisig. That way you can change individual keys if you wish (though you'll need a threshold of signatures to approve).

PR #468, opened 2023-04-20
algod: Add endpoints for devmode timestamps, sync, and ready

algochoi:
This PR implements the devmode timestamp offset, sync, and ready endpoints in the Python SDK.

Tested locally and added some cucumber tests.

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="path-unit-tests"
+SDK_TESTING_BRANCH="master"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="timestamp-tests"
+SDK_TESTING_BRANCH="path-unit-tests"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="timestamp-tests"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -646,6 +646,21 @@ class AlgodClient:
         )
         return self.simulate_transactions(request, **kwargs)
 
+    def get_ledger_state_delta(
+        self, round: int, **kwargs: Any
+    ) -> AlgodResponseType:
+        """
+        Get the state delta from the ledger.
+
+        Args:
+            round (int): State delta round
+
+        Returns:
+            Dict[str, Any]: Response from algod
+        """
+        req = f"/deltas/{round}"
+        return self.algod_request("GET", req, **kwargs)
+
     def get_sync_round(self, **kwargs: Any) -> AlgodResponseType:
         """
         Get the minimum sync round for the ledger.
--- a/algosdk/v2client/algod.py
+++ b/algosdk/v2client/algod.py
@@ -113,6 +113,8 @@ class AlgodClient:
             try:
                 return json.load(resp)
             except Exception as e:
+                if resp.status == 200:
+                    return {}
                 raise error.AlgodResponseError(
                     "Failed to parse JSON response from algod"
                 ) from e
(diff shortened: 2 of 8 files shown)


algosdk/v2client/algod.py
```diff
@@ -113,6 +113,8 @@ def algod_request(
             try:
                 return json.load(resp)
             except Exception as e:
+                if resp.status == 200:
```
algochoi, line 116: We have 5 endpoints that return no response upon success.

[204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) might be better, but would entail a breaking change

tzaffi: Is this solvable but providing `response_format` = some new field and a new `elif` code-path?

.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="timestamp-tests"
```
algochoi, line 3: TODO: revert back to master

algosdk/v2client/algod.py
```diff
@@ -641,6 +646,78 @@ def simulate_raw_transactions(
...
+        Set the minimum sync round for the ledger.
+
+        Args:
+            round (int): Sync round
+
+        Returns:
+            Dict[str, Any]: Response from algod
+        """
+        req = f"/ledger/sync/{round}"
+        return self.algod_request("GET", req, **kwargs)
```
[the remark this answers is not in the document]
algochoi: thank you 😄

algosdk/v2client/algod.py
```diff
@@ -641,6 +646,78 @@ def simulate_raw_transactions(
...
+
+    def unset_sync_round(self, **kwargs: Any) -> AlgodResponseType:
+        """
+        Unset the minimum sync round for the ledger.
+
+        Returns:
+            Dict[str, Any]: Response from algod
+        """
+        req = "/ledger/sync"
+        return self.algod_request("DELETE", req, **kwargs)
```
algochoi, line 680: DELETE does not work on Python right now - might be because `http.server` does not have a do_delete() method: https://docs.python.org/3/library/http.server.html#http.server.SimpleHTTPRequestHandler.do_GET

algochoi: resolved - add do_delete method in the mock server

algosdk/v2client/algod.py
```diff
@@ -113,6 +113,11 @@ def algod_request(
             try:
                 return json.load(resp)
             except Exception as e:
+                # Some algod responses currently return a 200 OK
+                # but have an empty response.
+                # Do not return an error, and just return an empty response.
+                if resp.status == 200:
```
tzaffi, line 119: nit. If at all possible, prefer to explicitly anticipate the "happy path" without catching an exception. I.e., even though its a bit more convoluted to explicitly check for an empty response along with a `200` then to try/catch as here, it'll be a bit more robust if there's a bug whereby algod returns a 200 errantly and has an unparseable response that is alerting of an error. Alternately, if `json.load()` gives a special error when an empty body is provided, you can selectively catch that _specific_ `Exception`-type instead of the generic type.

tzaffi: Another approach would be to have an explict "white list" of acceptable endpoints that we know shouldn't return a body... But that seems like a maintenance hassle.

tzaffi: A little less annoying would be to provide a new parameter such as `empty_response` defaulting to `False`.

algochoi: > Alternately, if json.load() gives a special error when an empty body is provided, you can selectively catch that specific Exception-type instead of the generic type.

Unfortunately, I don't think there is a separate exception for empty bodies vs bad bodies (parsing error).

> A little less annoying would be to provide a new parameter such as empty_response defaulting to False.

I like this approach at the moment - Ideally the server should return a different response (`204`) if there is an empty body, but that seems like a bigger hill to climb.

algochoi: I jumped the gun here, it looks like there are two ways an empty response is returned from algod:

1. We return a `NoContent` response in the v2 endpoints: https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/v2/handlers.go#L1114
2. We encode a `nil` into the response in the common endpoints: https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/common/handlers.go#L92

For 1, the server returns a length 0 response which causes an exception when parsing in Python, but for 2, the response is of length 5 for some reason and it parses to a `None` object in Python (without raising an exception). To handle case 1, it seems sufficient to just check the response length and ensure that it is 0.

tests/integration.tags
```diff
@@ -14,4 +14,3 @@
 @rekey_v1
 @send
 @send.keyregtxn
-@simulate
```
algochoi, line 17: @ahangsu I decided to take this tag out until the tests are fixed

ahangsu: sure, will fix soon!

tzaffi (approved):
LGTM

Eric-Warehime (approved):

PR #469, opened 2023-04-24
Simulation: Lift log limits option in SimulateRequest

ahangsu:


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="simulate-request-log-option"
```
ahangsu, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

tests/steps/other_v2_steps.py
```diff
@@ -1441,9 +1442,12 @@ def simulate_transaction(context):
 @then("the simulation should succeed without any failure message")
 def simulate_transaction_succeed(context):
     if hasattr(context, "simulate_response"):
-        assert context.simulate_response["would-succeed"] is True
+        assert len(context.simulate_response["failure-message"]) == 0
```
jasonpaulos, line 1445: Is this right? I would think you'd need to index into `txn-groups` to get at this

tests/steps/other_v2_steps.py
```diff
@@ -1469,11 +1473,42 @@ def simulate_atc_failure(context, group, path, message):
...
+def attach_sim_request_to_txn_group_simulation(context):
+    context.simulate_response = (
+        context.atomic_transaction_composer.simulate_with_request(
+            context.app_acl, context.simulate_request
+        )
+    )
+
+
+@then('the simulation with "{power}" should not have failure message.')
+def power_pack_simulation_should_pass(context, power: str):
```
jasonpaulos, line 1500: This step is checking two different things (power pack in response and no failure message), perhaps break this up into 2 different ones? I'd rather have a new step for each power pack than trying to fit all the future ones into here.

tests/steps/other_v2_steps.py
```diff
@@ -1469,11 +1473,42 @@ def simulate_atc_failure(context, group, path, message):
...
 
 
+@when("I make a new simulate request.")
+def make_simulate_request(context):
+    context.simulate_request = SimulateRequest(txn_groups=[])
+
+
+@then("I lift log limits on that simulate request.")
+def lift_log_limits_in_request(context):
+    context.simulate_request.lift_log_limits = True
```
jasonpaulos, line 1487: If not too much trouble I'd probably suggest changing to `allow_more_logging` here too...

algosdk/atomic_transaction_composer.py
```diff
@@ -754,19 +784,56 @@ def simulate(
...
             failed_at=txn_group.get("failed-at"),
             simulate_response=simulation_result,
             tx_ids=self.tx_ids,
             results=sim_results,
+            eval_overrides=SimulateEvalOverrides.from_simulation_result(
+                simulation_result
+            ),
+        )
+
+    def simulate_with_request(
```
jasonpaulos, line 802: Can we combine this with the existing `simulate` method? I don't see the point in having both

The added logic in this function to accommodate signed vs unsigned txns is interesting, but I'm not convinced it's necessary. #420 added `EmptySigner`, which can be used in the ATC if the caller really doesn't want to simulate with signatures

tests/steps/other_v2_steps.py
ahangsu: no, this is def not right, I am mucking around with changed response, will update later.

tests/steps/other_v2_steps.py
ahangsu: right, that should align with naming

algosdk/atomic_transaction_composer.py
```diff
@@ -766,7 +800,46 @@ def simulate(
...
+        self,
+        client: algod.AlgodClient,
+        request: models.SimulateRequest = models.SimulateRequest(
```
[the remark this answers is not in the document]
ahangsu: yea, I will take that off, feels of when I start writing test step

algosdk/atomic_transaction_composer.py
ahangsu: ok! that sounds great to me, for I was not aware of `EmptySigner` existence, but that def helps me to transform from `txn` to `stxn` easily.

algosdk/atomic_transaction_composer.py
```diff
@@ -255,15 +256,41 @@ def __init__(
...
         self.tx_info = tx_info
         self.method = method
-        self.missing_signature = missing_signature
+
+
+class SimulateEvalOverrides:
+    def __init__(
+        self,
+        *,
+        max_log_calls: Optional[int] = None,
```
jasonpaulos, line 272: Should we address `allow-empty-signatures` in another PR?

algosdk/atomic_transaction_composer.py
```diff
@@ -694,6 +725,7 @@ def simulate(
 
         Args:
             client (AlgodClient): Algod V2 client
+            request (models.SimulateReuqest): SimulateRequest with options in simulation.
```
jasonpaulos, line 728: Can you say that the txn groups from the request will be overrwritten by the composer's group?

Also, spelling error ("SimulateReuqest" - > "SimulateRequest")

tests/steps/other_v2_steps.py
```diff
@@ -1446,9 +1447,12 @@ def simulate_transaction(context):
 @then("the simulation should succeed without any failure message")
 def simulate_transaction_succeed(context):
     if hasattr(context, "simulate_response"):
-        assert context.simulate_response["would-succeed"] is True
+        assert "failure-message" not in context.simulate_response["txn-groups"]
```
jasonpaulos, line 1450: Is this correct? I think something like this might be better (untested):

```python
assert all("failure-message" not in group for group in context.simulate_response["txn-groups"])
```

tests/steps/other_v2_steps.py
```diff
@@ -1462,47 +1466,50 @@ def simulate_atc(context):
...
-        resp = context.atomic_transaction_composer_return.simulate_response
+@then("I allow more logs on that simulate request.")
+def allow_more_logs_in_request(context):
+    context.simulate_request.allow_more_logs = True
 
-    group_idx: int = int(group)
-    tx_idxs: list[int] = [int(pe) for pe in path.split(",")]
 
-    assert resp["would-succeed"] is False
+@then("I attach the simulate request to simulate the transaction group.")
```
jasonpaulos, line 1491: nit: maybe "use" instead of "attach" here? Saying this because at first glance, attaching something does not necessarily imply executing the simulation

algosdk/atomic_transaction_composer.py
ahangsu: good callout, I will put the field here soon

tests/steps/other_v2_steps.py
ahangsu: see [renaming a step](https://github.com/algorand/py-algorand-sdk/pull/469/commits/eb40f7226619c571800dacd91009f186cb9c8b4a)

tests/steps/other_v2_steps.py
ahangsu: see [pr response](https://github.com/algorand/py-algorand-sdk/pull/469/commits/939e1216a567006d6667ecd622965788719a41cf)

algosdk/atomic_transaction_composer.py
ahangsu: see [pr response](https://github.com/algorand/py-algorand-sdk/pull/469/commits/939e1216a567006d6667ecd622965788719a41cf)

algosdk/atomic_transaction_composer.py
ahangsu: see [pr response](https://github.com/algorand/py-algorand-sdk/pull/469/commits/939e1216a567006d6667ecd622965788719a41cf)

algosdk/atomic_transaction_composer.py
```diff
@@ -271,9 +271,11 @@ def __init__(
         *,
         max_log_calls: Optional[int] = None,
         max_log_size: Optional[int] = None,
+        allow_empty_signatures: Optional[int] = None,
```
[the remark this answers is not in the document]
ahangsu: oops, fat finger.

algosdk/atomic_transaction_composer.py
```diff
@@ -271,7 +271,7 @@ def __init__(
         *,
         max_log_calls: Optional[int] = None,
         max_log_size: Optional[int] = None,
-        allow_empty_signatures: Optional[int] = None,
+        allow_empty_signatures: Optional[bool] = None,
```
bbroder-algo, line 274: nice

jasonpaulos (approved):
Once you revert `.test-env`, looks good to me!

Issue #475, opened 2023-05-08
Indexer: Small breaking changes required


tzaffi:
## Problem

The indexer client accepts redundant and error-causing parameters in a few of it's methods. In particular:
* in several places the synonomous `block` and `round_num` are accepted. [EG here](https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/v2client/indexer.py#L189_L193)
* in several places we alert users with `DO NOT USE` a particular param. [EG here](https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/v2client/indexer.py#L836)

## Solution

Remove the offending parameters. As this is a breaking change, we may want to include it into a **V3** upgrade. 
We can start by deprecating usage ASAP.

## Related Issues

#471 

## Urgency

Medium - this is mostly a nuisance, but it does cause confusion

Issue #471, opened 2023-04-29
round_num and block no longer allowed when querying balances


BunsanMuchi:
### Your environment

`py-algorand-sdk    2.0.0`

### Steps to reproduce

1. Call the method `asset_balances(asset_id, round_num=number)` or `asset_balances(asset_id, block=number)`

### Expected behaviour

A balance return at the given block/round number according to the docs: https://py-algorand-sdk.readthedocs.io/en/latest/algosdk/v2client/indexer.html

### Actual behaviour

It seems like the method got changed, and doesn't allow for this parameter anymore, I believe it's a case of outdated docs or an undocumented change, since I get the following TypeErrors:

`TypeError: IndexerClient.indexer_request() got an unexpected keyword argument 'round_num'`
`TypeError: IndexerClient.indexer_request() got an unexpected keyword argument 'block'`

tzaffi:
Thanks for identifying this issue. It appears that the docs getting generated are inaccurate and do not conform with the actual v2 API which [does not accept block/round number](https://developer.algorand.org/docs/rest-apis/indexer/#get-v2assetsasset-idbalances).

This discrepancy ought to be fixed as soon as possible, even though it probably is a longstanding issue.

BunsanMuchi:
Yep, I've noticed it happens with a lot of places on both the SDK and the docs of the SDK. Plus the few methods that still support it end up with either rewind issues or just printing the current round, leading to confusing results. Thanks!

tzaffi:
If you know off hand other instances, we welcome your feedback.

tzaffi:
The fix didn't make it into today's release, but will go in the next one.

PR #476, opened 2023-05-08
bugfix: incorrect indexer docs

tzaffi:
## Fix Docs Generated for Indexer

This is a non-functional change

## Related issues:
* #471
* #475

### DIFF
--- a/algosdk/v2client/indexer.py
+++ b/algosdk/v2client/indexer.py
@@ -108,7 +108,7 @@ class IndexerClient:
         **kwargs
     ):
         """
-        Return accounts that hold the asset; microalgos are the default
+        Return accounts that match the search; microalgos are the default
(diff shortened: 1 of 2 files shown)


algosdk/v2client/indexer.py
```diff
@@ -108,7 +108,7 @@ def accounts(
         **kwargs
     ):
         """
-        Return accounts that match the search; microalgos are the default
+        Return accounts that hold the asset; microalgos are the default
```
tzaffi, line 111:
```suggestion
        Return accounts that match the search; microalgos are the default
```

algochoi (approved):
Looks like a good doc fix

bbroder-algo (approved):
lgtm tyvm

Issue #419, opened 2022-12-12
API docs are stale


michaeldiamant:
## Problem

While considering https://github.com/algorand/js-algorand-sdk/issues/709, @algochoi pointed out https://py-algorand-sdk.readthedocs.io/en/latest/ does _not_ contain recently introduced classes (e.g. https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/box_reference.py).

The story requests:
* Update docs to be current.
* Triage to identify if there's easier ways to keep docs in-sync.
  * Current process relies on manual creation of _rst_ files.
  * If improvements are identified, port to pyteal or open an _issue_.

algochoi:
Could investigate something like: https://github.com/copier-org/copier

tzaffi:
> The story requests:
> 
> * Update docs to be current.

#481 addresses the first concern

> * Triage to identify if there's easier ways to keep docs in-sync.
>   
>   * Current process relies on manual creation of _rst_ files.

## We can automate as follows:

### Create a module RST template:

```add_page.py
module_name = "constants" 

template = """{name}
{equals}

.. automodule:: algosdk.{name}
   :members:
   :undoc-members:
   :show-inheritance:
"""

# Format the template
content = template.format(name=module_name, equals='='*len(module_name))

# Write to the file
with open(f"{module_name}.rst", 'w') as f:
    f.write(content)
```

tzaffi:
>   * Current process relies on manual creation of _rst_ files.
>   * If improvements are identified, port to pyteal or open an _issue_.

We can also write a `docs_test.py` which simply looks for `algosdk.*` modules and asserts that there aren't any unaccounted for in `docs/algosdk/index.rst` or `docs/algosdk/*.rst`.

PR #481, opened 2023-05-15
Enhancement: Adding `box_reference.py` to Read The Docs

tzaffi:
## Box Reference was Missing from our Docs

This addresses a small portion of #419 .

### Testing
Verified locally that I was able to get the page:

image

### DIFF
--- a/docs/algosdk/box_reference.rst
+++ b/docs/algosdk/box_reference.rst
@@ -1,0 +1,7 @@
+box_reference
+=============
+
+.. automodule:: algosdk.box_reference
+   :members:
+   :undoc-members:
+   :show-inheritance:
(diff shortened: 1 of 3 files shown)


algochoi (approved):
Thanks!

It also looks like the `conf.py` file has an outdated copyright and would be nice to update: https://github.com/algorand/py-algorand-sdk/pull/481/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R22

tzaffi:
> Thanks!
> 
> It also looks like the `conf.py` file has an outdated copyright and would be nice to update: https://github.com/algorand/py-algorand-sdk/pull/481/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R22

See latest commit:
image

PR #497, opened 2023-06-01
CICD: Release PR Creation Workflow and Slack Messaging

algobarb:
# Summary
- Adds a Release PR Creation github workflow
- Add Slack Messaging after a release branch completes CircleCI tests
- Add a script/bump_version.py to handle bumping version in setup.py 
- Support Hot-Fixes

## Testing
- The Description is populated with the changelog as expected in this PR: https://github.com/algorand/py-algorand-sdk/pull/495
- Slack message can be found in our internal #devops-experimental-notifications slack channel.


scripts/bump_version.py
```diff
@@ -0,0 +1,28 @@
+# This script bumps up the version in setup.py for new releases.
+# Usage: python bump_version.py {new_version} (--setup_py_path <path-to setup.py>)
+
+import argparse
+import re
+
+parser = argparse.ArgumentParser(
```
tzaffi, line 7: To conform a bit more with the conventions of this repo (and python best practices) I recommend that all the code from line 7 be wrapped inside of a function:
```python
def bump_version():
    parser = argparse.ArgumentParser(
        description="updates the version for a release",
        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
    )
    ... etc. ....
```
and then at the bottom call this function with:

```python
if __name__ == "__main__":
    bump_version()
```

See for example `generate_init.py` in the same directory.

tzaffi (commented):
Thanks for this wonderful addition! It looks great, but I had one suggestion regarding the python script.

scripts/bump_version.py
algobarb: Thanks for the tips! I should follow python best practices.

algobarb: Please take a look at the updated script. I'm open to suggestions to make it better.

algobarb: oh wait hold on. I should run lint

tzaffi (approved):
LGTM

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,127 @@
...
+          echo "release-branch=release/$RELEASE_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Get Latest Release
+        id: get-release
+        run: |
+          releases=$(curl --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+                      --header "Accept: application/vnd.github.v3+json" \
+                      "https://api.github.com/repos/${{ github.repository }}/releases")
+          tag=$(echo "$releases" | jq -r '.[0].tag_name')
+          echo "latest-tag=$tag" >> $GITHUB_OUTPUT
```
onetechnical, line 50: Will the above only get published releases, not pre-releases? Or maybe we want that for comparison?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,127 @@
...
+      - name: Commit Changes
+        uses: EndBug/add-and-commit@v9.1.3
+        with:
+          message: 'update CHANGELOG.md and setup.py version'
+
+      - name: Create Pull Request
+        env:
+          CHANGELOG_CONTENT: ${{ steps.build-changelog.outputs.changelog }}
+          PREVIOUS_VERSION: ${{ steps.get-release.outputs.latest-tag }}
+          GH_TOKEN: ${{ github.token }}
```
onetechnical, line 123: Is this scoped to only work for release branches?

onetechnical (approved):
I had some questions and one nit but nothing blocking.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,127 @@
...
+                "pr_template": "- #{{TITLE}} by #{{AUTHOR}} in ##{{NUMBER}}"
+            }
+
+      - name: Update Changelog
+        env:
+          CHANGELOG_CONTENT: ${{ steps.build-changelog.outputs.changelog }}
+          PREVIOUS_VERSION: ${{ steps.get-release.outputs.latest-tag }}
+        run: |
+          git branch --show-current
+          echo "$(tail -n +2 CHANGELOG.md)" > CHANGELOG.md
```
excalq, line 107: Verifying this should be `>` rather than `>>`?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,127 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      source_ref:
+        description: 'Source branch, commit, or tag to create the release branch from'
+        default: 'develop'
+        type: string
+      release_version:
+        description: 'In the vX.X.X format and will be used for tag and branch name'
+        default: 'vx.x.x'
```
excalq, line 12: non-blocking: Is there any means of validating the presence of the `v` prefix? I'm anticipating `python3 scripts/bump_version.py ${RELEASE_VERSION:1}` doing something bad if it gets forgotten.

excalq (approved):
Only the non-blocking question re version v-prefix, but great work here.

.github/workflows/create-release-pr.yml
algobarb: Well, the source_ref to create the release_branch can come from anywhere including a hot_fix branch. 

The parameter: ```source_ref:
        description: 'Source branch, commit, or tag to create the release branch from'```

.github/workflows/create-release-pr.yml
algobarb: Updated with the releases/latest api and tested it here: https://github.com/algorand/py-algorand-sdk/commit/f064fa2d9f47580f0f1b531924fcb1b8b878e9a1 (will delete this branch later)

.github/workflows/create-release-pr.yml
algobarb: Yes, I set this input as a variable in line 17: 
```
env:
  RELEASE_VERSION: ${{ inputs.release_version }}
```
and it gets checked when it's first used in the second step:
```

      - name: Create Release Branch
        id: create-release-branch
        run: |
          if [[ $RELEASE_VERSION =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
If it doesn't match the `^v` then it will fail the build after checking out the code.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Build Changelog
+        uses: mikepenz/release-changelog-builder-action@v3.7.2
+        id: build-changelog
+        env:
+          PREVIOUS_VERSION: ${{ steps.get-release.outputs.latest-tag }}
+        with:
+          fromTag: ${{ env.PREVIOUS_VERSION }}
+          toTag: ${{ steps.create-release-branch.outputs.release-branch }}
+          ignorePreReleases: true
```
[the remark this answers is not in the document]
algobarb: I fixed the other part where we got PREVIOUS_VERSION... so I think it's ok. But I guess it doesn't handle the case of when we have two pre-releases in a row...

.github/workflows/create-release-pr.yml
excalq: Awesome, sorry I missed that!

algobarb:
Closing to test adding pre-release changes. Will re-open PR afterwards.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,189 @@
...
+      release_version:
+        description: 'The release_version used for the release branch name, e.g. release/vx.x.x'
+        default: 'vx.x.x'
+        required: true
+        type: string
+      pre_release_version:
+        description: "Pre-Release version, e.g. 'b1', will be added behind the release_version in setup.py."
+        required: false
+        type: string
+      release_branch_exists:
```
excalq, line 19: If the step on line 66 created the release branch (if it doesn't exist) _prior_ to the checkout steps, would this still be necessary as an input?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,189 @@
...
+        with:
+          ref: ${{ env.SOURCE_REF }}
+
+      - name: Checkout Release Branch
+        if: ${{ inputs.release_branch_exists }}
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.RELEASE_BRANCH }}
+
+      - name: Create Release Branch
```
excalq, line 66:
```suggestion
      - name: Create Release Branch if it does not exist
```

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,189 @@
...
+        if: ${{ ! inputs.release_branch_exists }}
```
excalq, line 67:
```suggestion
```

algobarb:
Thanks @excalq for the pointers!

I added the recommended changes as well as a few more changes in the rest of the yml file to remove the input parameters: **release_branch_exists** and **source_ref**:
- I added `fetch-option: 0` to the checkout step so it will fetch all of the branches rather than just one commit, so we can check if the release branch exists on remote. 
- I also added a condition that if the release branch exists, it will error if we aren't running the workflow from the release branch. It could indicate that someone may not have realized the release branch was already created. 
- I removed source_ref parameter and instead will use the "workflow_ref", i.e. the ref used to run the workflow.

.github/workflows/create-release-pr.yml
algobarb: Actually I removed the source_ref but I added a check to the "Create Release Branch if it does not exist" step which should ensure that we are on working on the release branch.

PR #507, opened 2023-08-03
bug-fix: include currency-greater-than param for 0 value

### LINKED ISSUE #296: Faulty check in `indexer_client.asset_balances` method makes it impossible to get balances grater than 0.
### Subject of the issue

In the V2 Indexer, when trying to retrieve balances, you cannot specify `min_balance=0` because of the following check, as 0 is falsely: 
```python
        if min_balance:
            query["currency-greater-than"] = min_balance
```

### Your environment

I'm using `py-algorand-sdk==1.8.0` but this issue is present in the `develop` branch of the repository as of 02/23/2022

### Steps to reproduce

1. Set up a v2 indexer client with any node.
2. retreive balances using client.asset_balances(asset_id, min_balance=0)
3. the reponse will be equivalent to client.asset_balances(asset_id)

### Expected behavior

The function should return accounts with balances greater than 0 (as instructions state)

### Actual behavior

The function returns all accounts including those with a balance of 0.

### Solution

Instead of
```python
if min_balance:
    query["currency-greater-than"] = min_balance
```
The code should be:
```py
if min_balance is not None:
    query["currency-greater-than"] = min_balance
```

### Addendum:

This faulty check is actually present in many other places around the sdk including the same function.
I will make a PR if life allows it.
Take care

shiqizng:
This PR fixes #296

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="shiqi/pathupdates"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/algosdk/v2client/indexer.py
+++ b/algosdk/v2client/indexer.py
@@ -152,9 +152,9 @@ class IndexerClient:
             query["limit"] = limit
         if next_page:
             query["next"] = next_page
-        if min_balance is not None and min_balance >= 0:
+        if min_balance is not None:
             query["currency-greater-than"] = min_balance
-        if max_balance is not None and max_balance > 0:
+        if max_balance:
             query["currency-less-than"] = max_balance
         _specify_round(query, block, round_num)
         if auth_addr:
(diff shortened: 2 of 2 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="shiqi/pathupdates"
```
shiqizng, line 3: TODO: revert this to master before merge.

algosdk/v2client/indexer.py
```diff
@@ -152,9 +152,9 @@ def accounts(
             query["limit"] = limit
         if next_page:
             query["next"] = next_page
-        if min_balance:
+        if min_balance is not None and min_balance >= 0:
```
algochoi, line 155: The second part of the conditional (>=0) seems to change the existing behavior of the request. Before, a negative param would return an error from the server, but now the client just ignores the negative. 

I don't feel too strongly about it, but I think returning an error (either client or server side) seems to be a bit more desirable.

algochoi (commented):
Looks good, one comment about changing existing behavior to the request

winder (approved):

shiqizng:
license/cal check hanging. opening a new PR.

PR #521, opened 2023-12-04
DX: Keyreg bytes

jannotti:
Allow the various key-ish fields of a keyreg txn to be bytes

Previously, these fields were required to be base64 encoded, which
seems like a strange requirement.

Added a few doc string cleanups.


algosdk/transaction.py
```diff
@@ -440,17 +440,17 @@ class KeyregTxn(Transaction):
...
+        selkey (byte[32])
         votefst (int)
         votelst (int)
         votekd (int)
         type (str)
         lease (byte[32])
         rekey_to (str)
         nonpart (bool)
-        sprfkey (str)
+        sprfkey (byte[64])
```
jasonpaulos, line 453: Expanding the input types to support bytes is a no-brainer, but I'm reluctant to change the attribute types as I view that as a breaking change.

The reality is this SDK uses base64 encoded strings to shuttle bytes around in a lot of places (e.g. public/private keys as well), and it's really in need of an upgrade to bytes across the board.

algosdk/transaction.py
```diff
@@ -440,17 +440,17 @@ class KeyregTxn(Transaction):
...
     """
 
     def __init__(
```
jasonpaulos, line 456: minor: it'd be nice to annotate the args with `Union[bytes, bytearray, str]` so it's clearer what types are accepted (though I realize this class does not do any type annotating)

algosdk/transaction.py
```diff
@@ -520,6 +520,19 @@ def __eq__(self, other):
             and self.sprfkey == other.sprfkey
         )
 
+    @staticmethod
+    def fixed_bytes(key, size):
+        if key is None:
+            return None
+        if isinstance(key, (bytes, bytearray)) and len(key) == size:
+            return key
```
jasonpaulos, line 528: I'd want to see `bytearray` coerced to `bytes` here if we were keeping the return value of this function in an attribute (for consistency and immutability). Though this isn't applicable if we don't change the attributes

jannotti:
going to close this an re-open with a different strategy

PR #522, opened 2023-12-11
DX: Keyreg bytes

jannotti:
Allow the various key-ish fields of a keyreg txn to be bytes

Previously, these fields were required to be base64 encoded, which
seems like a strange requirement.

Added a few doc string cleanups.


algosdk/transaction.py
```diff
@@ -520,6 +520,18 @@ def __eq__(self, other):
             and self.sprfkey == other.sprfkey
         )
 
+    @staticmethod
+    def fixed_bytes64(key, size):
```
jasonpaulos, line 524: Very minor, but prefixing this with `_` would clearly mark this as an internal function

jasonpaulos (approved):
Looks good and nice code cleanup. Consider my suggestion optional but encouraged

algosdk/transaction.py
jannotti: ok, I'll hold my nose and do it.

leading underscore to fake privacy in a language that doesn't have it is one my least favorite traditions.

algosdk/transaction.py
```diff
@@ -520,6 +520,18 @@ def __eq__(self, other):
             and self.sprfkey == other.sprfkey
         )
 
+    @staticmethod
+    def _fixed_bytes64(key, size):
+        if key is None:
+            return None
+        if isinstance(key, (bytes, bytearray)) and len(key) == size:
+            return base64.b64encode(key)
+        if len(base64.b64decode(key)) == size:
```
algorandskiy, line 529: maybe also add `isinstance(key, str)` otherwise `b64decode` will probably throw an exception (did not check tho)

algorandskiy (approved):

algosdk/transaction.py
jannotti: I suppose it will, but that's life in a dynamic language, I think.  Before you could have stuffed something crazy in that field, and you'd get an exception somewhere when we tried to serialize.  Now it happens earlier, which I think is better.

But I don't want to try to turn Python into Go and insert type checks all over the place.

PR #524, opened 2024-01-06
Fix typographic error when printing offline participation transaction

hsoerensen:
This fixes a typographical error when intending to print offline participation.


jannotti (approved):
That's certainly correct.

Issue #526, opened 2024-01-20
algod_request method hangs forever on network change


grzracz:
### Subject of the issue

When network is changed or momentarily dropped, the request will hang forever because a timeout is not set in algod_request

### Steps to reproduce

1. Send any request using any algod method
2. Change your network (by for example connecting to a VPN)

### Expected behaviour

Request fails after timeout

### Actual behaviour

Request hangs infinitely, blocking the execution of the program

### Solution

Change this line in AlgodClient.algod_request from:

`resp = urlopen(req)` to `resp = urlopen(req, timeout=10)`

10 seconds default seems good to me. Could also be 30 seconds since its a standard.

grzracz:
Fixed in #527

PR #527, opened 2024-01-20
fix: no timeout for urlopen issue #526

grzracz:


algosdk/kmd.py
```diff
@@ -35,6 +35,7 @@ def kmd_request(self, method, requrl, params=None, data=None):
             requrl (str): url for the request
             params (dict, optional): parameters for the request
             data (dict, optional): data in the body of the request
+            timeout (int, optional): request timeout in seconds
```
jannotti, line 38: Let's document the 30 second default here.

algosdk/v2client/algod.py
```diff
@@ -72,6 +73,7 @@ def algod_request(
             data (bytes, optional): data in the body of the request
             headers (dict, optional): additional header for request
             response_format (str, optional): format of the response
+            timeout (int, optional): request timeout in seconds
```
jannotti, line 76: And here.

jannotti (commented):
`urlopen` docs say: The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt (if not specified, the global default timeout setting will be used). Any idea what that global is?

Regardless, it seems right to me to let the caller have access to the setting. I'd like @jasonpaulos to weigh in.

Another possible is to allow arbitrary `**kwargs`, and pass them through, if there are other useful features of `urlopen` that it would give access to. It doesn't seem like it though, the rest appears to about SSL details.

grzracz:
Seems like the default timeout is implementation-dependent and in Python it just never times out:
https://stackoverflow.com/questions/29649173/what-is-the-global-default-timeout

algosdk/kmd.py
grzracz: Would be inconsistent since "response_format" is default "json" and is not documented in the docstring either

jannotti: Agreed, the signature shows in the docs, so the defaults are visible.

PR #530, opened 2024-01-20
txns: Uses sp.min_fee if available

jannotti:
Rather than use a hard coded constant for min_fee, use the one that is returned in SuggestedParams if available.

Since some people may create a SuggestedParams by hand rather than get one from server (because our APIs require one even though the caller may feel they "know better"), continue to use the constant if the sp.min_fee is None.

First noted by @plasmatech8 in #299

### DIFF
--- a/algosdk/transaction.py
+++ b/algosdk/transaction.py
@@ -368,9 +368,8 @@ class PaymentTxn(Transaction):
             raise error.WrongAmountType
         self.close_remainder_to = close_remainder_to
         if not sp.flat_fee:
-            self.fee = max(
-                self.estimate_size() * self.fee, constants.min_txn_fee
-            )
+            mf = sp.min_fee or constants.min_txn_fee
+            self.fee = max(self.estimate_size() * self.fee, mf)
 
     def dictify(self):
         d = dict()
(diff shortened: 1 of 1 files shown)


algosdk/transaction.py
```diff
@@ -368,9 +368,8 @@ def __init__(
             raise error.WrongAmountType
         self.close_remainder_to = close_remainder_to
         if not sp.flat_fee:
-            self.fee = max(
-                self.estimate_size() * self.fee, constants.min_txn_fee
-            )
+            mf = sp.min_fee or constants.min_txn_fee
```
jasonpaulos, line 371: I believe `constants.min_txn_fee` will be used if `sp.min_fee` is 0? I'd rather only fall back to the constant if `sp.min_fee` is None

jasonpaulos (commented):
This is a good change, and I'm currently doing something very similar to the javascript v3 branch (except there I'm willing to go a step further and require min fee in the suggested params object, since it's a new major version. I will probably leave the constant exposed, but it wouldn't be used internally at all)

algosdk/transaction.py
jasonpaulos: I believe flat fee in both of these SDKs is a boolean which indicates if fee should be taken as-is (even if it's below min fee) or interpreted as a fee-per-byte.

So there are two routes to creating a 0 fee transaction
1. Specify `fee=0` and `min_fee=0`
2. Or, specify `fee=0` and `flat_fee=True` (`min_fee` doesn't matter in this case)

>  I worried people might make SuggestedParams objects with 0 when they meant "I don't care about that"

IMO omitting it completely or setting it to None is more indicative of not caring than setting it to 0 explicitly. In a language like python where any value can be None, I'd rather use that as a special case than 0

jasonpaulos (approved):

PR #533, opened 2024-02-26
Fix: Fix indexer sync issue in cucumber tests

jasonpaulos:
Implements test changes from https://github.com/algorand/algorand-sdk-testing/pull/309

### DIFF
--- a/examples/indexer.py
+++ b/examples/indexer.py
@@ -1,7 +1,12 @@
 import json
 from algosdk import transaction
 from algosdk.v2client import indexer
-from utils import get_accounts, get_algod_client, get_indexer_client
+from utils import (
+    get_accounts,
+    get_algod_client,
+    get_indexer_client,
+    indexer_wait_for_round,
+)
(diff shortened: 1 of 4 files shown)


examples/indexer.py
```diff
@@ -41,11 +46,8 @@
     algod_client, algod_client.send_transaction(ptxn.sign(acct.private_key)), 4
 )
 
-# sleep for a couple seconds to allow indexer to catch up
-import time
-
-time.sleep(2)
-
+# allow indexer to catch up
+indexer_wait_for_round(indexer_client, res["confirmed-round"], 30)
```
algorandskiy, line 50: why 30?

algorandskiy (approved):

examples/indexer.py
jasonpaulos: In local testing the maximum number of retries I saw was 17. 30 seems like a safe upper bound before bailing out

PR #534, opened 2024-02-28
Testing: Add Python 3.12 to test matrix

jasonpaulos:
Add Python 3.12 to the test matrix


onetechnical (approved):
When do we deprecate python versions?

jasonpaulos:
Perhaps once 3.8 goes EOL in October? https://devguide.python.org/versions/

PR #537, opened 2024-04-22
Simulate: Support newer simulate options

### LINKED ISSUE #519: Support latest simulate features
Support the latest simulate features, similar to this JS PR: https://github.com/algorand/js-algorand-sdk/pull/818

### LINKED ISSUE #536: Missing allow_unnamed_resources in simulate
## Problem

We need to be able to call simulate without knowing the box references like the other sdks.

## Solution

Add the allow_unnamed_resources flag into the SimulateRequest method.

## Dependencies

I don't believe so.

## Urgency

The urgency is pretty high because the sdk is unusable for the majority of readonly method calls on smart contracts that have unknown box requirements. It is also very important for any method call when you don't know the box resources, because you need to do a simulate call first to get them.

jasonpaulos:
This SDK is behind on support for simulate options. This PR adds all simulate request options that are currently available in algod.

Similar to this JS PR: https://github.com/algorand/js-algorand-sdk/pull/818

Closes #519, closes #536

### DIFF
--- a/tests/steps/application_v2_steps.py
+++ b/tests/steps/application_v2_steps.py
@@ -701,7 +701,7 @@ def abi_method_adder(
             b"I should be unique thanks to this nonce: "
             + context.nonce.encode()
         )
-    
+
     boxes = None
     if comma_separated_boxes_string is not None:
         boxes = split_and_process_boxes(comma_separated_boxes_string)
@@ -747,7 +747,10 @@ def abi_method_adder(
         exception_key == "none"
     ), f"should have encountered an AtomicTransactionComposerError keyed by '{exception_key}', but no such exception has been detected"
 
-@when('I add a method call with the transient account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments, boxes "{boxes}".')
+
+@when(
+    'I add a method call with the transient account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments, boxes "{boxes}".'
+)
 def add_abi_method_call_with_boxes(context, operation, boxes):
     abi_method_adder(
         context,
@@ -755,6 +758,7 @@ def add_abi_method_call_with_boxes(context, operation, boxes):
         operation=operation,
         comma_separated_boxes_string=boxes,
     )
+
 
 @step(
     'I add a method call with the {account_type} account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments; any resulting exception has key "{exception_key}".'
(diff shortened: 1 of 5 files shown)


tests/steps/application_v2_steps.py
```diff
@@ -24,6 +24,8 @@ def operation_string_to_enum(operation):
         return transaction.OnComplete.NoOpOC
     elif operation == "create":
         return transaction.OnComplete.NoOpOC
+    elif operation == "create-and-optin":
```
jannotti, line 27: How did this come to be called "create-and-optin"?

tests/steps/other_v2_steps.py
```diff
@@ -1636,6 +1642,220 @@ def compare_avm_value_with_string_literal(string_literal, avm_value):
...
+        assert len(txn_group_path_split) > 0
+
+        traces = simulation_response["txn-groups"][0]["txn-results"][
+            txn_group_path_split[0]
+        ]["exec-trace"]
+        assert traces
+
+        for i in range(1, len(txn_group_path_split)):
+            traces = traces["inner-trace"][txn_group_path_split[i]]
+            assert traces
```
jannotti, line 1771: `for p in txn_group_path_split[1:]:` ?

tests/steps/other_v2_steps.py
```diff
@@ -1636,6 +1642,220 @@ def compare_avm_value_with_string_literal(string_literal, avm_value):
...
+        int(p) for p in txn_group_path.split(",") if p != ""
+    ]
+    assert len(txn_group_path_split) > 0
+
+    traces = simulation_response["txn-groups"][0]["txn-results"][
+        txn_group_path_split[0]
+    ]["exec-trace"]
+    assert traces
+
+    for i in range(1, len(txn_group_path_split)):
```
jannotti, line 1842: Simpler iteration?

jannotti (approved):

tests/steps/application_v2_steps.py
jasonpaulos: I added it when adding this cucumber test: https://github.com/algorand/algorand-sdk-testing/blob/7f06a6bd5fa352f87d62e108e2dc35bf05021343/features/integration/simulate.feature#L212

For legacy reasons, this step has "operations" which are the standard on completion options, and a "create" operation, which means noop but also don't include any app ID. I needed an app create that also opted in, so I chose to define another "operation" type that did that. Not the best solution but it was the easiest

tests/steps/other_v2_steps.py
jasonpaulos: 9a7f9e4

tests/steps/other_v2_steps.py
jasonpaulos: 9a7f9e4

PR #570, opened 2025-05-22
API: Algorand 4.1.x API updates support, includes enhanced Application Box…

gmalouf:
Algorand 4.1.x API updates support, includes enhanced Application Box search.

### DIFF
--- a/algosdk/v2client/models/application_params.py
+++ b/algosdk/v2client/models/application_params.py
@@ -20,6 +20,7 @@ class ApplicationParams(object):
         "local_state_schema": "ApplicationStateSchema",
         "global_state_schema": "ApplicationStateSchema",
         "global_state": "list[TealKeyValue]",
+        "version": "int",
     }
 
     attribute_map = {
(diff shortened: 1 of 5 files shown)


algosdk/v2client/models/application_params.py
```diff
@@ -193,6 +199,29 @@ def global_state(self, global_state):
...
+        """Gets the version of this Application programs.  # noqa: E501
+
+        Represents the number of updates to the application programs.  # noqa: E501
+
+        :return: The version of this Application programs.  # noqa: E501
+        :rtype: int
+        """
+        return self._version
+
+    @version.setter
```
jannotti, line 213: Does version need a setter?

jannotti (approved):

algosdk/v2client/models/application_params.py
gmalouf: It gets used in the initializer, I was just following the pattern from other fields, perhaps could avoid?

jannotti: Nah, stay with the pattern.

PR #572, opened 2025-05-29
API: Revert application boxes pagination

gmalouf:


tests/steps/application_v2_steps.py
```diff
@@ -1243,10 +1243,7 @@ def check_all_boxes(context, from_client: str, box_names: str = None):
             base64.b64decode(box_encoded)
             for box_encoded in box_names.split(":")
         ]
-    expected_num_boxes = len(expected_box_names)
```
algorandskiy, line 1246: looks like this and below could retain, it is not in boxes support original PR: https://github.com/algorand/py-algorand-sdk/pull/570/files#diff-e07bff20fd1563a5ec2f7fd913dc42c0983e4b2e51d59433bb4fb00a9e7fe9ae

PR #594, opened 2026-07-17
Fix: Wallet.delete_key/delete_multisig return False for non-existent addresses

IcanBENCHurCAT:
## Issue

Fixes GitHub issue #472: `Wallet.delete_key()` and `Wallet.delete_multisig()` now correctly return `False` when the specified address does not exist in the wallet, instead of always returning `True`.

## Problem

The original implementation of `delete_key()` and `delete_multisig()` in `algosdk/wallet.py` did not verify that the address existed before calling the KMD delete endpoint. The KMD API returned a success response even for non-existent keys, so the Python wrapper always returned `True`.

## Solution

Add an existence check using `list_keys()` / `list_multisig()` before calling the KMD delete endpoint.

## Changes

- **algosdk/wallet.py**: Add existence check in `delete_key()` and `delete_multisig()`; updated docstrings
- **tests/unit_tests/test_wallet.py**: New test file with 6 unit tests

## Testing

All 157 existing unit tests pass. No regressions.

### DIFF
--- a/algosdk/wallet.py
+++ b/algosdk/wallet.py
@@ -149,9 +149,12 @@ class Wallet:
             address (str): base32 address of account to be deleted
 
         Returns:
-            bool: True if the account has been deleted
+            bool: True if the account has been deleted,
+                False if the address was not found in the wallet
         """
         self.automate_handle()
+        if address not in self.list_keys():
+            return False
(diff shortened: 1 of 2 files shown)


IcanBENCHurCAT:
Closing for now — will reopen if this fix is still wanted upstream.

PR #595, opened 2026-07-23
feat: add post-quantum (Falcon-1024) signature support

mrcointreau:
## Summary

Adds post-quantum signing support to the Python SDK, ported to feature parity with the `js-algorand-sdk` `feat/pq` branch. It lets accounts and delegated LogicSigs authorize transactions with a Falcon-1024 signature, and introduces a callback-based signer model ("bring your own key backend": HD wallet, hardware device, KMS, an external Falcon implementation) alongside the existing secret-key APIs.

The SDK does not bundle a Falcon implementation. As in the JS branch, the developer supplies a small "raw signer" callback that produces the signature; the SDK handles address derivation, domain separation, wire encoding, and transaction assembly.

Wire-compatible with go-algorand: the shared signing vectors under `tests/unit_tests/pq_test_data/` are byte-identical to the ones in the JS branch (both vendor the fixtures the go-algorand data factory generated and verified), and the suite asserts exact-byte equality against them.

The design follows the [PQ SDK Implementation Guide](https://gist.github.com/joe-p/6bb0dae74dee2c9d919bb51335b3571f) that the SDK implementations share: the SHA-512/256 address derivation with canonical-salt selection and Ed25519-point rejection, the `PQA` / `PQK` / `PQProgram` domain-separation prefixes, the `f1` scheme identifier, and the transaction and logic-sig signing preimages all conform.

Two points have moved on since that gist was last updated (2026-07-02), and this branch tracks go-algorand rather than the gist on both:

- **No pre-hash.** [algorand/go-algorand#6669](https://github.com/algorand/go-algorand/pull/6669) ("falcon: sign ToBeHashed, not Hashed") removed the outer SHA-512/256 that wrapped the post-quantum payloads, so Falcon signs the preimage directly, the same way ed25519 does. The gist still specifies `SHA512_256("TX" || encodedTransaction)` for transactions and `SHA512_256("PQProgram" || address[32] || lsigProgram)` for logic sigs, and states that this is where PQ differs from ed25519. After #6669 there is no such difference, and neither payload is hashed by the SDK.
- **Wire key is `pqsig`.** The gist describes `pq` as the current-phase key, noting it may later become `pqsig`. It has. This branch uses `pqsig`, matching both the regenerated go-algorand vectors and the JS branch.

## What it adds

Public API (all reachable via their documented submodule paths):

- `algosdk.encoding`
  - `address_from_pq_key(scheme, public_key) -> (address, salt)`: derive a PQ account address and its canonical salt.
  - `is_ed25519_point(public_key) -> bool`: the broad curve-point predicate used by salt selection (matches go-algorand's `basics.IsEdwards25519Point`).
- `algosdk.mnemonic`
  - `to_pq_seed(mnemonic, scheme) -> bytes`: derive a PQ key seed from a 25-word mnemonic.
- `algosdk.transaction`
  - `PQSig`, `PQSignedTransaction`, `LogicSig.pqsig`, plus `LogicSigAccount.sign_with_signer` / `append_to_multisig_with_signer`.
- `algosdk.signer` (new module)
  - `RawSigner` (callback type), `Ed25519TransactionSigner`, `Ed25519MultisigTransactionSigner`, `PQTransactionSigner`, `Falcon1024TransactionSigner`.
- `algosdk.atomic_transaction_composer`
  - `sign_transaction_with_signer(txn, signer)`: single-transaction convenience over any `TransactionSigner`.
- `algosdk.constants`: `PQ_ADDRESS_PREFIX`, `PQ_SEED_PREFIX`, `PQ_PROGRAM_PREFIX`, `PQ_SCHEME_LEN`, `FALCON_1024_SCHEME`, `PQSIG_KEY` (plus lowercase aliases).
- `algosdk.error`: `PQSchemeLengthError`, `NoCanonicalSaltError`, `PQMultisigUnsupportedError`.

Covered flows: PQ address derivation, Falcon transaction signing (direct and rekeyed, with the `sgnr` auth address), PQ-delegated LogicSigs, the `pqsig` msgpack wire round-trip, and mnemonic-to-seed derivation.

## Usage

The developer supplies the Falcon-1024 backend as a `RawSigner` callback (`Callable[[bytes], bytes]`). The SDK builds the signing preimage and hands the callback those exact bytes, unhashed; the callback performs whatever hashing its scheme requires internally and returns the raw detached signature. This is the same contract the ed25519 callback signer uses.

```python
from algosdk import transaction
from algosdk.signer import Falcon1024TransactionSigner
from algosdk.atomic_transaction_composer import sign_transaction_with_signer

# falcon_sign is your Falcon-1024 backend: given the exact preimage bytes the
# SDK builds, it returns the raw signature bytes.
signer = Falcon1024TransactionSigner(public_key, falcon_sign)

txn = transaction.PaymentTxn(signer.address, sparams, receiver, 0)
signed = sign_transaction_with_signer(txn, signer)  # -> PQSignedTransaction
```

`Falcon1024TransactionSigner` is a `TransactionSigner`, so it also plugs directly into `AtomicTransactionComposer` via `add_transaction(TransactionWithSigner(txn, signer))`.

The preimages the callback can receive are:

| what is being signed | preimage handed to the callback |
|---|---|
| transaction | `"TX" + msgpack(txn)` |
| delegated LogicSig | `"PQProgram" + pq_address[32] + program` |

## Relationship to js-algorand-sdk feat/pq

This is a port of the JS `feat/pq` branch. Same protocol behavior, same wire format, same shared test vectors. The differences below are intentional and idiomatic to Python; none change the wire format.

### Idiomatic differences a reviewer should know

- **Class-based signers vs functional factories.** JS models a signer as a function and returns capability bundles from factories (e.g. `addressWithSignersFromRawFalcon1024Signer`). Python subclasses `TransactionSigner` (consistent with the existing `Account`/`LogicSig`/`Multisig` signers). Practical consequence for a reviewer reading both: in JS you pass the `.txnSigner` function out of the bundle to the composer; in Python you pass the signer object itself, because the objects are `TransactionSigner` instances.
- **Sync, not async.** Python signers are synchronous; JS's are `async`.
- **Separate PQ signed-transaction type.** Python adds a distinct `PQSignedTransaction` class (alongside `SignedTransaction` / `MultisigTransaction` / `LogicSigTransaction`), whereas JS adds a `pqsig` field to its single unified `SignedTransaction`. The wire bytes are identical either way, and `msgpack_decode` / `retrieve_from_file` return the right type automatically.
- **Generic base plus Falcon specialization.** `PQTransactionSigner(pk, signer, scheme)` is the generic base and `Falcon1024TransactionSigner(pk, signer)` is the scheme-fixed subclass. This mirrors JS's `pq-signer.ts` + `falcon-signer.ts` split, with JS using its generic PQ signer internally and Python exporting both. The Falcon class name carries `1024` to match JS's `Falcon1024SigningKey` / `...RawFalcon1024Signer` and this repo's `FALCON_1024_SCHEME` constant.
- **One delegation method instead of two.** Python's `LogicSigAccount.sign_with_signer(signer, multisig=None)` covers what JS splits into `signWithSigner` and `signMultisigWithSigner`. There is no `sign_multisig_with_signer`.
- **Program-data signing location.** Python exposes it on the signer (`Ed25519TransactionSigner.sign_program_data(data, lsig)`) rather than on the LogicSig (`LogicSig.signDataWithSigner` in JS).
- **`sign_bytes` returns base64.** `Ed25519TransactionSigner.sign_bytes` returns a base64 `str`, diverging from the gist's `MxBytesSigner => Promise<Uint8Array>` and JS's raw `Uint8Array`. JS returns raw because its own `signBytes` and `verifyBytes` are raw; Python's `util.sign_bytes` returns base64 and `util.verify_bytes` consumes base64, so returning raw bytes made the signer's output fail `util.verify_bytes` silently, returning `False` rather than raising. Every other method on the class already encodes to match its Python counterpart (`sign_transactions` and single-key `sign_logicsig` return base64; `sign_program_data` and the multisig subsig paths return raw bytes).
- **Error types follow each repo's existing convention.** Python raises `PQSchemeLengthError` / `NoCanonicalSaltError` / `PQMultisigUnsupportedError`, joining the typed exceptions already in `algosdk.error`; JS throws `Error`, consistent with how it reports errors elsewhere.
- **Naming that follows each repo's own convention.** `to_pq_seed` (Python's `to_*` mnemonic family) vs JS's `pq25WordMnemonicToSeed`; `PQSig` (no `Encoded` prefix, matching Python's `Multisig`) vs JS's `EncodedPQSig`; `PQ_*_PREFIX` constants (matching `LOGIC_PREFIX`) vs JS's `*_TAG`.

### Differences in what each branch exposes

- `to_pq_seed` raises `PQSchemeLengthError` on a wrong-length scheme, for consistency with `address_from_pq_key` in the same module. JS's `pq25WordMnemonicToSeed` leaves the check to the caller.
- `is_ed25519_point` is exported from `algosdk.encoding`. Its JS counterpart, `couldBeCurvePoint`, lives in `src/utils/ed25519-check.ts` and is used internally rather than re-exported from the package index.
- `encoding.msgpack_decode` decodes a standalone `PQSig`, dispatching on its `sch` key ahead of `sig`. This is specific to how the two decoders work: Python infers the target type from the payload, so a bare `PQSig` needs a discriminator to tell it apart from a `SignedTransaction`, while JS's `decodeMsgpack` is handed an explicit target class and so has no equivalent path.

## Deprecations (non-breaking)

The secret-key signing methods are documented as deprecated in favor of the signer-based APIs, but remain fully functional:

- `Transaction.sign`, `Transaction.raw_sign`
- `MultisigTransaction.sign`
- `LogicSig.sign` / `append_to_multisig` / `sign_program` / `multisig_sign_program` / `single_sig_multisig`
- `LogicSigAccount.sign` / `sign_multisig` / `append_to_multisig`

These are docstring-only deprecations (no runtime warning, no `@deprecated` marker). As in the JS branch, internal code (the recommended `AccountTransactionSigner` / `MultisigTransactionSigner` paths) still calls the deprecated methods; adding a machine-readable marker was deliberately left out to avoid a new dependency.

## Testing

- The ed25519 point predicate is pinned by `lsig_address_kat.json`, the known-answer vectors published by [falcon-signatures](https://github.com/algorandfoundation/falcon-signatures), including the adversarial encodings (small-order points, non-canonical field elements) on which widely used crypto libraries disagree with go-algorand. The JS branch vendors the same file.
- Unit tests in `tests/unit_tests/test_pq.py` and `test_signer.py`, mapped case-by-case to the JS `12.PQ.ts` and `11.Ed25519Signer.ts` suites, asserting exact-byte equality against the shared vectors plus decode and re-encode round-trips.
- The vectors under `tests/unit_tests/pq_test_data/` were regenerated against a go-algorand build that includes #6669, so they pin both the unhashed payload and the `pqsig` key. `update.sh` documents how to refresh them and which submodule pin determines the protocol behavior they capture.
- Passes `make lint` (black line length 79, mypy) and the unit suite.

## Not included (by design)

- No bundled Falcon-1024 implementation. The developer supplies the `RawSigner` callback, exactly as the JS branch relies on the `falcon-1024` package only in `examples/falcon.ts` and `tests/12.PQ.ts`.
- No runnable example script (the standard Sandbox does not support PQ transactions, and there is no vetted Python Falcon backend to depend on).


algosdk/atomic_transaction_composer.py
```diff
@@ -13,7 +13,7 @@
     cast,
 )
 
-from algosdk import abi, error, transaction
+from algosdk import abi, constants, encoding, error, transaction
```
iglosiggio: These two are unused right?
```suggestion
from algosdk import abi, error, transaction
```

algosdk/encoding.py
```diff
@@ -79,8 +79,14 @@ def msgpack_decode(enc):
...
             return transaction.LogicSigTransaction.undictify(decoded)
         return transaction.LogicSigAccount.undictify(decoded)
+    if "sch" in decoded:
+        # A standalone PQSig also carries a "sig" key, so this check must
+        # come before the SignedTransaction dispatch.
+        return transaction.PQSig.undictify(decoded)
     if "sig" in decoded:
         return transaction.SignedTransaction.undictify(decoded)
+    if constants.pqsig_key in decoded:
+        return transaction.PQSignedTransaction.undictify(decoded)
```
iglosiggio: Either use constants for everyone or literal strings for everyone.

I prefer the later one, if you agree then remove `pqsig_key` and `PQSIG_KEY` from the `algosdk.constants` module and then
```suggestion
    if "pqsig" in decoded:
        return transaction.PQSignedTransaction.undictify(decoded)
```

algosdk/error.py
```diff
@@ -227,3 +227,26 @@ def __init__(self, version):
...
             "Only SourceMap version 3 is supported, got: {}".format(version),
         )
+
+
+class PQSchemeLengthError(Exception):
+    def __init__(self, length):
+        Exception.__init__(
+            self,
+            "post-quantum scheme must be 2 bytes, got {}".format(length),
+        )
```
iglosiggio, line 237: Sincere question, why do it this way instead of:
```suggestion
        super().__init__("post-quantum scheme must be 2 bytes, got {}".format(length))
```
Am I missing something?

algosdk/error.py
```diff
@@ -227,3 +227,26 @@ def __init__(self, version):
...
+            self,
+            "no canonical salt exists for this public key and scheme",
+        )
+
+
+class PQMultisigUnsupportedError(Exception):
+    def __init__(self):
+        Exception.__init__(
+            self, "post-quantum signatures do not support multisig"
+        )
```
iglosiggio, line 252: Same Q for these two.

algosdk/signer.py
```diff
@@ -0,0 +1,357 @@
...
+    and is responsible for any hashing the scheme itself requires.
+
+    For Falcon-1024 use the `Falcon1024TransactionSigner` subclass, which fixes
+    the scheme for you.
+
+    Args:
+        public_key (bytes): the scheme's public key
+        signer (RawSigner): callback that signs exact preimage bytes and
+            returns the raw post-quantum signature
+        scheme (bytes): 2-byte scheme identifier (e.g. b"f1" for Falcon-1024)
```
iglosiggio, line 290: I would design it differently so the scheme is tied to the signer (we can make the signed a lightweight class or a protocol for a callable with a `.scheme` property.

algosdk/signer.py
```diff
@@ -0,0 +1,357 @@
...
+    "RawSigner",
+    "Ed25519TransactionSigner",
+    "Ed25519MultisigTransactionSigner",
+    "PQTransactionSigner",
+    "Falcon1024TransactionSigner",
+]
+
+# A low-level signing callback: signs the exact preimage bytes and returns the
+# raw signature.
+RawSigner = Callable[[bytes], bytes]
```
iglosiggio, line 39: Alternative design: `RawSigner` is a protocol with `__call__(encoded: [bytes]) -> bytes `and `scheme: PQScheme`. Then `alias PQScheme = Literal["f1"]` but it can be later updated to `Literal["f1"] | Literal["XX"]`.

algosdk/transaction.py
```diff
@@ -132,6 +132,8 @@ def sign(self, private_key):
         """
         Sign the transaction with a private key.
 
+        Deprecated: use `sign_transaction_with_signer(txn, AccountTransactionSigner(private_key))` instead.
```
iglosiggio, line 141: Is this soft-deprecated or hard-deprecated?

If it is hard deprecated then https://docs.python.org/3/library/warnings.html#warnings.warn should be used :)

See https://peps.python.org/pep-0387/#soft-deprecation for more info

Alternatively you can also use the `@deprecated` decorator:
- https://typing.python.org/en/latest/spec/directives.html#deprecated
- https://docs.python.org/3.14/library/warnings.html#warnings.deprecated

algosdk/transaction.py
```diff
@@ -162,19 +164,32 @@ def _sign_and_skip_rekey_check(self, private_key):
...
+        """
+        txn = encoding.msgpack_encode(self)
+        return constants.txid_prefix + base64.b64decode(txn)
+
     def raw_sign(self, private_key):
         """
         Sign the transaction.
 
+        Deprecated: use `sign_transaction_with_signer(txn, AccountTransactionSigner(private_key)).signature` instead.
+
```
iglosiggio, line 207: Same comment as before

algosdk/transaction.py
```diff
@@ -2246,6 +2320,68 @@ def __eq__(self, other):
...
+        Get the transaction's ID.
+
+        Returns:
+            str: transaction ID
+        """
+        return self.transaction.get_txid()
+
+    def dictify(self):
+        od = OrderedDict()
+        od[constants.pqsig_key] = self.pqsig.dictify()
```
iglosiggio: Same as the first comment on constants. Maybe a literal string is ok here :)

```suggestion
        od["pqsig"] = self.pqsig.dictify()
```

algosdk/transaction.py
```diff
@@ -2274,6 +2410,8 @@ def sign(self, private_key):
         """
         Sign the multisig transaction.
 
+        Deprecated: use `MultisigTransactionSigner(multisig, [private_key])` instead.
+
```
iglosiggio: Same as the previous comment on deprecations.

algosdk/transaction.py
```diff
@@ -2630,6 +2769,8 @@ def dictify(self):
             od["msig"] = self.msig.dictify()
         elif self.lmsig:
             od["lmsig"] = self.lmsig.dictify()
+        elif self.pqsig:
+            od[constants.pqsig_key] = self.pqsig.dictify()
```
iglosiggio: Alternative:
```suggestion
        elif self.pqsig:
            od["pqsig"] = self.pqsig.dictify()
```

algosdk/transaction.py
```diff
@@ -2641,13 +2782,16 @@ def undictify(d):
             lsig.msig = Multisig.undictify(d["msig"])
         elif "lmsig" in d:
             lsig.lmsig = Multisig.undictify(d["lmsig"])
+        elif constants.pqsig_key in d:
+            lsig.pqsig = PQSig.undictify(d[constants.pqsig_key])
```
iglosiggio: Alternative:
```suggestion
        elif constants.pqsig_key in d:
            lsig.pqsig = PQSig.undictify(d["pqsig"])
```

algosdk/transaction.py
```diff
@@ -2715,6 +2871,9 @@ def sign_program(program, private_key):
 
     @staticmethod
     def multisig_sign_program(program, private_key, multisig):
+        """
+        Deprecated: use `LogicSigAccount.sign_with_signer` instead.
+        """
```
iglosiggio, line 2969: Same as the previous comment on deprecations. Also: if this is deprecated because the new way is preferred It would be nice to implement this in terms of the new API :)

algosdk/transaction.py
```diff
@@ -2727,6 +2886,9 @@ def multisig_sign_program(program, private_key, multisig):
 
     @staticmethod
     def single_sig_multisig(program, private_key, multisig):
+        """
+        Deprecated: use `LogicSigAccount.sign_with_signer` instead.
+        """
```
iglosiggio, line 2993: Same as previous comment

algosdk/transaction.py
```diff
@@ -2744,6 +2906,8 @@ def sign(self, private_key, multisig=None):
         """
         Creates signature (if no pk provided) or multi signature
 
+        Deprecated: use `LogicSigAccount.sign_with_signer` instead.
+
```
iglosiggio, line 3021: Same as previous comment

algosdk/transaction.py
```diff
@@ -2772,6 +2936,8 @@ def append_to_multisig(self, private_key):
         """
         Appends a signature to multi signature
 
+        Deprecated: use `LogicSigAccount.append_to_multisig_with_signer` instead.
+
```
iglosiggio, line 3066: Same as previous comment.

algosdk/transaction.py
```diff
@@ -2900,6 +3072,8 @@ def sign_multisig(self, multisig: Multisig, private_key: str) -> None:
         of another account, called the delegating account. Use this function if
         the delegating account is a multisig account.
 
+        Deprecated: use `sign_with_signer(signer, multisig=...)` instead.
+
```
iglosiggio, line 3232: Same as previous comment.

algosdk/transaction.py
```diff
@@ -2919,6 +3093,8 @@ def append_to_multisig(self, private_key: str) -> None:
         Adds an additional signature from a member of the delegating multisig
         account.
 
+        Deprecated: use `append_to_multisig_with_signer(signer)` instead.
+
```
iglosiggio, line 3257: Same as previous comment.

algosdk/transaction.py
```diff
@@ -2937,6 +3113,8 @@ def sign(self, private_key: str) -> None:
         of another account, called the delegating account. If the delegating
         account is a multisig account, use `sign_multisig` instead.
 
+        Deprecated: use `sign_with_signer(signer)` instead.
+
```
iglosiggio, line 3281: Same as previous comment.

algosdk/transaction.py
```diff
@@ -3302,6 +3507,8 @@ def retrieve_from_file(path):
             txns.append(SignedTransaction.undictify(txn))
         elif "lsig" in txn:
             txns.append(LogicSigTransaction.undictify(txn))
+        elif constants.pqsig_key in txn:
```
iglosiggio: Alternative:

```suggestion
        elif "pqsig" in txn:
```

tests/unit_tests/test_pq.py
```diff
@@ -0,0 +1,457 @@
...
+        self.assertIsInstance(salt, int)
+        self.assertTrue(0 <= salt <= 255)
+
+    def test_derived_address_is_well_formed(self):
+        # a PQ-derived address is a normal 58-char Algorand address that
+        # round-trips through decode/encode
+        pubkey = base64.b64decode(_load("pqMnemonic.json")["publicKey"])
+        address, _ = encoding.address_from_pq_key(
+            constants.falcon_1024_scheme, pubkey
+        )
```
iglosiggio, line 76: Note for @joe-p I didn't add a "from key pair" way to create falcon accounts on the go-sdk. Is it expected?

On the specific test: I would rename it to "canonical_address_from_pq_key" because non-canonical addresses may appear 😬

tests/unit_tests/test_pq.py
```diff
@@ -0,0 +1,457 @@
...
+        txn = encoding.msgpack_decode(fx["txnBlob"])
+        # sanity: the unsigned txn re-encodes to the fixture byte-for-byte
+        self.assertEqual(encoding.msgpack_encode(txn), fx["txnBlob"])
+        captured = {}
+
+        def fake(to_sign):
+            captured["to_sign"] = to_sign
+            return expected_sig
+
+        signer = Falcon1024TransactionSigner(pk, fake)
```
iglosiggio: I'm not sure this is a useful way to test it. Shouldn't we be testing a reference backend instead?

tests/unit_tests/test_pq.py
```diff
@@ -0,0 +1,457 @@
...
+            fx["stxn"]["lsig"]["pqsig"]["sig"],
+        )
+        self.assertEqual(encoding.msgpack_encode(stxn), fx["stxnBlob"])
+
+
+class TestPQPublicAPI(unittest.TestCase):
+    """Guards the public import paths for the post-quantum surface. `make lint`
+    does not catch a missing/renamed export, so assert every public name is
+    reachable via its documented (submodule) path -- matching how the SDK
+    exposes SignedTransaction, AccountTransactionSigner, etc."""
```
iglosiggio: This tests feels a bit weird to me. Does this SDK have similar tests elsewhere?

tests/unit_tests/test_signer.py
iglosiggio, line 1: The tests here feel ok but I think we should port the tests for the deprecated codepaths instead of using them as a reference impl (IMO a deprecation without eventual removal is not a deprecation).

iglosiggio (commented):
Changes feel mostly ok. Left some comments on deprecation/api stability and some stylistic stuff that differs from what's already on the SDK.

algosdk/encoding.py
mrcointreau: Good catch, the constant was a kind of dev-time insurance: the upstream key changed once already (from `pq` to `pqsig`, see [4541936](https://github.com/algorand/py-algorand-sdk/pull/595/commits/4541936738e5f53afb1b437beb56ee1c5e41725f)) and the constant made that a one-line migration. Key's settled now, so inlined and removed in [8716f2e](https://github.com/algorand/py-algorand-sdk/pull/595/commits/8716f2e53c07576a588b86fb4dcc3901284e4b5a).

algosdk/transaction.py
mrcointreau: Addressed in [8716f2e](https://github.com/algorand/py-algorand-sdk/pull/595/commits/8716f2e53c07576a588b86fb4dcc3901284e4b5a)

algosdk/transaction.py
mrcointreau: Addressed in [8716f2e](https://github.com/algorand/py-algorand-sdk/pull/595/commits/8716f2e53c07576a588b86fb4dcc3901284e4b5a)

algosdk/transaction.py
mrcointreau: Addressed in [8716f2e](https://github.com/algorand/py-algorand-sdk/pull/595/commits/8716f2e53c07576a588b86fb4dcc3901284e4b5a), I replaced also the `elif constants.pqsig_key` for consistency purposes

algosdk/transaction.py
mrcointreau: Addressed in [8716f2e](https://github.com/algorand/py-algorand-sdk/pull/595/commits/8716f2e53c07576a588b86fb4dcc3901284e4b5a)

algosdk/error.py
mrcointreau: AFAIK those are equivalent for a single-parent exception, I just matched the neighboring style (`SourceMapVersionError` right above uses the same form, as do most of the older errors). That said, the file is a bit of a mix: the newer Algod/ABI errors go with `super()`, the older ones all do `Exception.__init__`.

iglosiggio: Then lets keep it this way :)

tests/unit_tests/test_pq.py
mrcointreau: It's actually less "fake" than it looks :) Each fixture in `pq_test_data` is gnerated by go-algorand (via the algokit-polytest data factory, see `update.sh` in that folder) and is a complete test vector: the unsigned txn bytes, the real Falcon signature go produced for it + `stxnBlob` (the final signed transaction exactly as go encoded it). `expected_sig` is that real signature and the fake callback is just the injection point since the SDK deliberately doesn't bundle a Falcon implementation. So the test ends up checking two things: that we hand the callback exactly `bytes_to_sign()` and that the signed txn we assemble comes out byte identical to go's `stxnBlob`, which is effectively a reference backend comparison with the signature injected instead of computed in-process. The crypto itself is covered by the upstream KATs in the [falcon package](https://github.com/mrcointreau/falcon-det1024) and I have an `examples/falcon.py` draft doing the full e2e flow with that package (that's on the test pypi registry for now and should be properly published soon).

tests/unit_tests/test_pq.py
mrcointreau: Sounds good! Addressed in [4804150](https://github.com/algorand/py-algorand-sdk/pull/595/commits/480415040887db2cfd25dc16f2196156d4b5c8ae)

algosdk/signer.py
mrcointreau: Yeah that's fair, nothing stops you from using `PQTransactionSigner` directly with whatever scheme. That's intentional though: it's the extension point for future schemes (go already reserves `f2` for Falcon-512 in its fee schedule) and keeping it public means supporting a new scheme doesn't have to wait for an SDK release. For normal usage the mismatch can't happen: the documented signer (`Falcon1024TransactionSigner`) takes no scheme at all and hardcodes `b"f1"`. What I'd prefer not to lose is being able to pass a plain function as the signer: a method from the falcon package, a KMS call or a lambda for empty-signature fee estimation all plug in directly. It's also the design of the whole module, `Ed25519TransactionSigner` takes the same bare `RawSigner` callback, so a scheme-carrying signer object would make the PQ one the odd one out. If the scheme lived on the signer none of those callables would work as-is and every backend would need a small wrapper class first. The js-algorand-sdk pq branch made the same call, bare function for the raw signer and the scheme pinned by the named wrapper. A wrong scheme also fails loudly at address verification rather than doing anything silent. I'd happily take the `Literal["f1"]` typing from the comment below though!

algosdk/signer.py
mrcointreau: Implemented the `PQScheme` in [876d971](https://github.com/algorand/py-algorand-sdk/pull/595/commits/876d9719628a1c06eafb4a2a026e0105c9d815b5). I declared the alias in `constants.py` next to the scheme constants, annotating both `FALCON_1024_SCHEME` and `falcon_1024_scheme` with it and importing the alias in `signer.py` for the `scheme` parameter. That way the scheme registry stays in one place and adding Falcon-512 later means widening one union in one file. Does this shape look right to you or would yo have placed `Literal` somewhere else?

tests/unit_tests/test_pq.py
mrcointreau: Fair point, there aren't, removed in [3aaebd7](https://github.com/algorand/py-algorand-sdk/pull/595/commits/3aaebd7d791cbf385edaa2628b9e18a02b0c6cc4).

algosdk/transaction.py
mrcointreau: Yeah, the doc-only notes were me mimicking `js-algorand-sdk`, where these methods just carry a jsdoc `@deprecated` tag, your suggestions convinced me to go all the way though. [0df3799](https://github.com/algorand/py-algorand-sdk/pull/595/commits/0df3799e0704bf6036788f8dd38313e29f2ded76) adds `typing_extensions.deprecated(category=None)` everywhere (IDE strikethrough, zero runtime change) and moves the SDK internals off the deprecated methods, since the ATC signers were built on top of them and would have warned from inside the recommended path. Then [7d75004](https://github.com/algorand/py-algorand-sdk/pull/595/commits/7d75004d3c60d7b47425283c11f656c73c129545) adds `MultisigTransaction.append_with_signer` because appending had no signer replacement yet (while js has `appendSignMultisigTransactionWithSigner` for this), [d477a9c](https://github.com/algorand/py-algorand-sdk/pull/595/commits/d477a9c44d11e56869f340eee4a92f415f0c42e9) ports the examples and [bf1a19b](https://github.com/algorand/py-algorand-sdk/pull/595/commits/bf1a19b63af76d909ad476fea8a9b6892227e757) flips it all to a real `DeprecationWarning` naming the removal in a future major release. Wdyt?

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/transaction.py
mrcointreau: Answered in the `Transaction.sign` thread to keep the discussion in one place.

algosdk/encoding.py
```diff
@@ -202,6 +208,68 @@ def checksum(data):
     return chksum.digest()
 
 
+_ED25519_P = 2**255 - 19
+_ED25519_D = (-121665 * pow(121666, _ED25519_P - 2, _ED25519_P)) % _ED25519_P
+
+
+def is_ed25519_point(public_key: bytes) -> bool:
```
joe-p: nit: feels weird to have this in encoding.py

algosdk/signer.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def _subsig_index(multisig: "transaction.Multisig", public_key: bytes) -> int:
+    for i, subsig in enumerate(multisig.subsigs):
+        if subsig.public_key == public_key:
+            return i
+    raise error.InvalidSecretKeyError
+
+
+class Ed25519TransactionSigner(TransactionSigner):
```
joe-p: Rather than `Ed25519TransactionSigner` maybe `Ed25519AlgorandSigner` since the signing capabilities extend beyond just transaction signing. Same for other signers here

mrcointreau: Addressed in [9130a0c](https://github.com/algorand/py-algorand-sdk/pull/595/commits/9130a0c70e4d4613a2f2ecdd57f5741635cd2308)

algosdk/encoding.py
mrcointreau: You're right, I saw that in `js-algorand-sdk` you put that in a dedicated file under `src/utils/ed25519-check.ts`, I moved it to a dedicated `algosdk/ed25519_check.py` in here (an alternative location could be `algoksd/util.py`). Addressed in [80bd179](https://github.com/algorand/py-algorand-sdk/pull/595/commits/80bd179ac4c480e2ba522107b9c0f40f97426890).

tests/unit_tests/test_signer.py
joe-p: Going to resolve this for now. Once we've released everything across all the languages I want to go back and look at our testing strategy as a whole

algosdk/signer.py
iglosiggio: looks good to me :)

examples/README.md
```diff
@@ -9,3 +9,5 @@ Assuming a sandboxed node is running locally, any example can be run with the fo
     python3 <example_name>.py
 ```
 
+The post-quantum example, `falcon.py`, has two extra requirements: the Falcon-1024 implementation from the `algorand-falcon` package (`pip install algorand-falcon`, not a dependency of the SDK), and a node running the `future` consensus version. It skips itself when either is missing.
```
iglosiggio: The `future` is now! (pls update the docs)

```suggestion
The post-quantum example, `falcon.py`, has two extra requirements: the Falcon-1024 implementation from the `algorand-falcon` package (`pip install algorand-falcon`, not a dependency of the SDK), and a node running at least v42 consensus.
```

iglosiggio (dismissed):

joe-p (dismissed):

joe-p (approved):

examples/README.md
mrcointreau: Good catch, I removed the `future` reference in `examples/falcon.py` but missed this 😬

PR #596, opened 2026-07-31
feat: support for consensus v42

mrcointreau:


algosdk/atomic_transaction_composer.py
```diff
@@ -485,13 +493,17 @@ def add_method_call(
             local_schema (StateSchema, optional): restricts what can be stored by created application;
                 must be omitted if not creating an application
             global_schema (StateSchema, optional): restricts what can be stored by created application;
-                must be omitted if not creating an application
+                must be omitted if not creating or updating an application. On an
+                update, the node installs global_schema and extra_pages together:
+                a field left at its default sets that size to zero, so pass the
+                current value of any size that should not change
```
jannotti: I don't know if it's worth this detail, but: if you leave _both_ extra pages and global schema in default state, then there is no update of sizes at all.  This is to maintain backward compatibility with updates from the past that did not set either one.

algosdk/atomic_transaction_composer.py
```diff
@@ -485,13 +493,17 @@ def add_method_call(
...
             approval_program (bytes, optional): the program to run on transaction approval;
                 must be omitted if not creating or updating an application
             clear_program (bytes, optional): the program to run when state is being cleared;
                 must be omitted if not creating or updating an application
             extra_pages (int, optional): additional program space for supporting larger programs.
-                A page is 1024 bytes.
+                A page is 1024 bytes; must be omitted if not creating or updating
+                an application.
```
jannotti: This is probably confusing to most readers.  Yes, a page is 1024 bytes, but it applies to both approval and clear state, _and_ is pooled. So in practice, it's 2k, which is usually used entirely by the approval program.

algosdk/atomic_transaction_composer.py
```diff
@@ -775,7 +787,10 @@ def simulate(
                 transaction group, a list of txIDs of the simulated transactions,
                 an array of results for each method call transaction in this group.
                 If a method has no return value (void), then the method results array
-                will contain None for that method's return value.
+                will contain None for that method's return value. When the node
+                reports them, group_fees_paid carries the total fees the group
```
jannotti: `group_fees_paid` is much simpler.  It's how much *was* paid by this group. So it's little more than the recursive sum of Fee fields.  It's perhaps worth saying that multiplying `usage` by `minfee` is better was to estimate the fee _needed_ instead of the fee paid during simulate.

We should be _very_ clear that usage is a 6 digit fixed point number. So it will say 2100000, if 2.1 minfees are required.

algosdk/atomic_transaction_composer.py
mrcointreau: Addressed in [0247af0](https://github.com/algorand/py-algorand-sdk/pull/596/commits/0247af0ccd72e707ecf524cb1b03d885aba93e47). Worth including I think, so the entry now ends with `"Leaving both out keeps the current sizes"`.

algosdk/atomic_transaction_composer.py
mrcointreau: Addressed in [10f20af](https://github.com/algorand/py-algorand-sdk/pull/596/commits/10f20aff2826b542724c4098db7ea5693a07e709).

algosdk/atomic_transaction_composer.py
mrcointreau: Addressed in [19c6c28](https://github.com/algorand/py-algorand-sdk/pull/596/commits/19c6c28748a6c094616433ec1e2e066370abf384), did I get that right?

jannotti: Yup!

algosdk/atomic_transaction_composer.py
```diff
@@ -485,13 +495,19 @@ def add_method_call(
...
             approval_program (bytes, optional): the program to run on transaction approval;
                 must be omitted if not creating or updating an application
             clear_program (bytes, optional): the program to run when state is being cleared;
                 must be omitted if not creating or updating an application
             extra_pages (int, optional): additional program space for supporting larger programs.
-                A page is 1024 bytes.
+                A page is pooled between the approval and clear state programs, and
+                is usually spent on the approval program. Must be omitted if not
+                creating or updating an application; on an update it installs
+                together with global_schema.
```
jannotti:
```suggestion
                Each extra page grants 2048 extra bytes which are available for use by the approval or clear state programs. Must be omitted if not
                creating or updating an application; on an update it installs
                together with global_schema.
```
You probably don't want to apply my change in github, since it's hard to get whitespace right, but I think this language would be sufficient.

mrcointreau: Thank you 🙏, addressed in [dcf9ce1](https://github.com/algorand/py-algorand-sdk/pull/596/commits/dcf9ce1b0fa5c970869fdfa7d1da4b0677d605bb).

joe-p (approved):

PR #598, opened 2026-08-01
fix: Wallet.delete_key returns False for missing addresses

### LINKED ISSUE #472: Python algosdk wallet.Wallet.delete_key() always returns True
### wallet.Wallet.delete_key() always return True

Per the docs, `Wallet.delete_key()`, should return True if the account has been deleted.
Thus, if the wallet does not contain the account, then False is the expected return result.
However, `Walletdelete_key()` always returns True.

### Your environment

py-algorand-sdk         2.1.2 
algokit                         1.0.1

### Steps to reproduce

```python
import unittest

from algosdk.wallet import Wallet
from beaker import sandbox


class KmdWalletDeleteKeyTestCase(unittest.TestCase):
    def test_delete_key(self):
        kmd_client = sandbox.kmd.get_client()

        # create new wallet
        name = "foo"
        password = "bar"
        kmd_client.create_wallet(name, password)
        wallet = Wallet(name, password, kmd_client)

        # generate new wallet account
        address = wallet.generate_key()
        self.assertTrue(address in wallet.list_keys())

        # delete wallet account
        self.assertTrue(wallet.delete_key(address))
        self.assertFalse(address in wallet.list_keys())

        # delete wallet account again
        self.assertFalse(
            wallet.delete_key(address),
            "should return False because the wallet does not contain the account",
        ) # this assertion fails


if __name__ == "__main__":
    unittest.main()

```

### Expected behaviour
`Wallet.delete_key(address)` should return False if the wallet does not contain the specified address

### Actual behaviour
`Wallet.delete_key(address)` always returns True, even if the wallet does not contain the specified address

mitre88:
## Summary
- `Wallet.delete_key(address)` now returns `False` when the address is not in the wallet (instead of always succeeding via KMD).
- Adds unit tests covering missing and present addresses.

## Context
Fixes #472. Docs say `True` only when the account has been deleted; callers had no way to detect a missing key.

## Validation
```bash
python -m unittest tests.unit_tests.test_wallet -v
```

### DIFF
--- a/algosdk/wallet.py
+++ b/algosdk/wallet.py
@@ -149,9 +149,12 @@ class Wallet:
             address (str): base32 address of account to be deleted
 
         Returns:
-            bool: True if the account has been deleted
+            bool: True if the account has been deleted, False if the wallet
+            does not contain the address
         """
         self.automate_handle()
+        if address not in self.kcl.list_keys(self.handle):
+            return False
         return self.kcl.delete_key(self.handle, self.pswd, address)
 
     def sign_transaction(self, txn):
--- a/tests/unit_tests/test_wallet.py
+++ b/tests/unit_tests/test_wallet.py
@@ -1,0 +1,36 @@
+import unittest
+from unittest.mock import MagicMock
+
+from algosdk.wallet import Wallet
+
+
+class TestWalletDeleteKey(unittest.TestCase):
+    def _wallet_with_keys(self, addresses):
+        kcl = MagicMock()
+        kcl.list_wallets.return_value = [{"name": "w", "id": "id-1"}]
+        kcl.init_wallet_handle.return_value = "handle-1"
+        kcl.list_keys.return_value = list(addresses)
+        kcl.delete_key.return_value = True
+        # bypass __init__ side effects by constructing then overwriting
+        wallet = Wallet.__new__(Wallet)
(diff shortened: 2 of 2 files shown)


jannotti:
I'm not in love with doing a bunch of extra work before the delete just to report whether the key was there. If the API reported the info, fine. Not only is this more work for the rare caller that cares, it's racy because it checks for existence separately from the delete.

mitre88:
You're right — thanks for the pushback.

KMD's `DELETE /v1/key` is effectively idempotent: the sqlite wallet driver runs `DELETE FROM keys WHERE address=?` without checking rows affected, so the API returns `200`/`{}` even when the address was not present. The only client-side way to approximate a `False` return is a `list_keys` pre-check, which (as you note) is both extra work for the common path and racy relative to the delete.

Closing this PR. Honest options if we still want the semantics from #472:

1. **Docs** — document that `delete_key` returns `True` when the KMD request succeeds, not when a key was actually present.
2. **KMD** — return an error when 0 rows are deleted in `go-algorand`, then the SDK can surface that without a pre-check.

Happy to take either direction if maintainers prefer.

mitre88:
Closing per maintainer feedback: pre-check design is racy/extra work; root cause is KMD idempotent DELETE.

